PR #49, opened 2018-02-14
Added XSS prevention helpers

nathanws:
Created a few helper functions within LEAF_Nexus/helpers.php and
LEAF_Request_Portal/helpers.php which will assist in mitigating any
potential Cross Site Scripting vulnerabilities when presenting
output to the user.

The function sanitizeHTML() is a copy of sanitizeInput() from
LEAF_Request_Portal/form.php. The sanitizeInput() function has been
deprecated.


nathanws:
@mgaoVA 

I duplicated `helpers.php` in both `LEAF_Nexus` and `LEAF_Request_Portal`. I didn't see a great place to put shared PHP. The `libs` directory seems like it should only contain third-party code and libraries that are available to custom reports and templates (such as the Javascript APIs), so it should remain that way.

Thoughts on a better way to handle this? Since there are several instances of duplicated code between both projects, it makes sense to create a "commons" for LEAF. Maybe just another directory at the same level as `LEAF_Nexus` and `LEAF_Request_Portal`?

PR #54, opened 2018-02-15
Testing sanitizeHTML

nathanws:
Testing the existing `sanitizeHTML()` function (renamed from `sanitizeInput()`).

Includes code from #49.


nathanws:
Couple of points on testing the `sanitizeHTML()` function:

* Currently it strips all `<ul>` tags out completely, but not `<ol>`, but it leaves all `<li>` intact. Should this be consistent and scrub both `ul` and `ol` out completely? Or should it leave both intact?
* Processing paragraph (`<p>`) tags that contain line breaks (`\n`, `\r`) completely removes the line breaks. Is this the expected behavior? (see `testSanitizeHTML_LineBreaks_Paragraphs()`)
* There is still an issue with unclosed inner tags (see `testSanitizeHTML_UnclosedTags_Inner()`)

Are there any test cases that are missing?

PR #55, opened 2018-02-15
Leaf383 portal form accessibility issues

Nick-Ciardiello:
This pull request also includes the test data


LEAF_Request_Portal/templates/form.tpl
```diff
@@ -18,11 +18,12 @@
                         <!-- <button id="prevQuestion">Previous question</button> -->
                         <!-- <button class="button" dojoType="dijit.form.Button" onclick="checkForm(false);"><div id="save_indicator">Save Change</div></button> -->
                     </div>
+                    <input type="submit" value="Submit" aria-disabled="true" hidden>
                 </form>
             </div>
             <div id="progressArea2" style="height: 34px; background-color: #feffd2; padding: 4px; border-top: 1px solid black">
                 <div style="float: left"><button id="prevQuestion" type="button" class="buttonNorm prevQuestion"><img src="../libs/dynicons/?img=go-previous.svg&amp;w=22" alt="Previous" /> Previous Question</button></div>
```
nathanws: We should also include an aria-label for going back to the previous question. Something like `aria-label="Previous"`.

LEAF_Request_Portal/templates/form.tpl
```diff
@@ -18,11 +18,12 @@
...
-                <div style="float: right"><button id="nextQuestion2" type="button" class="buttonNorm nextQuestion"><img src="../libs/dynicons/?img=go-next.svg&amp;w=22" alt="Next" /> Next Question</button></div>
+                <div style="float: right"><button id="nextQuestion2" type="button" class="buttonNorm nextQuestion"><img src="../libs/dynicons/?img=go-next.svg&amp;w=22" alt="Next" aria-label="Submit"/> Next Question</button></div>
```
nathanws: This should read something like "Next", since "Submit" implies that the form is complete rather than you are navigating to the next question.

PR #56, opened 2018-02-16
Addressed the empty anchor element, h1 and h2 tags for 508 compliance

apgates:


LEAF_Request_Portal/templates/print_form_ajax.tpl
```diff
@@ -33,7 +33,7 @@
 </div>
 <div class="tags<!--{if count($tags) == 0}--> noprint<!--{/if}-->" id="tags" style="border: 1px solid black; padding: 2px; text-align: right">
     <!--{include file="print_form_ajax_tags.tpl" tags=$tags}-->
-    <!--{if count($tags) == 0}--><a href="#" onclick="getTags(<!--{$recordID}-->);"></a><!--{/if}-->
+    <!--{if count($tags) == 0}--><a href="&nbsp" onclick="getTags(<!--{$recordID}-->);"></a><!--{/if}-->
```
nathanws: This should read:

```
<a href="#" onclick="getTags(<!--{$recordID}-->);">&nbsp;</a>
```

There's no guarantee that will make it pass the scan, but it _might_ work. There's also no guarantee it won't affect the layout of the elements, though it shouldn't.

apgates:
the element has been updated. Thanks for the feedback.

PR #59, opened 2018-02-21
Fix xss vulnerability ajaxEmployee and templates

Nick-Ciardiello:


LEAF_Nexus/templates/print_subindicators.tpl
```diff
@@ -27,7 +27,7 @@
             <!--{/if}-->
```
nathanws: I would go ahead and `strip_tags|escape` the following since they are being written to the DOM:
* the two `div` ids above
* the `$uid`, `$categoryID`, and `$indicator.indicatorID` in the two `getForm()` calls above

LEAF_Nexus/templates/print_subindicators_ajax.tpl
```diff
@@ -33,13 +33,13 @@
                 </span>
         <!--{/if}-->
         <!--{if $indicator.format == 'text'}-->
-            <span class="printResponse" id="data_<!--{$indicator.indicatorID}-->_<!--{$categoryID}-->_<!--{$uid}-->">
-                <!--{$indicator.data}-->
+            <span class="printResponse" id="data_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$categoryID|strip_tags|escape}-->_<!--{$uid|strip_tags|escape}-->">
+                <!--{$indicator.data|strip_tags}-->
```
nathanws: As we've very recently discovered, when dealing with `indicator` data, it's possible it contains valid HTML markup, so stripping the tags would remove that valid input. As of right now, we don't have a good way to determine if data should be stripped/escaped/unescaped. So, for now, the best thing to do is nothing when it comes to an `indicator.format` of text. We'll have to revisit this when we come up with a good solution.

This is only a consideration since the `indicator.format` is "text".

LEAF_Nexus/templates/print_subindicators_ajax.tpl
```diff
@@ -33,13 +33,13 @@
...
             </span>
         <!--{/if}-->
         <!--{if $indicator.format == 'number'}-->
             <span class="printResponse" id="data_<!--{$indicator.indicatorID}-->_<!--{$categoryID}-->_<!--{$uid}-->">
-                <!--{$indicator.data}-->
+                <!--{$indicator.data|escape}-->
```
nathanws: Unlike above, this should be fine since the `indicator.format` is "number".

LEAF_Nexus/templates/print_subindicators_ajax.tpl
```diff
@@ -49,18 +49,18 @@
         <!--{/if}-->
         <!--{if $indicator.format == 'date'}-->
             <span class="printResponse" id="data_<!--{$indicator.indicatorID}-->_<!--{$categoryID}-->_<!--{$uid}-->">
-                <!--{$indicator.data|date_format:"%A, %B %e, %Y"}-->
+                <!--{$indicator.data|date_format:"%A, %B %e, %Y"|strip|escape}-->
```
nathanws: This one is also fine, since we know it's a date.

PR #200, opened 2018-06-04
Show "Send Back" time in request timeline

nathanws:
Added option to include any "Send Back" time in the request timeline.
Enabling this option will add an extra bar segment for the average time
elapsed between a step being sent back to the requestor and when the
request is re-submitted.


nathanws:
Forgot to request a merge into dev instead of master...

PR #216, opened 2018-06-13
Leaf497 get indicator logs

jeanmarcablevets:


LEAF_Request_Portal/form.php
```diff
@@ -561,8 +562,17 @@ public function getIndicatorLog($indicatorID, $series, $recordID)
...
             }
 
+            $is_sensitiveVar = array(
+                ':indicatorID' => $indicatorID,
+            );
+
+            $is_sensitive = $this->db->prepared_query(
+                'SELECT is_sensitive FROM indicators
+                WHERE indicatorID=:indicatorID', $is_sensitiveVar
+            );
```
nathanws: In this instance, we should include the `is_sensitive` column along with the query on line 531. This can be done with another LEFT JOIN on the `indicators` table. Only the `is_sensitive` column should be pulled from the `indicators` table, so you'll probably have to select every column by name and table in that query. Something like:

```sql
SELECT h.recordID, h.indicatorID, /* other data_history columns */, i.is_sensitive 
   FROM data_history h, indicators i 
   LEFT JOIN indicators USING (indicatorID) blahblahblah
```

This way all the relevant data is contained in that initial query and the load on the database is reduced. As it is now, this is looping over all the `data_history` results and spawning another database call for each result, which could potentially be a lot.

LEAF_Request_Portal/sources/FormEditor.php
```diff
@@ -156,8 +156,8 @@ function setRequired($indicatorID, $input) {
     }
 
     function setSensitive($indicatorID, $input) {
-	if(is_int((int)$input) == false){
-            return "invalid entry";
+	      if(is_int($input) == false){
+            return 'Not an integer';
```
nathanws: Double check this to make sure I merged it correctly.

LEAF_Request_Portal/form.php
```diff
@@ -526,13 +526,15 @@ public function getIndicatorLog($indicatorID, $series, $recordID)
...
-                AND series=:series
-                ORDER BY timestamp DESC', 
+            'SELECT h.recordID, h.indicatorID, h.series, h.data, h.timestamp, h.userID, i.is_sensitive 
+                FROM data_history h, indicators i 
+                    LEFT JOIN indicator_mask USING (indicatorID)
+                    WHERE h.recordID=:recordID
+                    AND h.indicatorID=:indicatorID
+                    AND h.series=:series
+                    ORDER BY timestamp DESC',
             $vars
```
nathanws: Looks like I made a mistake in my first comment 🤷‍♂️  The query should read:

```sql
SELECT h.recordID, h.indicatorID, h.series, h.data, h.timestamp, h.userID, i.is_sensitive 
                FROM data_history h
                    LEFT JOIN indicator_mask USING (indicatorID)
                    LEFT JOIN indicators i USING (indicatorID)
                    WHERE h.recordID=:recordID
                    AND h.indicatorID=:indicatorID
                    AND h.series=:series
                    ORDER BY timestamp DESC
```

As it is now, when doing a SELECT FROM on `indicators`, it returns multiple rows of the same item from {{data_history}} which results in the indicator log having multiple entries for the same bit of history.

PR #234, opened 2018-07-16
Leaf 551 remediations group.php

jeanmarcablevets:


LEAF_Nexus/sources/Group.php
```diff
@@ -103,7 +103,10 @@ public function addNew($groupTitle, $parentGroupID = 0)
         if(!isset($groupTitle) || strlen($groupTitle) == 0) {
             throw new Exception('Group title must not be blank');
         }
-
+        if(!is_numeric($parentGroupID)){
+            throw new Exception('invalid parent group');
+            return 0;
```
nathanws: The `return 0` can be omitted, throwing the Exception is enough.

LEAF_Nexus/sources/Group.php
```diff
@@ -162,9 +165,13 @@ public function addNew($groupTitle, $parentGroupID = 0)
      */
     public function deleteGroup($groupID)
     {
-    	if($groupID <= 10) {
-    		throw new Exception('Cannot delete system groups.');
-    	}
+        if(!is_numeric($groupID)){
+            throw new Exception('invalid group');
+            return 1;
```
nathanws: This should return `0` indicating failure, since it returns `1` when it is successful.

LEAF_Nexus/sources/Group.php
```diff
@@ -402,6 +412,9 @@ public function getGroupLeader($groupID)
 
     public function listGroupPositions($groupID)
     {
+        if(!is_numeric($groupID)){
+            return array();
```
nathanws: Since `Exceptions` are thrown elsewhere when the input is invalid, instead of returning an empty array, it should throw `new Exception('invalid group');`

LEAF_Nexus/sources/Group.php
```diff
@@ -417,6 +430,9 @@ public function listGroupPositions($groupID)
      */
     public function listGroupEmployees($groupID)
     {
+        if(!is_numeric($groupID)){
+            return array();
```
nathanws: Since `Exceptions` are thrown elsewhere when the input is invalid, instead of returning an empty array, it should throw `new Exception('invalid group');`

LEAF_Nexus/sources/Group.php
```diff
@@ -439,6 +455,9 @@ public function listGroupEmployeesDetailed($groupID, $offset = 0, $limit = -1)
     {
         // Cannot use $this->listGroupEmployees() since that query does not
         // include Employee position data
+        if(!is_numeric($groupID)) {
+            return array();
```
nathanws: Since `Exceptions` are thrown elsewhere when the input is invalid, instead of returning an empty array, it should throw `new Exception('invalid group');`

LEAF_Nexus/sources/Group.php
```diff
@@ -775,6 +802,9 @@ public function getPrivileges($groupID)
      */
     public function togglePermission($groupID, $categoryID, $UID, $permissionType)
     {
+        if(!is_numeric($groupID)||!is_numeric($UID)){
```
nathanws: There shouldn't be a check against `$UID` being numeric, since it's not always guaranteed that it will be numeric, it depends on the class.

LEAF_Nexus/sources/Group.php
```diff
@@ -804,6 +834,9 @@ public function togglePermission($groupID, $categoryID, $UID, $permissionType)
      */
     public function addPermission($groupID, $categoryID, $UID, $permissionType)
     {
+        if(!is_numeric($groupID) || !is_numeric($UID)){
+            return null;
```
nathanws: There shouldn't be a check against `$UID` being numeric, since it's not always guaranteed that it will be numeric, it depends on the class.

nathanws: Looks like you missed this one.

LEAF_Nexus/api/controllers/Group/GET.php
```diff
@@ -18,13 +18,21 @@
     return $group->listGroups($args[0], $_GET['offset'], $_GET['quantity']);
 });
 $this->index['GET']->register('group/[digit]/positions', function($args) use ($group) {
-    return $group->listGroupPositions($args[0]);
+	try {
```
nathanws: Might seem a bit weird, but we should allow the `$group->listGroupPositions()` method to handle any exceptions. The method returns an `Exception` if there is an issue, and the response knows how to handle it.

LEAF_Nexus/api/controllers/Group/GET.php
```diff
@@ -18,13 +18,21 @@
...
+	} catch (Exception $e) {
+        	return $e->getMessage();
+    	}
 });
 $this->index['GET']->register('group/[digit]/leader', function($args) use ($group) {
 	return $group->getGroupLeader($args[0]);
 });
 $this->index['GET']->register('group/[digit]/employees', function($args) use ($group) {
-    return $group->listGroupEmployees($args[0]);
+	try {
```
nathanws: Same as above, no need to catch an `Exception` here.

LEAF_Nexus/api/controllers/Group/GET.php
```diff
@@ -39,8 +47,12 @@
     if (isset($_GET['offset'])) {
         $offset = (int)$_GET['offset'];
     }
-
-    return $group->listGroupEmployeesDetailed($args[0], $searchText, $offset, $limit);
+	
+    try {
```
nathanws: Same as above, no need to catch an `Exception` here.

PR #240, opened 2018-07-18
Bugfix/leaf 561 remediations

dastan-rahimi:


test/LEAF_Request_Portal_Tests/tests/api/FormStackControllerTest.php
```diff
@@ -0,0 +1,40 @@
...
+     }
+
+     /**
+      * Tests the DELETE `formStack/[text]` endpoint.
+      */
+      public function testDeleteForm() : void
+      {
+        $delResponse = self::$client->Delete('?a=formStack/_form_f4688');
+        $this->assertNotNull($delResponse);
+        $this->assertEquals(true, $delResponse);
```
nathanws: There should be an additional check here to ensure it no longer exists. So try to get the form again and make sure the response is empty and/or null.

PR #244, opened 2018-07-18
LEAF554 remediations for LEAF_Nexus/sources/Tag.php

jeanmarcablevets:


LEAF_Nexus/api/controllers/TagController.php
```diff
@@ -26,7 +26,7 @@ public function get($act)
         });
 
         $this->index['GET']->register('tag/[text]/parent', function($args) use ($tag) {
-            return $tag->getParent($args[0]);
+            return $tag->getParent($tag->sanitizeInput($args[0]));
```
nathanws: This should use the sanitize functions in `XSSHelpers.php`. `Tag.php` uses the sanitize function in `Data.php`, which is an old version of the sanitation functions.

LEAF_Nexus/api/controllers/TagController.php
```diff
@@ -39,7 +39,7 @@ public function post($act)
     	$this->index['POST'] = new ControllerMap();
     
     	$this->index['POST']->register('tag/[text]/parent', function($args) use ($tag) {
-    		return $tag->setParent($args[0], $_POST['parentTag']);
+    		return $tag->setParent($args[0], $tag->sanitizeInput($_POST['parentTag']));
```
nathanws: This should use the sanitize functions in `XSSHelpers.php`. `Tag.php` uses the sanitize function in `Data.php`, which is an old version of the sanitation functions.

PR #246, opened 2018-07-23
ATO remediations for LEAF_Request_portal/sources/service.php

jeremyrl:


test/LEAF_Request_Portal_Tests/tests/api/ServiceControllerTest.php
```diff
@@ -0,0 +1,627 @@
+<?php
```
nathanws: This is a great test.

PR #264, opened 2018-07-31
Leaf571 508compliance fixed view form

jeanmarcablevets:
Clear cashe before testing page

### DIFF
--- a/LEAF_Request_Portal/templates/form.tpl
+++ b/LEAF_Request_Portal/templates/form.tpl
@@ -31,7 +31,7 @@
     </div>
     <div class="col span_1_of_5" style="float: left">
         <div id="tools" class="tools"><h1 style="font-size: 12px; text-align: center; margin: 0; padding: 2px">Tools</h1>
-            <button tabindex="0" aria-label="Show Single Page" class="tools" onclick="window.location='?a=printview&amp;recordID=<!--{$recordID}-->'" alt="View full form" title="Show Single Page" style="height: 38px; text-align: left; margin-bottom: 15px; font-size: 120%"/> <img src="../libs/dynicons/?img=edit-find-replace.svg&amp;w=32" style="position: absolute; margin-top: -6px; vertical-align: middle"/>&nbsp &nbsp &nbsp &nbsp Show Single Page</button>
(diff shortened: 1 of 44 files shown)


LEAF_Request_Portal/templates/form.tpl
```diff
@@ -31,9 +31,9 @@
     </div>
     <div class="col span_1_of_5" style="float: left">
         <div id="tools" class="tools"><h1 style="font-size: 12px; text-align: center; margin: 0; padding: 2px">Tools</h1>
-            <div onclick="window.location='?a=printview&amp;recordID=<!--{$recordID}-->'"><img src="../libs/dynicons/?img=edit-find-replace.svg&amp;w=32" alt="View full form" title="View full form" /> Show single page</div>
+            <button tabindex="0" aria-label="Show Single Page" class="tools" onclick="window.location='?a=printview&amp;recordID=<!--{$recordID}-->'" alt="View full form" title="Show Single Page" style="height: 38px; text-align: left; margin-bottom: 15px; font-size: 120%"/> <img src="../libs/dynicons/?img=edit-find-replace.svg&amp;w=32" style="position: absolute; margin-top: -6px; vertical-align: middle"/>&nbsp &nbsp &nbsp &nbsp Show Single Page</button>
```
nathanws: We'll get dinged on a future security scan if we keep the `onclick` that changes the `window.location` instead of using a `<a/>`. Even though the element that was changed is a `<div/>`, it should still be updated to the correct form in these changes.

See the link below for aria role info, note the warning under the "Keyboard and focus" section.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role

LEAF_Request_Portal/templates/form.tpl
```diff
@@ -53,6 +53,24 @@ var currIndicatorID = 0;
 var currSeries = 0;
 var CSRFToken = '<!--{$CSRFToken}-->';
 
+$('#showSinglePage').keypress(function(event) {
+    if(event.keyCode === 32) {
+        alert('#showSinglePageLink'.innerHTML);
```
nathanws: Pretty sure you didn't mean to commit this.

PR #270, opened 2018-08-01
Bugfix/leaf 576 dropdown

dastan-rahimi:


dastan-rahimi:
Feedback has been addressed.

PR #279, opened 2018-08-08
Made printview readable to screen readers

dastan-rahimi:


LEAF_Nexus/js/employeeSelector.js
```diff
@@ -34,6 +34,7 @@ employeeSelector.prototype.initialize = function() {
 	$('#' + this.containerID).html('<div id="'+this.prefixID+'border" class="employeeSelectorBorder">\
 			<div style="float: left"><img id="'+this.prefixID+'icon" src="'+ t.rootPath +'../libs/dynicons/?img=search.svg&w=16" class="employeeSelectorIcon" alt="search" />\
 			<img id="'+this.prefixID+'iconBusy" src="'+ t.rootPath +'images/indicator.gif" style="display: none" class="employeeSelectorIcon" alt="busy" /></div>\
+			<span style="position: absolute; width: 60%; height: 1px; margin: -1px; padding: 0; overflow: hidden; clip: rect(0,0,0,0); border: 0;" aria-atomic="true" aria-live="polite" class="status" role="status"></span>\
```
nathanws: This should generate an ID similar to how the `` does above. Since there can be multiple employee selectors per form, using just the class name to find the element will result in all elements with that class name to be set.

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -98,6 +99,7 @@ function doSubmit(recordID) {
 		data: {CSRFToken: '<!--{$CSRFToken}-->'},
 		success: function(response) {
             if(response.errors.length == 0) {
+                $('.submitStatus').text('Request submmited');
```
nathanws: This should use an ID selector instead of a class selector.

dastan-rahimi: Done

LEAF_Nexus/js/employeeSelector.js
dastan-rahimi: Done

PR #283, opened 2018-08-08
Leaf594 digital signature stamp

jeanmarcablevets:
clear cache before testing

### DIFF
--- a/LEAF_Request_Portal/api/controllers/SignaturesController.php
+++ b/LEAF_Request_Portal/api/controllers/SignaturesController.php
@@ -1,4 +1,7 @@
 <?php
+/*
+ * As a work of the United States government, this project is in the public domain within the United States.
+ */
 
 require '../sources/Signature.php';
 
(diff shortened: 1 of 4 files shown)


LEAF_Request_Portal/api/controllers/SignaturesController.php
```diff
@@ -20,12 +23,22 @@ public function __construct($db, $login)
...
             return $this->API_VERSION;
         });
 
+        $this->index['GET']->register('signature/[digit]', function ($args) use ($signature) {
+            return $signature->getSignature((int)$args[0]);
+        });
+
+        $this->index['GET']->register('signature/[digit]/history', function ($args) use ($signature) {
+            return $signature->getSignatureHistory((int)$args[0]);
+        });
```
nathanws: The controller endpoints `signature/[digit]` and `signature/[digit]/history` should be moved to `FormController` to reflect the changes made in the `Signature` class methods, since these really operate on a record instead of a signature. However the methods below should stay in `Signatures.php`.

LEAF_Request_Portal/sources/Signature.php
```diff
@@ -39,4 +39,46 @@ public function create($signature, $recordID, $message)
...
+
+    public function getSignatureHistory($recordID)
+    {
+        $signatures = $this->getSignature($recordID);
+        $returnArray = array();
+        for ($i = 0; $i < count($signatures); $i++)
+        {
+            $vars = array(
+                ':recordID' => $recordID,
+                ':actionType' => 'sign',
```
jeremyrl: This isn't incorrect exactly, we just never talked about it. But when I worked on LEAF597 I added new action_types and the one I added for this was `signed`, not `sign`. This should be made consistent.

PR #297, opened 2018-08-20
LEAF496 sensitive indicators are hidden

jeremyrl:
Can be unhidden with a button click


LEAF_Request_Portal/templates/print_subindicators_ajax.tpl
```diff
@@ -1,4 +1,13 @@
 <!--{**}-->
+        <!--{if $indicator.is_sensitive == 1}-->
+                <div class="sensitiveIndicatorMaskToggle">
+                    <input type="checkbox" id="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" onClick="toggleSensitiveIndicator(<!--{$indicator.indicatorID|strip_tags}-->, <!--{$indicator.series|strip_tags}-->, this.checked);">
```
jeanmarcablevets: This should be navigate-able via keyboard and interact-able via keyboard.  If you add a `tabindex` and an `onkeypress` function, it will be fine.

jeremyrl: This is done. I also changed the cursor to a pointer.

jeanmarcablevets: The toggler is not interacte-able without NVDA in IE

PR #303, opened 2018-08-25
Leaf581 view inbox page

jeanmarcablevets:
Cherry picked commits from previous branch

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -1865,7 +1865,7 @@ class Form
         $out = array();
         foreach ($recordID_list as $id)
         {
-            if (!is_numeric($id) && id != '') {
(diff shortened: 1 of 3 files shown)


nathanws:
There is still something funky going on with the Inbox. On this branch, any requests that require my attention are not displaying, which they are on the `dev` branch.

Looking at the network calls:

```
GET LEAF_Request_Portal/api/?a=inbox/dependency/_-2&_=1536849579698
```

That is returning the correct data that needs to be displayed, the results just aren't being rendered anymore.

jeanmarcablevets:
Issue was related to the 'form/customData' returning 'false' because of incorrect sanitation on the `508_Compliance` branch that had since been fixed on `dev`.  Commit `a871` addresses this.

PR #304, opened 2018-08-27
LEAF623 admin view form library

jeanmarcablevets:
Remediations for admin view form library


LEAF_Request_Portal/admin/templates/view_form_library.tpl
```diff
@@ -135,16 +136,17 @@ $(function() {
...
-    <div class="buttonNorm" onclick="applyFilter('Fiscal');" style="width: 150px"><img src="../../libs/dynicons/?img=x-office-spreadsheet.svg&amp;w=32" alt="Icon" title="Icon" /> Fiscal</div>
+    <div role="button" onkeydown="triggerKeydown(event, this.id)" class="buttonNorm" tabindex="0" onclick="applyFilter('Administrative')" style="width: 150px" id="Administrative"><img aria-hidden="true" src="../../libs/dynicons/?img=applications-office.svg&amp;w=32" alt="Icon" title="Icon" /> Administrative</div>
+    <div role="button" onkeydown="triggerKeydown(event, this.id)" class="buttonNorm" tabindex="0" onclick="applyFilter('Human Resources')" style="width: 150px" id="Human Resources"><img aria-hidden="true" src="../../libs/dynicons/?img=system-users.svg&amp;w=32" alt="Icon" title="Icon" /> Human Resources</div>
+    <div role="button" onkeydown="triggerKeydown(event, this.id)" class="buttonNorm" tabindex="0" onclick="applyFilter('Information Technology')" style="width: 150px" id="Information Technology"><img aria-hidden="true" src="../../libs/dynicons/?img=network-idle.svg&amp;w=32" alt="Icon" title="Icon" /> Information Technology</div>
```
jeremyrl: ID elements cannot contain spaces. 

Unless the id elements are needed for the screen reader, my suggestion would be to remove the ids all together and just pass `this` as the second parameter to triggerkeydown(). `this` in this case would refer to the whole element triggering the onClick.

jeanmarcablevets: Addressed in commit 0c0

PR #310, opened 2018-08-29
LEAF666 cloning authority form property

jeanmarcablevets:
Created form editor property, "clone authority"

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_form.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_form.tpl
@@ -57,6 +57,10 @@ function openContent(url) {
                                  <td>Sort Priority</td>\
                                  <td><input id="sort" type="number"></input></td>\
                              </tr>\
+                             <tr class="isSubForm">\
+                                 <td>Clone authority</td>\
+                                 <td><select id="cloneGroupID"><option value="0">No group</option></select></td>\
+                             </tr>\
                            </table>');
         $('#name').val(categories[currCategoryID].categoryName);
         $('#description').val(categories[currCategoryID].categoryDescription);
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/admin/templates/mod_form.tpl
```diff
@@ -93,6 +98,25 @@ function openContent(url) {
...
 
+        // load group data
+        dialog.indicateBusy();
+        $.ajax({
+            type: 'GET',
+            url: '../api/?a=group/members',
+            success: function(res) {
+                if(res.length > 0) {
+                    for(var i in res) {
+                        $('#cloneGroupID').append('<option value="'+ res[i].groupID +'">'+ res[i].name +'</input>');
```
nathanws: The `sysadmin` group needs to be filtered out of these results, it will always have a `groupID` of `1`. Sysadmins can clone any form on their RP.

jeanmarcablevets: added conditional in commit 96c excluding sysAdmin

LEAF_Request_Portal/admin/templates/mod_form.tpl
```diff
@@ -93,6 +98,28 @@ function openContent(url) {
...
         });
 
+        // load group data
+        dialog.indicateBusy();
+        $.ajax({
+            type: 'GET',
+            url: '../api/?a=group/members',
+            success: function(res) {
+                if(res.length > 0) {
+                    for(var i in res) {
```
nathanws: The `for..in` loop often causes problems with older versions of IE that we have to support at least in the short term. It's better to use `for(var i=0; i<whatever.length; i++)` loops, which is annoying but it works.

jeanmarcablevets: Addressed in commit c131

PR #312, opened 2018-08-30
LEAF663 Clone Form Button

jeanmarcablevets:
Added clone button to print view page and ran code through php-cs-fixer.

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -113,6 +113,39 @@ class Form
         return $fullForm;
     }
 
+    public function getCloneAuthority($recordID)
+    {
+        $login = $this->login;
+
+        //admin, where groupID = 1, always has authority
+        if ($login->checkGroup(1) == true)
+        {
+            return 1;
+        }
+
+        $vars = array(
+            ':recordID' => $recordID,
+        );
+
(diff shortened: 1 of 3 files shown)


LEAF_Request_Portal/form.php
```diff
@@ -113,6 +113,39 @@ public function getFullForm($recordID, $limitCategory = null)
         return $fullForm;
     }
 
+    public function getCloneAuthority($recordID)
+    {
+        $login = $this->login;
+
+        //admin, where groupID = 1, always has authority
+        if ($login->checkGroup(1) == true)
```
jeremyrl: `== true` is not necessary here

LEAF_Request_Portal/form.php
```diff
@@ -113,6 +113,39 @@ public function getFullForm($recordID, $limitCategory = null)
...
+            'SELECT cloneGroupID
+             FROM category_count 
+             LEFT JOIN categories
+             USING (categoryID) 
+             WHERE recordID=:recordID;',
+            $vars
+        );
+
+        //if cloneGroupID = user's groupID, can clone
+        if ($login->checkGroup((int)$res['cloneGroupID']) == true)
```
jeremyrl: `== true` is not necessary here

PR #314, opened 2018-08-31
LEAF650 ATO Remediations

jeremyrl:


LEAF_Request_Portal/index.php
```diff
@@ -91,17 +91,23 @@ function customTemplate($tpl)
         {
             $categoryArray[$key] = array_map('XSSHelpers::xscrub', $cat );
         }
+
+        $servicesArray = $stack->getServices2();
```
nathanws: This should be `$form->getServices2()`. `FormStack` has no method `getServices2()` so this throws an error when creating new Requests.

jeremyrl: Fixed

PR #355, opened 2018-09-26
LEAF697 Redirect To PIV Login

jeanmarcablevets:
Redirect user to PIV login screen on error e.g. when domain is down

### DIFF
--- a/LEAF_Nexus/auth_domain/index.php
+++ b/LEAF_Nexus/auth_domain/index.php
@@ -103,11 +103,15 @@ if (isset($_SERVER['REMOTE_USER']))
         }
         else
         {
-            echo 'Unable to log in: User not found in global database.';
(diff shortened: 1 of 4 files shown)


LEAF_Nexus/auth_domain/index.php
```diff
@@ -103,11 +103,15 @@
         }
         else
         {
-            echo 'Unable to log in: User not found in global database.';
+            header('Refresh: 4;URL=' . $_SERVER["HTTP_REFERER"]);
```
mgaoVA: HTTP_REFERER is not very reliable, this should instead use a relative path to ../login/index.php. Check syntax for this.

jeanmarcablevets: Addressed in commit `134`

PR #362, opened 2018-10-03
LEAF 384 Secure Entry for PII/PHI

jeanmarcablevets:
Enabled secure entry for PII/PHI and made sensitive forms need to know

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_form.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_form.tpl
@@ -7,6 +7,16 @@
 <script>
 
 var currCategoryID = '';
+function checkSensitive(indicatorID, index = 0) {
+    if(indicatorID[index] === undefined) {
+        return 0;
+    } else if (indicatorID[index].is_sensitive === '1') {
+        return 1;
+    } else {
+        checkSensitive(indicatorID[index + 1]);
+    }
+}
+
 function openContent(url) {
 	var isSubForm = categories[currCategoryID].parentID == '' ? false : true;
 	var formTitle = categories[currCategoryID].categoryName == '' ? 'Untitled' : categories[currCategoryID].categoryName;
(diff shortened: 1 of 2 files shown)


LEAF_Request_Portal/admin/templates/mod_form.tpl
```diff
@@ -58,6 +68,19 @@ function openContent(url) {
...
                            </table>');
+        $.ajax({
+            type: 'GET',
+            url: '../api/form/_' + currCategoryID,
+            success: function(res) {
+                if(res.length > 0) {
+                    $("#needToKnow").val(checkSensitive(res));
+                    if($("#needToKnow").val() === '1') {
+                        $("#needToKnow option[value='0']").remove();
+                        $("#needToKnow option[value='1']").html('Forced on because sensitive fields are present');
```
jeremyrl: Set the value of `#needToKnow` within the conditional here. If the form is set to need to know in the `Edit Properties` popup, but has no sensitive indicators, it will show as `off` instead of `on`. Just use the call to  `checkSensitive()` in the conditional: `if(checkSensitive(res) === 1) `

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -690,6 +690,36 @@
...
+                    $('#<!--{$indicator.indicatorID|strip_tags}-->').css('display', 'none');
+                } else {
+                    $('#<!--{$indicator.indicatorID|strip_tags}-->').attr('type', 'password');
+                }
+                $('#sensitiveStatus').attr('aria-label', 'sensitive data hidden');
+                visible = 0;
+            }
+        }
+        </script>
+        <!--{/if}-->
```
jeremyrl: It works fine if it's just the parent indicator but it goes nuts if there are any children marked as sensitive.

jeremyrl:
Also, I think that if we are setting the form to need to know when an indicator is set to sensitive, we should probably update the category table at that time and not just the visual queues.

LEAF_Request_Portal/admin/templates/mod_form.tpl
```diff
@@ -7,6 +7,16 @@
...
+function checkSensitive(indicatorID, index = 0) {
+    if(indicatorID[index] === undefined) {
+        return 0;
+    } else if (indicatorID[index].is_sensitive === '1') {
+        return 1;
+    } else {
+        checkSensitive(indicatorID[index + 1]);
+    }
+}
+
```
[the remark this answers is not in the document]
jeanmarcablevets: The code you provided worked perfectly, actually.  Addressed in commit `7a4`

LEAF_Request_Portal/admin/templates/mod_form.tpl
jeanmarcablevets: Addressed in commit `7a4`

LEAF_Request_Portal/templates/subindicators.tpl
jeanmarcablevets: Addressed in commit `7a4`

jeanmarcablevets:
> Also, I think that if we are setting the form to need to know when an indicator is set to sensitive, we should probably update the category table at that time and not just the visual queues.

Addressed in commit `7a4`

PR #363, opened 2018-10-03
LEAF703 JSON scrubbed

jeremyrl:
ATO remediations

### DIFF
--- a/LEAF_Request_Portal/api/controllers/ConverterController.php
+++ b/LEAF_Request_Portal/api/controllers/ConverterController.php
@@ -42,7 +42,6 @@ class ConverterController extends RESTfulResponse
         });
 
         $this->index['POST']->register('converter/json', function ($args) {
-            trigger_error("done did the thing", E_USER_WARNING);
             return XSSHelpers::scrubObjectOrArray(json_decode($_POST['input'], true));
         });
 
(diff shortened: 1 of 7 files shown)


LEAF_Request_Portal/api/controllers/ConverterController.php
```diff
@@ -42,7 +42,8 @@ public function post($act)
         });
 
         $this->index['POST']->register('converter/json', function ($args) {
-            return json_decode($_POST['input'], true);
+            trigger_error("done did the thing", E_USER_WARNING);
```
mgaoVA: Remove

LEAF_Request_Portal/sources/FormStack.php
```diff
@@ -110,12 +110,12 @@ public function importForm($overwiteExisting = false)
         }
         if (isset($_POST['formPacket']))
         {
-            $formPacket = json_decode($_POST['formPacket'], true);
+            $formPacket = XSSHelpers::scrubObjectOrArray(json_decode($_POST['formPacket'], true));
```
mgaoVA: Need to revert this one and the one below it as it's necessary for functionality. Might need further discussion.

jeremyrl: Would it be better to do `sanitizeHTML` or `sanitizeHTMLRich`? Or do we need to try a different approach?

mgaoVA: Totally different approach. The issue is that importForm() is used as part of the Form Library system, where admins would be able to include and deploy custom HTML/JS.

jeremyrl:
The requested changes have been made.

PR #375, opened 2018-10-10
LEAF-384_addressedFeedback

jeanmarcablevets:
Addressed Michael's feedback from sprint planning secure entry for PII/PHI demo.
-made printview sensitive data change on hover
-filtered sensitive toggler on textareas

### DIFF
--- a/LEAF_Request_Portal/templates/print_subindicators_ajax.tpl
+++ b/LEAF_Request_Portal/templates/print_subindicators_ajax.tpl
@@ -1,17 +1,17 @@
 <!--{**}-->
         <!--{if $indicator.is_sensitive == 1}-->
             <div class="sensitiveIndicatorMaskToggle">
-                <input type="checkbox" id="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" onClick="toggleStayVisible(); toggleSensitiveIndicator(<!--{$indicator.indicatorID|strip_tags}-->, <!--{$indicator.series|strip_tags}-->, this.checked);">
+                <input type="checkbox" id="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" onClick="toggleStayVisible_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->(); toggleSensitiveIndicator(<!--{$indicator.indicatorID|strip_tags}-->, <!--{$indicator.series|strip_tags}-->, this.checked);">
                 <label for="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" title="Show Sensitive Data" alt="Show Sensitive Data"  tabindex="0" onkeydown="if (event.keyCode==13){ this.click();}"></label>
             </div>
             <span class="sensitiveIndicator-masked" id="<!--{$indicator.indicatorID|strip_tags}-->_masked">
                 *****
             </span>
             <script>
-                var stayVisible = false;
+                var stayVisible_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}--> = false;
                 $("#<!--{$indicator.indicatorID|strip_tags}-->_masked").on({
                     mouseenter: function () {
-                        if (stayVisible === false) {
+                        if (stayVisible_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}--> === false) {
                             $("#sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->").prop('checked', true);
                             toggleSensitiveIndicator(<!--{$indicator.indicatorID|strip_tags}-->, <!--{$indicator.series|strip_tags}-->, $("#sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->").prop('checked'));
                         }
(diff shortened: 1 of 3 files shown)


LEAF_Request_Portal/templates/print_subindicators_ajax.tpl
```diff
@@ -1,12 +1,19 @@
...
+            <span class="sensitiveIndicator-masked">
+                *****
+            </span>
+            <script>
+                $("div.sensitiveIndicator#xhrIndicator_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->").hover(function() {
+                    $("div.sensitiveIndicator#xhrIndicator_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}--> > div.sensitiveIndicatorMaskToggle > label").trigger('click');
+                }, function() {
+                    $("div.sensitiveIndicator#xhrIndicator_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}--> > div.sensitiveIndicatorMaskToggle > label").trigger('click');
+                })
+            </script>
```
jeremyrl: A few things:
1.) If an indicator has any children, the hidden text is toggled when you hover over any of the child elements, this is especially noticeable when you hover over a sensitive child indicator to a sensitive parent (they both reveal at the same time). I would just make the selector more specific.
2.) On the printview page - if you edit a field, the toggle stops working. This is because the div is removed and re-added wit javascript, so the event is not bound to the new element. To bind to dynamically added elements, you should use the .on function in jquery. This link will show you how, specifically the edit from DECEMBER 11, 2012 (https://stackoverflow.com/questions/9827095/is-it-possible-to-use-jquery-on-and-hover).
3.) If you are hovering, the field becomes visible. Then you click the eye, it becomes invisible. Then you un-hover and it becomes permanently visible. I'm not sure how we want the click to effect the visibility in the end, but I would suggest adding a param to the toggle function that takes a visibility state, on or off basically. And if you mouseover toggle(thisid, on), if you mouseout toggle(thisid, off).

jeanmarcablevets: Addressed 1 and 2 in commit `a89`.  @mgaoVA please give clarification on what should be done with 3.

nathanws: For 3, If the indicator has been clicked to be shown, the hover effect should go away. In other words, if it is shown, hovering over it does nothing. The hover effect should only happen if the indicator is hidden.

jeanmarcablevets: Addressed in commit `b5f`

LEAF_Request_Portal/templates/print_subindicators_ajax.tpl
```diff
@@ -1,21 +1,28 @@
 <!--{**}-->
         <!--{if $indicator.is_sensitive == 1}-->
             <div class="sensitiveIndicatorMaskToggle">
-                <input type="checkbox" id="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" onClick="toggleSensitiveIndicator(<!--{$indicator.indicatorID|strip_tags}-->, <!--{$indicator.series|strip_tags}-->, this.checked);">
-                <label for="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" title="Show Sensitive Data" alt="Show Sensitive Data"  tabindex="0" onkeydown="if (event.keyCode==13){ this.click(); }"></label>
+                <input type="checkbox" id="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" onClick=" if (stayVisible === 0) {stayVisible = 1;} else {stayVisible = 0;} toggleSensitiveIndicator(<!--{$indicator.indicatorID|strip_tags}-->, <!--{$indicator.series|strip_tags}-->, this.checked);">
```
nathanws: Inline Javascript like this should be avoided as much as possible. Instead, just create a function to handle toggling the `stayVisible` variable.

Also if the `stayVisible` variable is changed to a boolean, you could just say `stayVisible = !stayVisible` in the toggle function.

jeanmarcablevets: Addressed in commit `af0`

LEAF_Request_Portal/templates/print_subindicators_ajax.tpl
```diff
@@ -1,12 +1,36 @@
...
-                    <label for="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" title="Show Sensitive Data" alt="Show Sensitive Data"  tabindex="0" onkeydown="if (event.keyCode==13){ this.click(); }"></label>
-                </div>
-                <span class="sensitiveIndicator-masked">
-                    *****
-                </span>
+            <div class="sensitiveIndicatorMaskToggle">
+                <input type="checkbox" id="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" onClick="toggleStayVisible(); toggleSensitiveIndicator(<!--{$indicator.indicatorID|strip_tags}-->, <!--{$indicator.series|strip_tags}-->, this.checked);">
+                <label for="sensitiveIndicatorMaskCheckbox_<!--{$indicator.indicatorID|strip_tags}-->_<!--{$indicator.series|strip_tags}-->" title="Show Sensitive Data" alt="Show Sensitive Data"  tabindex="0" onkeydown="if (event.keyCode==13){ this.click();}"></label>
```
nathanws: The `title` and `alt` for this label should also be updated depending on the state of the UI. So if the data is hidden, it should say "Show Sensitive Data". If the data is not hidden, it should say "Hide Sensitive Data".

jeanmarcablevets: Addressed in commit `2f5`

PR #401, opened 2018-10-30
Leaf690 file name scrubber

jeremyrl:


jeremyrl:
To test the script:

1. Make sure you have a form in your test environment that has file uploads and image uploads
2. Switch to the dev branch and add some files/images to a requestusing that form
3. Switch to this branch and call `php portal/scripts/updateUploadedFiles.php`
4. Reload the request you created in step 2 and make sure that all images and files show up properly
5. Spot check the uploads folder to make sure that all the uploaded files are hashed (if there are old files in here that do not show up in the data table, they will not be hashed)

PR #404, opened 2018-11-05
LEAF769 editing masked indicators

jeanmarcablevets:
Regardless of toggle, unmasked indicator data appears in the edit box.


jeanmarcablevets:
@jeremyrl I removed the toggle from the indicator editor on the printview page in commit `89b`, but are you sure we should do that for when answering individual questions?  That was a feature added in LEAF384 and removing it seems like it would be a regression.

PR #439, opened 2018-11-29
Implemented Edit/Delete for Workflow Actions

dastan-rahimi:


LEAF_Request_Portal/api/controllers/SystemController.php
```diff
@@ -81,6 +81,10 @@ public function get($act)
             return $system->getReportTemplate($args[0]);
         });
 
+        $this->index['GET']->register('system/action/[text]', function ($args) use ($login, $system) {
```
nathanws: This should be moved to `WorkflowController.php`.

LEAF_Request_Portal/api/controllers/SystemController.php
```diff
@@ -104,6 +108,10 @@ public function post($act)
             return $system->addAction();
         });
 
+        $this->index['POST']->register('system/editAction/[text]', function ($args) use ($db, $login, $system) {
```
nathanws: This should be moved to `WorkflowController.php`.

LEAF_Request_Portal/api/controllers/SystemController.php
```diff
@@ -163,6 +171,10 @@ public function delete($act)
             return $system->removeFile($args[0]);
         });
 
+        $this->index['DELETE']->register('system/action/[text]', function ($args) use ($db, $login, $system) {
```
nathanws: This should be moved to `WorkflowController.php`.

LEAF_Request_Portal/sources/System.php
```diff
@@ -273,6 +286,49 @@ public function addAction()
 										VALUES (:actionType, :actionText, :actionTextPasttense, :actionIcon, :actionAlignment, :sort, :fillDependency)', $vars);
     }
 
+    public function editAction($actionType)
```
nathanws: This should be moved to `Workflow.php`.

LEAF_Request_Portal/sources/System.php
```diff
@@ -273,6 +286,49 @@ public function addAction()
...
+                ':fillDependency' => $_POST['fillDependency'],
+        );
+
+        $this->db->prepared_query('UPDATE actions SET actionType=:actionType, actionText=:actionText, actionTextPasttense=:actionTextPasttense, actionIcon=:actionIcon, actionAlignment=:actionAlignment, sort=:sort, fillDependency=:fillDependency WHERE actionType=:oldActionType', $vars);
+
+        return 1;
+    }
+
+
+    public function removeAction($actionType)
```
nathanws: This should be moved to `Workflow.php`.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -6,6 +6,7 @@
     <div id="btn_newWorkflow" class="buttonNorm" onclick="newWorkflow();" style="font-size: 120%" role="button" tabindex="0"><img src="../../libs/dynicons/?img=list-add.svg&w=32" alt="New Workflow" /> New Workflow</div><br />
     <br />
     <div id="btn_deleteWorkflow" class="buttonNorm" onclick="deleteWorkflow();" style="font-size: 100%; display: none" role="button" tabindex="0"><img src="../../libs/dynicons/?img=list-remove.svg&w=16" alt="Delete workflow" /> Delete workflow</div><br />
+    <div id="btn_listActionType" class="buttonNorm" onclick="listActionType();" style="font-size: 100%; display: none" role="button" tabindex="0">List Action Type</div><br />
```
nathanws: Should probably change "List Action Type" to "Edit Actions".

In LEAF, there is a difference between "action type" and "action", so "Type" should be left out. This also applies to any methods/variables used when editing Actions.

Also, the default actions that ship with LEAF should not be editable/deletable, I think there are seven of them.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -413,6 +414,113 @@ function setInitialStep(stepID) {
...
+	dialog.hide();
+  $("#button_save").hide();
+	dialog.setTitle('List of Actions');
+	dialog.show();
+  $.ajax({
+		type: 'GET',
+		url: '../api/?a=workflow/userActions',
+		success: function(res) {
+			var buffer = '';
+			buffer += '<br /><table id="actions" border="1"><caption><h2>List of Current Actions:</h2></caption><tr><th scope="col">Action Name</th><th scope="col">Action</th></tr>';
```
nathanws: When showing Actions that are available to edit, it should also display the actions Past Tense name. 

And the "Action" column that contains the Edit/Delete buttons should not have a column name, calling it "Action" while editing Actions might be confusing to some Users.

dastan-rahimi: done

PR #446, opened 2018-12-06
supress warnings in sanitizer

jeremyrl:
There are tons of warnings clogging up the error logs, but nothing is functionally broken. We're just turning the warnings off.

### DIFF
--- a/libs/php-commons/XSSHelpers.php
+++ b/libs/php-commons/XSSHelpers.php
@@ -61,6 +61,7 @@ class XSSHelpers
      */
     public static function sanitizer($in, $allowedTags = array(), $encoding = 'UTF-8')
     {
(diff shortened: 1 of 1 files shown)


libs/php-commons/XSSHelpers.php
```diff
@@ -61,6 +61,7 @@ public static function xscrub($data)
      */
     public static function sanitizer($in, $allowedTags = array(), $encoding = 'UTF-8')
     {
+        $errorReportingLevel = error_reporting(0);//turn off errors for the next few lines
```
nathanws: It might be a little better to use:

```
error_reporting(E_ALL ^ E_WARNING); 
```

instead, since it's only the warnings that should be suppressed.

PR #464, opened 2018-12-12
LEAF783 mass action feature

jeremyrl:


LEAF_Request_Portal/templates/reports/LEAF_mass_action.tpl
```diff
@@ -0,0 +1,452 @@
...
+    if(totalActions)
+    {
+        $('button.takeAction').attr("disabled", "disabled");
+    }
+    $.each(selectedRequests, function(key, item) {
+        var ajaxPath = '';
+        var ajaxData = {};
+        var recordID = $(item).val();
+        switch(actionValue) {    
+            case 'submit':
```
jeanmarcablevets: When a request is submitted with an error, the UI says it was a success.  Specifically, a request tied to an incorrectly configured workflow (like a workflow that that has no finish point) says the submission was successful, even though the network log says there was an error.

jeremyrl: This will be addressed next sprint.

PR #469, opened 2018-12-12
quick fix for CommonConfg include

jeremyrl:


jeremyrl:
There was a small typo in the last fix

PR #492, opened 2018-12-20
LEAF839 Import Data Into New Form

jeanmarcablevets:
Added ability to make new a form and requests from an uploaded sheet.

Also added:

* activity/import progress indicator
* error checking: 
     * prevents columns with blank entries from being selected as required or in a required indicator
* generates report of requests made

### DIFF
--- a/LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
+++ b/LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
@@ -386,13 +386,13 @@
 
                                                                 },
                                                                 function (err) {
-                                                                    console.log("Error retrieving employee: " + err);
+                                                                    console.log("Error retrieving employee for indicator at row " + index + ": " + err);
                                                                 }
                                                             );
                                                         }
                                                     },
                                                     function (err) {
-                                                        console.log("Error retrieving email: " + err);
+                                                        console.log("Error retrieving email for indicator at row " + index + ": " + err);
                                                     }
                                                 );
                                             } else {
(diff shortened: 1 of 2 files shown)


LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
```diff
@@ -169,37 +270,54 @@
...
         var columnSelect = $(document.createElement('td'))
-            .append(buildSheetSelect(indicator.indicatorID, sheet_data))
+            .append(buildSheetSelect(indicator.indicatorID, sheet_data, indicator.required))
             .appendTo(row);
 
         return row;
     }
 
-    function doStuff(initiator, requestData) {
+    function doStuff(categoryID, initiator, requestData) {
```
jeremyrl: Rename this function to something that describes what it does.

libs/jsapi/portal/LEAFPortalAPI.js
```diff
@@ -386,6 +386,99 @@ var PortalFormEditorAPI = function (baseAPIURL) {
                 .fail(onFail);
         },
 
+        /**
+         * Assigns workflow to form
+         *
+         * @param categoryID    string              name of form
+         * @param workflowID    int                 description of form
```
jeremyrl: These comments are incorrect.

LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
jeanmarcablevets: Addressed in commit `ddf`

libs/jsapi/portal/LEAFPortalAPI.js
jeanmarcablevets: Addressed in commit `ddf`

PR #495, opened 2018-12-26
Fixed Type error caused in reports

dastan-rahimi:


LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -281,17 +281,19 @@ function loadSearchPrereqs() {
...
                     for(var j in groupIDmap[i].parentStaples) {
                         associatedCategories += ' ' + groupIDmap[i].parentStaples[j];
                     }
                 }
-                
+
                 var categoryLabel = groupNames[k].categoryName;
                 if(groupIDmap[i].parentCategoryID != '') {
-                    categoryLabel += "<br />" + groupIDmap[groupIDmap[i].parentCategoryID].categoryName;
+                    if(groupIDmap[groupIDmap[i].parentCategoryID]){
```
jeremyrl: Use `&&` instead of nesting if statements here.

dastan-rahimi: done

PR #496, opened 2018-12-26
Removed Unassign Button not needed

dastan-rahimi:


LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -651,10 +651,9 @@
             <!--{$indicator.html}-->
         <!--{/if}-->
         <!--{if $indicator.format == 'orgchart_employee' && ($indicator.isMasked == 0 || $indicator.data == '')}-->
-            <button id="btn_removeEmployee_<!--{$indicator.indicatorID}-->" class="buttonNorm" style="display: none">Unassign Employee</button>
```
jeremyrl: Why is this button unnecessary?

PR #533, opened 2019-02-12
[Leaf-903] refactor form editor ui mockup

thomasnguyen95:
Can add, edit, and remove sections
Can add, edit questions
Needs to address draggable question into section


thomasnguyen95:
Codacy checks for:
- shorthand function: works in ES6, but to be compatible with IE these checks are invalid.
- undefined variables (Vue, VueJSModal, VueChart, etc): are actually defined and available in the document when loading the html mockup

thomasnguyen95:
Closing this PR, ticket LEAF-923 continues off of this branch

PR #581, opened 2019-04-15
Return public key in json payload from ws

Nick-Ciardiello:
Json Contract:

Request from LEAF:
{
  "key": "my_key",
  "dataToSign": "{ "json": "string", "to": "sign" }"
}
Response from websocket:
{
  "key": "my_key",
  "message": "signed_data_as_hexadecimal",
  "status": "SUCCESS",
  "publicKey": "signing_certificate_as_string"
}


Nick-Ciardiello:
I added a verify() method to the SignEngine.  It prints whether you can verify the signature with the certificate.  To demo just run the jar in the console, try to sign something within LEAF, and it'll print the status of the verification

PR #585, opened 2019-04-19
LEAEF946 Textarea Fix

jeanmarcablevets:
replaced nl2br to handle double linebreak bug


Nick-Ciardiello:
You spelled LEAF wrong

nathanws:
Closing since the new library will not be implemented until the UI refresh.

PR #597, opened 2019-05-08
Leaf944 form printing upgrade

jeanmarcablevets:
Updates:
* fixed wrapping issue
* moved printing functionality to its own file


Nick-Ciardiello:
It took three commits to add one line of code?

PR #614, opened 2019-06-10
Leaf-1042 file name changes in portal

jeremyrl:


LEAF_Request_Portal/file.php
```diff
@@ -8,6 +8,7 @@
 include 'db_config.php';
 include 'login.php';
```
thomasnguyen95: For file name consistency, can you make ```login.php``` -> ```Login.php```

PR #615, opened 2019-06-10
Leaf 945 file upper case extensions

mohnishvaid:
Fixed case sensitive extensions of files in the File Manager.


LEAF_Request_Portal/sources/System.php
```diff
@@ -629,40 +532,34 @@ public function newFile()
...
         }
-
         if (!$this->login->checkGroup(1))
         {
             return 'Admin access required';
         }
-
-        move_uploaded_file($_FILES['file']['tmp_name'], __DIR__ . '/../files/' . $fileName);
+        
+        move_uploaded_file($_FILES['file']['tmp_name'], __DIR__ . '/../files/' . strtolower($fileName));
```
nathanws: The filename case itself should not be changed.

PR #616, opened 2019-06-11
Added width for dropdown menu.

mohnishvaid:
Made it a size of 150px to accomodate bigger words.

### DIFF
--- a/LEAF_Request_Portal/templates/initial_form.tpl
+++ b/LEAF_Request_Portal/templates/initial_form.tpl
@@ -87,7 +87,7 @@ $(function() {
                 <td>Priority</td>
                 <td>
                     <select id="priority" name="priority" aria-label="priority" style="width: 150px;">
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/templates/initial_form.tpl
```diff
@@ -86,8 +86,8 @@ $(function() {
             <tr>
                 <td>Priority</td>
                 <td>
-                    <select id="priority" name="priority" aria-label="priority">
-                    <option value="-10">EMERGENCY</option>
+                    <select id="priority" name="priority" aria-label="priority" style="width: 150px;">
+                    <option value="10">EMERGENCY</option>
```
nathanws: The value of the option was changed from "-10" to "10", it should remain "-10"

mohnishvaid: Sorry. Changed it back to -10.

PR #633, opened 2019-07-30
LEAF-1068

mohnishvaid:
Fixed grid formatting and color of styling, currency styling (thousands separator and dollar sign), priority queue ordering when printing to pdf.

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -115,7 +115,7 @@ class Form
      * @param string $limitCategory
      * @return array
      */
-    public function getFullForm($recordID, $limitCategory = null)
+    public function getFullForm($recordID, $limitCategory = null, $childArray = false)
     {
         $fullForm = array();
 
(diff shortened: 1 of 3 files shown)


LEAF_Request_Portal/form.php
```diff
@@ -136,6 +136,33 @@ public function getFullForm($recordID, $limitCategory = null)
         return $fullForm;
     }
 
+    /**
+     * Copy of getFullForm, but returns sub-indicators as array to preserve order
+     * @param int $recordID
+     * @param string $limitCategory
+     * @return array
+     */
+    public function getFullFormPrint($recordID, $limitCategory = null)
```
jeremyrl: Instead of creating a new function here, just add a param and conditional to getFullForm() to avoid repeating code.

LEAF_Request_Portal/js/formPrint.js
```diff
@@ -340,9 +340,16 @@ var printer = function() {
                                 .text()) : '';
                             textSub();
                             break;
+                        case 'currency':
```
jeremyrl: I would round the cents up to the nearest 2 decimal points, in cases like $100.555.This would keep it consistent to the normal view which already does this.

PR #674, opened 2019-10-23
make userID and empUID available in the template

mgaoVA:


jeremyrl:
After talking about this with Dastan, I think it might be better if we wait until after the uuid migration before doing this. The empUID won't be an int, for one, and we don't know what else might pop up.

PR #685, opened 2019-10-31
Adding Data Item Log Tables & Logging to Groups

g-theriot:
Added Logging to Data Class to allow classes extending Data to write logs in new Data Action Log Table.


LEAF_Nexus/templates/view_group.tpl
```diff
@@ -127,6 +129,23 @@ function announceAction(actionName) {
     $('#buttonClick').attr('aria-label', 'clicked ' + actionName);
 }
 
+function viewHistory(){
+    dialog_message.setContent('');
```
dastan-rahimi: Add the title to read "Acess Group History" dialog.setTitle('Acess Group History');

PR #818, opened 2020-02-21
Leaf 1189 create workflow view history

dastan-rahimi:


dastan-rahimi:
> Ran the branch locally and the History function looks good. I wasn't able to get any history data to show in the dialog, is it because it was running locally?

There was an issue with dialog rendering. it has been fixed.

thomasnguyen95:
Try this case: Take a step in the workflow, edit the step name > rename to whatever 
The view history says: "Workflow Step # jsSrc changed to empty stepTitle changed to step2"
'jsSrc' should not be included in the readable history

g-theriot:
> Try this case: Take a step in the workflow, edit the step name > rename to whatever
> The view history says: "Workflow Step # jsSrc changed to empty stepTitle changed to step2"
> 'jsSrc' should not be included in the readable history

Updated so the jssrc doesn't show in log

PR #844, opened 2020-03-12
bugfix/LEAF-1345-whitelist - disallow html uploads in non-admin contexts

mgaoVA:
cleanup commonconfig


patrickcgs:
title references files with extensions .html

PR also prevents attachments with extensions:

```
zip
ics
svg
```

@mgaoVA was this intended?

cc: @dastan-rahimi

mgaoVA:
nice catch, going to re-add a few extensions. Excluding svg, it might still be flagged since it can contain JS.

PR #852, opened 2020-03-18
return true since expected empID matches current user's empID

mgaoVA:


mgaoVA:
checkIfBackup returns false when $empUID != $this->login->getEmpUID()
but it never returns true when $empUID == $this->login->getEmpUID()

PR #896, opened 2020-04-09
1302-3: Change Directory Where CSS is Compiled

cameron-chilton:
- Edited gulpfile to compile CSS into 'libs/css'
- Only leaf.scss is in the 'sass' folder


cameron-chilton:
didn't base off dev in PR

PR #909, opened 2020-04-15
LEAF-1409: Sitemaps UI Design

cameron-chilton:
- New template file based on low fidelity prototype (attached), screenshots from browser also below
- Some new classes created in leaf.scss
- Used existing jquery UI dialog for example of editing/creating cards in a dialog

1316-sitemaps-2-2
1316-sitemaps-2
Screen Shot 2020-04-15 at 8 31 47 AM
Screen Shot 2020-04-15 at 8 31 36 AM


cameron-chilton:
> "Card Color:" field is a color picker? if so doesn't seem to work

Color picker control was a placeholder added while awaiting feedback on how the control should be configured. I've removed it for now.

PR #943, opened 2020-05-07
Adds reverse-proxy for /swagger/

patrickcgs:
Adds reverse-proxy for /swagger/

Swagger UI generates links for content served at a url. CORS
limitations are in effect that prevent swagger ui from executing
requests against local development environment.

Installs apache proxy modules and supporting configuration.  Requests
to http://localhost/swagger/ or https://localhost/swagger/ are now
redirected to swagger-ui.

Browser now only needs to use http://localhost/swagger/

Dockerfile has ben modified.  Will need to rebuild the docker image:

docker-compose build php
docker-compose up -d --force-recreate --no-deps php


docker/docker-compose.yml
```diff
@@ -29,15 +29,22 @@ services:
     image: swaggerapi/swagger-ui
     container_name: "swagger-ui"
     ports:
-      - "8081:8080"
+      - "8081:8080" # debug only; use reverse-proxy http://localhost/swagger/
     restart: 'always'
     volumes:
-      - ./swagger/nexus-swagger.json:/usr/share/nginx/html/api/nexus-swagger.json
-      - ./swagger/portal-swagger.json:/usr/share/nginx/html/api/portal-swagger.json
+      - ./swagger/nexus-swagger.json:/usr/share/nginx/html/apiapi/nexus-swagger.json
```
thomasnguyen95, line 35: why ```apiapi``` instead of just ```api```

PR #948, opened 2020-05-15
Feature/leaf1494 configuration strategy

jeremyrl:
I'm working on windows, but in my vhosts file, I had every request redirect to the new index.php file in the root level. That file sets the configurations and then goes into the rest of the app. 
Next steps: 
- develop a strategy to load static files (css, js, etc.) without going through this process
- add every landing page to the current functionality (any index.php that gets hit, any ajaxIndex.php, etc.)
- fix the login procedure. Currently there is a line in the login.php file that fakes the login, but this will need a permanent fix.  This may just be a matter of adding the page to the routing and fixing any config issues present, will have to look into it when the time comes.


LEAF_Nexus/sources/Login.php
```diff
@@ -198,7 +198,7 @@ public function loginUser()
             $authType = '/auth_cookie/?r=';
             $nonBrowserAuth = '/auth_cookie/?r=';
         }
-
+        $_SESSION['userID'] = 'VACOLayJ';//TODO MAKE STATELESS LOGIN WORK CORRECTLY
```
thomasnguyen95, line 201: Can this be the tester user, dont think you should include this specific username

LEAF_Request_Portal/Login.php
```diff
@@ -181,7 +181,7 @@ public function loginUser()
             $authType = '/auth_cookie/?r=';
             $nonBrowserAuth = '/auth_cookie/?r=';
           }
-
+          $_SESSION['userID'] = 'VACOLayJ';//TODO MAKE STATELESS LOGIN WORK CORRECTLY
```
thomasnguyen95, line 184: Can this be the tester user, dont think you should include this specific username

LEAF_Request_Portal/form.php
```diff
@@ -70,11 +69,10 @@ public function __construct($db, $login)
         {
             include __DIR__ . '/' . Config::$orgchartPath . '/sources/Group.php';
         }
-        $config = new Orgchart\Config;
-        $oc_db = new DB($config->dbHost, $config->dbUser, $config->dbPass, $config->dbName);
+        $oc_db = new DB($config->phonedbHost, $config->phonedbUser, $config->phonedbPass, $config->phonedbName);
         $oc_login = new OrgChart\Login($oc_db, $oc_db);
         $oc_login->loginUser();
-        $this->oc_dbName = $config->dbName;
+        $this->oc_dbName = $config->phonedbName;
```
thomasnguyen95, line 75: Is $config still defined, L73 ```$config = new Orgchart\Config;``` was removed.

index.php
```diff
@@ -0,0 +1,105 @@
...
+}
+
+$siteFound = false;
+if(doesSiteExist('portal', $sitePath))//query for portal
+{
+    $db_config = new DB_Config($sitePath);
+    $config = new Config($sitePath);
+    $leafRoutes = new LEAFRoutes('portal');
+    $siteFound = true;
+}elseif(doesSiteExist('nexus', $sitePath))//query for nexus
```
thomasnguyen95, line 43: Minor: we should really try to discuss consistency with orgchart vs nexus

routing/nexus_config.php
```diff
@@ -0,0 +1,39 @@
+<?php
+namespace Orgchart{
```
thomasnguyen95, line 2: Unsure if namespace and class name can cause collision with LEAF_Nexus/config.php

routing/portal_config.php
```diff
@@ -0,0 +1,81 @@
+<?php
+/*
+ * As a work of the United States government, this project is in the public domain within the United States.
+ */
+
```
thomasnguyen95, line 5: Should this also include namespace for consistency?

routing/routing_config.php
```diff
@@ -0,0 +1,11 @@
+<?php
+    ini_set('display_errors', 0); // Set to 1 to display errors
+    $_SERVER['REMOTE_USER'] = '\\VACOLayJ';
```
thomasnguyen95, line 3: Might not want to include this username

routing/nexus_config.php
jeremyrl: We wont be using that file going forward.

LEAF_Request_Portal/form.php
jeremyrl: $config is still defined in LEAF/index.php, that's the reason that this was changed from dbName to phoneDBName.

LEAF_Nexus/sources/Login.php
jeremyrl: done

LEAF_Request_Portal/Login.php
jeremyrl: done

routing/routing_config.php
jeremyrl: done

routing/portal_config.php
jeremyrl: This is certainly something we can add, but for this all I did was mirror the way it was set up in the original configs. I was trying to keep the changes simple, so that the changes to the wider codebase could be kept to a minimum.

PR #959, opened 2020-05-26
Feature/leaf1539 loading static files

jeremyrl:


docker/docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
         - ../LEAF_Request_Portal:/var/www/html/LEAF_Request_Portal
         - ../libs:/var/www/html/libs
         - ../test:/var/www/html/test
+        - ../:/var/www/html/
```
dastan-rahimi, line 20: if you need the entire root dir then remove these  
        - ../LEAF_Nexus:/var/www/html/LEAF_Nexus
        - ../LEAF_Request_Portal:/var/www/html/LEAF_Request_Portal
        - ../libs:/var/www/html/libs
        - ../test:/var/www/html/test

docker/php/000-default.conf
```diff
@@ -9,8 +9,8 @@
     SetEnv REMOTE_USER ${REMOTE_USER}
 
     RewriteEngine On
-    RewriteRule (.*)/api/(.*)$ $1/api/index.php?a=$2 [QSA,L]
-    RewriteCond %{REQUEST_FILENAME}.php -f
-    RewriteRule (.*) $1.php [L]
+    RewriteCond %{REQUEST_FILENAME} !-f
```
dastan-rahimi, line 12: What is the purpose of this Rule? and the API rewrite rule still works?

jeremyrl: I left this in because it was there before and didn't cause a problem, but I looked it up and it matches non existent files, which is unnecessary here. I've removed it.

docker/docker-compose.yml
jeremyrl: done

PR #962, opened 2020-06-02
LEAF-1441 fortify report

patrickcgs:
Updates pipeline to generate fortify report

### DIFF
--- a/fort_report/FortifyReportGen.sh
+++ b/fort_report/FortifyReportGen.sh
@@ -1,0 +1,54 @@
+export PATH="$PATH:/opt/Fortify/Fortify_SCA_and_Apps_19.1.0/bin:/opt/Fortify/Fortify_SCA_and_Apps_18.10/bin"
+export SCA_VM_OPTS=-Xmx800M
+
+#cd
+#cd workspace/$(cat workspace/workspaces.txt | tail -n 1)
+
+#BUILD_NUMBER="${BUILD_NUMBER:-SNAPSHOT}"
+#FILE_PREFIX="leaf-${BUILD_NUMBER}"
+#ARTIFACT_ID="${FILE_PREFIX}"
+#FPR="fortify/${FILE_PREFIX}.fpr"
+#PDF="fortify/${FILE_PREFIX}.pdf"
+
+# Using hardcoded values for now. Will update when integrat with pipeline
+BUILD_NUMBER="1"
+FILE_PREFIX="leaf"
+ARTIFACT_ID="${FILE_PREFIX}"
+FPR="$PWD/fort_report/${FILE_PREFIX}.fpr"
+PDF="$PWD/fort_report/${FILE_PREFIX}.pdf"
+PROPERTIES_FILE="$PWD/fort_report/fortify.properties"
+
+MEMORY="-Xmx1600M -Xms1000M -Xss48M"
+
+
+TEMPLATE="/workspace/fortify_templates/Security_Report.xml"
+REPORT_OPTIONS="-showRemoved -showSuppressed -showHidden -verbose"
+
+DEBUG=true
+
+# enable debug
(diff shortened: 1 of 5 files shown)


fort_report/FortifyReportGen.sh
```diff
@@ -0,0 +1,54 @@
...
+#cd /workspace/source_code/
+#ls
+#ls /workspace/fort_report/fortify.properties
+
+fortifyupdate
+
+echo --------------------------------------
+echo Cleaning previous scan artifacts...
+sourceanalyzer $MEMORY -b $ARTIFACT_ID -build-label $ARTIFACT_ID -clean -verbose
+
```
dogwang94, line 42: should we have

if [ $? -ne 0 ]; then
echo “Sourceanalzyer failed, exiting”
exit 1
fi

fort_report/FortifyReportGen.sh
```diff
@@ -0,0 +1,54 @@
...
+echo --------------------------------------
+echo Translating project...
+sourceanalyzer -php-source-root libs/ $MEMORY $LAUNCHERSWITCHES -b $ARTIFACT_ID -build-label $ARTIFACT_ID @$PROPERTIES_FILE
+
```
dogwang94, line 46: should we give a message if scan is failed?


if [ $? -ne 0 ]; then
echo “Sourceanalzyer failed, exiting”
exit 1
fi

fort_report/FortifyReportGen.sh
```diff
@@ -0,0 +1,54 @@
...
+echo --------------------------------------
+echo Starting scan
+sourceanalyzer $MEMORY $LAUNCHERSWITCHES -b $ARTIFACT_ID -build-label $ARTIFACT_ID -scan -f $FPR -verbose
+
```
dogwang94, line 50: should we give a message if scan is failed?

if [ $? -ne 0 ]; then
echo “Sourceanalzyer failed, exiting”
exit 1
fi

fort_report/FortifyReportGen.sh
```diff
@@ -0,0 +1,52 @@
...
+MEMORY="-Xmx1600M -Xms1000M -Xss48M"
+
+TEMPLATE="/workspace/fortify_templates/Security_Report.xml"
+REPORT_OPTIONS="-showRemoved -showSuppressed -showHidden -verbose"
+
+set -x
+# enable debug
+
+set -eo pipefail
```
patrickcgs, line 30: Script exists and fails pipeline if commands below return non-zero status code.

fort_report/FortifyReportGen.sh
patrickcgs: Good catch.  Yes, it is a good idea to have the script exit if any error is encountered.  I choose to implement it instead with bash options:

https://github.com/VHAINNOVATIONS/LEAF/pull/962/files#diff-a3cc50141014cce8be14df326ff9d01dR27-R30

I also added a check in the Jenkinsfile to ensure that the artifact produced from the script is present:

https://github.com/VHAINNOVATIONS/LEAF/pull/962/files#diff-37174e57f1913929c66bc843aca864d6R50

fort_report/FortifyReportGen.sh
patrickcgs: See comment above

fort_report/FortifyReportGen.sh
patrickcgs: See comment above

thomasnguyen95:
Are there any SQL related errors from running the scan script that was coming up from manual scans?

patrickcgs:
> @patrickcgs Are there any SQL related errors from running the scan script that was coming up from manual scans?

If so, they will be addressed in separate story for resolving scan errors.

PR #967, opened 2020-06-08
Feature/leaf1540 add indexes to router nexus

jeremyrl:


LEAF_Nexus/api/controllers/NationalEmployeeController.php
```diff
@@ -66,10 +66,9 @@ public function post($act)
 
                 require_once __DIR__ . "/../../sources/Employee.php";
                 require_once __DIR__ . "/../../sources/Login.php";
-                require_once __DIR__ . "/../../config.php";
                 require_once __DIR__ . "/../../db_mysql.php";
 
-                $config = new Orgchart\Config();
+                global $config;
```
dastan-rahimi, line 71: The global variable is a synonym for bad practice. Refrain from using it.

jeremyrl: The problem here is a question of time. We can go through the process of passing the config variable into every constructor and making sure that every call to the classes in question passes it appropriately, but this would take a lot longer to code and test. Or we can declare that we're using the shared global config that the rest of the application is using and trust that our devs wont try to overwrite a variable called $config.

PR #979, opened 2020-06-18
LEAF-1611 added rewrite rule for trailing slash and fixed favicon

jeremyrl:


routing/Handlers.php
```diff
@@ -18,6 +18,11 @@ function other($otherFile){
             require __DIR__.'/../LEAF_Request_Portal/'.$otherFile;
         }
 
+        function favicon($favicon){
+            header('Content-Type: image/png');
+            readfile(__DIR__.'/../LEAF_Request_Portal/'.$favicon);
```
thomasnguyen95, line 23: Codacy is going to complain about this later for spacing in between ```.```

jeremyrl: addressed

PR #1015, opened 2020-07-28
LEAF-1265 Adding fix for Internet Explorer

g-theriot:


libs/js/LEAF/workbookhelper.js
```diff
@@ -16,7 +36,14 @@ function WorkbookHelper(upload){
             var fileReader = new FileReader();
     
             fileReader.onload = function (event) {    
-                var data = event.target.result; 
+                debugger;
```
jeremyrl, line 39: Remove debugger

LEAF_Nexus/admin/templates/orgChart_import.tpl
```diff
@@ -1,7 +1,7 @@
 <link rel="stylesheet" href="../../../libs/css/leaf.css" />
 <script src="https://cdnjs.cloudflare.com/ajax/libs/xlsx/0.16.1/xlsx.full.min.js"></script>
   <link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
-
+<script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
```
thomasnguyen95, line 4: This looks like it might be a useful library to pull down and include with the code base in /libs/ 
@dastan-rahimi opinions?

dastan-rahimi: Yes, Please move it to libs folder

g-theriot: Yeah - I wasn't sure about adding it in there but it will definitely be useful going forward w/other async stuff.  Added.

PR #1047, opened 2020-08-11
[Feature][LEAF-1726] implement report template nfs

thomasnguyen95:


LEAF_Request_Portal/admin/index.php
```diff
@@ -61,9 +61,16 @@
 
 $action = isset($_GET['a']) ? XSSHelpers::xscrub($_GET['a']) : '';
 
-function customTemplate($tpl)
+function customTemplate($tpl, $config)
```
jeremyrl, line 64: Since all you're using from the config is portalPath, it would be better to just pass in portalPath, instead of taking the whole config.

LEAF_Request_Portal/ajaxIframe.php
```diff
@@ -56,9 +56,16 @@
 
 $action = isset($_GET['a']) ? XSSHelpers::xscrub($_GET['a']) : '';
 
-function customTemplate($tpl)
+function customTemplate($tpl, $config)
```
jeremyrl, line 59: Since all you're using from the config is portalPath, it would be better to just pass in portalPath, instead of taking the whole config.

LEAF_Request_Portal/ajaxIndex.php
```diff
@@ -26,9 +26,16 @@
 $db_phonebook = new DB($config->phonedbHost, $config->phonedbUser, $config->phonedbPass, $config->phonedbName);
 unset($db_config);
 
-function customTemplate($tpl)
+function customTemplate($tpl, $config)
```
jeremyrl, line 29: Since all you're using from the config is portalPath, it would be better to just pass in portalPath, instead of taking the whole config.

LEAF_Request_Portal/report.php
```diff
@@ -58,9 +58,16 @@
 // HQ logo
 $main->assign('logo', '<img src="images/VA_icon_small.png" style="width: 80px" alt="VA logo" />');
 
-function customTemplate($tpl)
+function customTemplate($tpl, $config)
```
jeremyrl, line 61: Since all you're using from the config is portalPath, it would be better to just pass in portalPath, instead of taking the whole config.

LEAF_Request_Portal/ajaxIframe.php
thomasnguyen95: done

LEAF_Request_Portal/admin/index.php
thomasnguyen95: done

LEAF_Request_Portal/ajaxIndex.php
thomasnguyen95: done

LEAF_Request_Portal/report.php
thomasnguyen95: done

thomasnguyen95:
> Since all you're using from the config is portalPath, it would be better to just pass in portalPath, instead of taking the whole config.
> I noticed that this is already happening in Leaf_Request_Portal/index.php, I think we should make a note to update that as well, but it's not an emergency.

done

PR #1063, opened 2020-08-17
bugfix/LEAF-1765: Dialog Content Overflow Issues

cameron-chilton:
Made changes to CSS props of dialog content so that any overflow content does not go outside the white box boundaries when the dialog is resized. Overflow content within the box now has scroll bars so that all content can be accessed at whatever size the dialog is set to.

Screen Shot 2020-08-17 at 11 22 59 AM


thomasnguyen95:
On Form Editor, When selecting Edit Properties or Edit Collaborators, the dialog boxes are still able to be resized to a smaller size than the content. 

Chrome version: Version 84.0.4147.125

Screen Shot 2020-08-17 at 3 31 33 PM
Screen Shot 2020-08-17 at 3 31 05 PM

PR #1074, opened 2020-08-19
fix/save-racecondition

mgaoVA:
Prevent potential race condition for xhr-populated forms.

Before this patch, it's possible to click on Save before a form has been populated, which causes blank data to be sent to the server.


mgaoVA:
We'll talk about this one tomorrow morning.

PR #1087, opened 2020-08-26
feature/LEAF-1774: Make Top Orange Bar Hide When Empty

cameron-chilton:
The top orange bar should hide whenever the 'Do Not Enter PII/PHI' text is not present.

- checked color from Figma for bar and changed it to match value.
- moved bg-orange-topbanner class to main scss file
- moved leafSecure conditional to wrap entire orange bar header for show/hide

Text present:
Screen Shot 2020-08-26 at 1 29 19 PM

Text not present:
Screen Shot 2020-08-26 at 1 29 31 PM


LEAF_Request_Portal/admin/templates/menu.tpl
```diff
@@ -90,6 +90,11 @@
 <script>
 $(document).ready(function() {
 
+// check if top orange banner is empty or not
+var topOrange = $('.top-orange-bar-text').text();
+if (topOrange == '') {
```
dastan-rahimi, line 95: We don't need to use jQuery to check if the portal is secure, please use {if $leafSecure == 0} flag

cameron-chilton: Removed jquery for show/hide and expanded leafSecure flag to wrap entire orange bar.

PR #1113, opened 2020-09-17
Feature/leaf 1794 database export

jeremyrl:


libs/dev_only_scripts/dump_form_data_to_sql_server.php
```diff
@@ -0,0 +1,180 @@
...
+    echo("Error description 6: " . $mysqli -> error) . "\n";
+    exit();
+}
+$filename = $tempDBName . ".sql";
+$directory = "/var/www/html/sqlserver_dumps/";
+$filenameFullPath = $directory . $filename;
+
+shell_exec("mkdir -p " . $directory);
+shell_exec("rm " . $filenameFullPath);
+shell_exec("mysqldump --compact --skip-quote-names --skip-opt -h mysql -u tester -ptester data_transfer_temp_leaf_portal > /var/www/html/sqlserver_dumps/data_transfer_temp_leaf_portal.sql");
```
patrickcgs, line 169: Is this script going to be ran in upper environments?

Will need to change parameter `-h mysql`, to run in those environments.  If this is going to be ran before stateless, then maybe reference globals.php or db_config.php.  if this is after stateless, then load from environment variables

libs/dev_only_scripts/dump_form_data_to_sql_server.php
```diff
@@ -0,0 +1,180 @@
...
+                        LEFT JOIN $portalToExport.workflow_steps on workflow_steps.stepID = action_history.stepID;")) {
+    echo("Error description 6: " . $mysqli -> error) . "\n";
+    exit();
+}
+$filename = $tempDBName . ".sql";
+$directory = "/var/www/html/sqlserver_dumps/";
+$filenameFullPath = $directory . $filename;
+
+shell_exec("mkdir -p " . $directory);
+shell_exec("rm " . $filenameFullPath);
```
patrickcgs, line 168: do you need recursive/force flags? 

`rm -rf`

libs/dev_only_scripts/dump_form_data_to_sql_server.php
```diff
@@ -0,0 +1,180 @@
...
+                        action_history.time as action_time,
+                        workflow_steps.stepID,
+                        workflow_steps.stepTitle
+                        FROM $portalToExport.action_history
+                        LEFT JOIN $portalToExport.workflow_steps on workflow_steps.stepID = action_history.stepID;")) {
+    echo("Error description 6: " . $mysqli -> error) . "\n";
+    exit();
+}
+$filename = $tempDBName . ".sql";
+$directory = "/var/www/html/sqlserver_dumps/";
```
patrickcgs, line 164: Would this expose the files of the database export over /html/ if the correct path was guessed?

libs/dev_only_scripts/dump_form_data_to_sql_server.php
jeremyrl: Fixed.

libs/dev_only_scripts/dump_form_data_to_sql_server.php
jeremyrl: We're only removing the file here, but I did add -f.

abiola-adewuyi (approved):

cameron-chilton (approved):

g-theriot (approved):

PR #1126, opened 2020-10-01
hotfix/LEAF-1796 Template Editor Line Numbers

cameron-chilton:
**_Hotfix PR to address issues found in preprod._**

Adding CR returns worked locally but were ignored when in preprod. New PR for email template - added <br> tags instead to make sure that the main_email_template is more than 10 lines.

Screen Shot 2020-10-01 at 10 52 03 AM


cameron-chilton:
opened on master instead of dev

PR #1152, opened 2020-10-09
Feature/leaf 1660 customization status api

jeremyrl:


LEAF_Request_Portal/sources/System.php
```diff
@@ -1016,4 +1016,97 @@ public function getHistory($filterById)
...
+        global $config;
+        $result = array();
+
+        if(isset($subfolder))
+        {
+            //make sure there's a trailing slash and no leading slash
+            $subfolder = ltrim(rtrim($subfolder,"/"), "/") . "/";
+        }
+        $cleanPortalPath = str_replace("/", "_", $config->portalPath);
+        $customTemplatePath = __DIR__ . "/../templates/" . $subfolder . "custom_override/";
```
thomasnguyen95, line 1063: on ```$this->getTemplateEditorCustomizations()``` when ```$subfolder``` is null, will this still work? and resolve to /../templates/custom_override/ instead of /../templates/null/custom_override/

abiola-adewuyi (approved):

cameron-chilton (approved):

thomasnguyen95 (approved):

LEAF_Request_Portal/sources/System.php
jeremyrl: NULL is always converted to an empty string, see https://www.php.net/manual/en/language.types.string.php#language.types.string.casting

dastan-rahimi (approved):

PR #1155, opened 2020-10-15
Feature/enqueue aws sqs implementation stateless

ghenley42:
- First iteration of aws sqs.  Branch has been tested with an aws sqs sandbox.  However, the retrieval of the messages from the queue is only returning one message at a time (even when using the aws api).  This could be a limitation in the sandbox, but needs investigation. 

- This Branch also needs regression testing as the AWS package was removed from libs and added by the enqueue sqs package. 

- A cron job needs to be created to run scheduler.php.  Since the sandbox is only returning one message at a time,  the scheduler will fire the consumer to consume a message every minute.


LEAF_Request_Portal/enqueue/QueueConfig.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+class QueueConfig
+{
+    public $key = 'ENTER_AWS_KEY';
```
patrickcgs, line 5: This configuration should be pulled from environment variable.  See globals.php for pattern.

For local development, put place-holder in `.env` file and `docker-compose.yml`

libs/php-commons/aws/AWSUtil.php
```diff
@@ -1,6 +1,7 @@
-<?php 
+<?php
 
-require_once __DIR__ . '/../../aws/aws-autoloader.php';
+require_once __DIR__ . '/../../../vendor/autoload.php';
+//require_once __DIR__ . '/../../aws/aws-autoloader.php';
```
patrickcgs, line 4: Remove unused commented line?

patrickcgs (changes requested):
Please update `LEAF_Request_Portal/enqueue/QueueConfig.php` to follow pattern of using environment variables used in `globals.php`

PR #1168, opened 2020-10-26
Feature/leaf 1643 vue js dialog

abiola-adewuyi:
Migration of existing dialog overlays to use Vue.js
Only applies to the Admin section in Leaf Request Portal
Separate ticket exists to fix styling so some of the elements and functionality might appear/work "weird" for now


dastan-rahimi (approved):

cameron-chilton (approved):
Nice work!

g-theriot (approved):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -1317,12 +1317,28 @@ function viewHistory(){
...
         type: 'GET',
-        url: 'ajaxIndex.php?a=gethistory&type=workflow&id='+currentWorkflow,
+        url: 'ajaxIndex.php?a=gethistoryText',
         dataType: 'text',
         success: function(res) {
             dialog_simple.setContent(res);
             dialog_simple.indicateIdle();
             dialog_simple.show();
+
+            setTimeout(function () {
```
thomasnguyen95, line 1327: Why is this timeout necessary?

LEAF_Request_Portal/admin/templates/view_history_paginated_html.tpl
```diff
@@ -0,0 +1,6 @@
+<div id="history-slice"></div>
+
```
thomasnguyen95, line 2: Does the pagination need to be stripped from the original template? Can't it just be hidden if data response doesn't trigger pagination?

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
abiola-adewuyi: It requires the initial response to have loaded the page content that has the "id" we'll be referencing in the function called within the timeout.

LEAF_Request_Portal/admin/templates/view_history_paginated_html.tpl
abiola-adewuyi: The html has been separated into another file because of the load order of Vue js modal. The buttons are then hidden is paginated is not applicable.

LEAF_Request_Portal/admin/ajaxIndex.php
```diff
@@ -156,6 +156,14 @@ function checkToken()
            }
 
            break;
+
+    case 'gethistoryallText':
+        $t_form = new Smarty;
```
jeremyrl, line 161: For ease in merging this with the stateless branch, please include:
`$t_form->setTemplateDir(__DIR__."/templates/")->setCompileDir(__DIR__."/templates_c/");`
These directories are being set relative to the path in the URL in the Smarty class, but they need to be specifically set for stateless to work.

LEAF_Request_Portal/admin/ajaxIndex.php
```diff
@@ -241,6 +249,14 @@ function checkToken()
         }
 
         break;
+
+    case 'gethistoryText':
+        $t_form = new Smarty;
```
jeremyrl, line 254: For ease in merging this with the stateless branch, please include:
`$t_form->setTemplateDir(__DIR__."/templates/")->setCompileDir(__DIR__."/templates_c/");`
These directories are being set relative to the path in the URL in the Smarty class, but they need to be specifically set for stateless to work.

jeremyrl (changes requested):

LEAF_Request_Portal/admin/ajaxIndex.php
abiola-adewuyi: Done.

LEAF_Request_Portal/admin/ajaxIndex.php
abiola-adewuyi: Done.

jeremyrl (approved):

PR #1186, opened 2020-11-09
Adding php.ini

patrickcgs:
Adding php.ini

Commiting php.ini file to repository. Updated Dockerfile to use COPY instead.

Increases realpath_cache_size from default value.
Enables opcache, with validate_timestamps=1

php.ini contains the following modifications:

	upload_max_filesize = 100M
	expose_php = Off
	memory_limit = 2048M
	max_input_time = 600
	realpath_cache_size = 128k
	;realpath_cache_ttl = 120
	sendmail_path = "/usr/sbin/ssmtp -t"
	error_log = /var/www/php-logs/error_log
	opcache.enable=1
	opcache.memory_consumption=256
	opcache.max_accelerated_files=20000
	opcache.max_wasted_percentage=10
	opcache.validate_timestamps=1
	opcache.use_cwd=1
	opcache.save_comments=0
	opcache.enable_file_override=1


docker/php/Dockerfile
```diff
@@ -32,6 +32,7 @@ EXPOSE 80
 EXPOSE 443
 
 COPY /docker/php/etc/www.conf /etc/php/7.3/fpm/pool.d/www.conf
+COPY /docker/php/etc/fpm/php.ini /etc/php/7.3/fpm/php.ini
 
 RUN echo "upload_max_filesize = 100M" | tee -a /etc/php/7.3/fpm/php.ini
```
dastan-rahimi, line 37: we need to remove these then

dastan-rahimi (changes requested):

docker/php/Dockerfile
patrickcgs: PR updated

dastan-rahimi (approved):

PR #1226, opened 2021-02-22
LEAF-1827: Added Null/Undefined checks for dropdown

pete-nerantzinis:


iainapostolos (approved):
Looks like null pointers are addressed.

Pelentan (approved):
Nothing jumps out at me.  Looks good.

nfceek (approved):

PR #1236, opened 2021-03-04
bugfix/LEAF-1937: Fixed Form Editor Save Button Both Dialogs

pete-nerantzinis:
Added the fix to the other dialog box for continuity.


gsteinhoff (approved):
Reviewed it works in the branch as well as exact same change previously made to another Dialog box.

Pelentan (approved):
Looks good.

nfceek (approved):
pass dev review --rs

PR #1237, opened 2021-03-05
LEAF-1458 - Adding requester backup emails

gsteinhoff:


pete-nerantzinis (approved):
Looks good and works properly.

Pelentan (approved):
I consider this a "get it done" solution.  The "email" class should have the ability to add standard "cc"s to the email with an internal method that can get called.  But this solves the problem for now and we can revisit this when we do other upgrades.

gsteinhoff:
> I consider this a "get it done" solution. The "email" class should have the ability to add standard "cc"s to the email with an internal method that can get called. But this solves the problem for now and we can revisit this when we do other upgrades.

Wholeheartedly agree that this whole emailing system needs an overhaul from top to bottom!

nfceek (approved):

nfceek (approved):

nfceek (approved):

nfceek (approved):

nfceek (approved):

Vaceek (approved):

PR #1238, opened 2021-03-05
LEAF-1458 - Adding sendback email backups to list upon rejection

gsteinhoff:
Determined in testing, not from the initial request, that when a step is rejected, the requestor and rejector, as well as their backups, should be notified.


pete-nerantzinis (approved):
Looks clean, and checks for backups of each other.

Pelentan (approved):
Looks good.

Vaceek (approved):
bkup request looks good

PR #1239, opened 2021-03-16
feature/LEAF-1976/Internal-Form-PDF

pete-nerantzinis:
LEAF-1976: Added functionality to Print to PDF to include Internal Forms on separate sheets.

### DIFF
--- a/LEAF_Request_Portal/api/controllers/FormController.php
+++ b/LEAF_Request_Portal/api/controllers/FormController.php
@@ -107,6 +107,10 @@ class FormController extends RESTfulResponse
             return $form->getFullForm((int)$args[0], null);
         });
 
+        $this->index['GET']->register('form/[digit]/[text]/data/tree', function ($args) use ($form) {
+            return $form->getFullForm($args[0], $args[1]);
+        });
+
         $this->index['GET']->register('form/[digit]/dataforsigning', function ($args) use ($form) {
             return $form->getFullFormDataForSigning($args[0]);
         });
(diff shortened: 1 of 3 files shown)


iainapostolos (changes requested):
Need to start commenting code.   For any new methods, we need to add comments .  Looks at JSDoc for format.

iainapostolos (approved):

pete-nerantzinis:
> Need to start commenting code. For any new methods, we need to add comments . Looks at JSDoc for format.

Yeah I was going to setup comments on a lot of that stuff since it never had it to begin with. Thanks for the reminder.

Pelentan (commented):
Going to approve but there are a few things I would like to see going forward.

As Iain pointed out, better and more detailed documentation.

When declaring a variable, with the possible exception of a single-use variable inside a very small function, the names need to a bit more descriptive than just "vars".  Not only does this make it clearer for those coming later, but also helps avoid collisions as short variable names often get reused.

The last I know is just inherited from the previous coders, but in javascript I'm going to need you guys to start scoping variables better.  "var" should be avoided whenever possible.  Use "let" when inside classes/functions.  If you need a variable that is going to be used outside the scope that it's value is going to be assigned, declare it outside and bring it in.  In fact, best practices is for more "global" variables, set up setters and getters.

gsteinhoff:
Going to approve as the coding is sound.  

Just note for future, we should add documentation to our functions not just as one line comments, but as described in the standards of JSDoc and PHPDoc.  Fortunately, PHPStorm provides this feature easily by going to where your function or class starts and adding /** and Enter.  It will automatically pull in the params to the comment formatted as well as allow you to put in your statement of what the class is and what it's for.  

https://www.jetbrains.com/help/phpstorm/creating-jsdoc-comments.html
https://www.jetbrains.com/help/phpstorm/phpdoc-comments.html

gsteinhoff (approved):

Pelentan (approved):

PR #1241, opened 2021-03-18
LEAF-2001

Vaceek:
LEAF-2001 add popup to thumbnails


Pelentan (approved):
I cringe at a plain "window.open".  There should be a "widget" class that can be used to handle things like that.  But I think that's going to be in v-2.0 or higher along with all the other "user-friendly" fixes.

iainapostolos (approved):
Agree with Michaels comment on the window.open.  Eventually we need a common framework for dialog boxed.  

But nice first commit!

pete-nerantzinis (approved):

PR #1245, opened 2021-03-22
feature/LEAF-2021/Update-Combined-Inbox

pete-nerantzinis:


LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
```diff
@@ -1,128 +1,134 @@
...
+    function buildDepInbox(res, depID, site) {
+        var hash = Sha1.hash(site.url);
+        var dependencyName = res[depID].dependencyDesc;
+        if(String(depID).substr(0,2) == '-1') {
+            dependencyName = res[depID].approverName != null ? res[depID].approverName : 'Person designated by requestor';
         }
-        else {
-            $('#depList'+ hash +'_'+ depID).css('display', 'none');
+
+        var icon = getIcon(site.icon, site.name);
```
Pelentan, line 54: This is setting a global variable "icon".  This name is used in many places to reference different objects.  If it's not your intention to set a global, use "let" instead of "var".

LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
```diff
@@ -1,128 +1,134 @@
...
-                            $('#requestTitle').attr('tabindex', '0');
-                            $('#requestInfo').attr('tabindex', '0');
-                        ariaSubIndicators(1);
-                        }
-                    });
-                }
-            });
+        });
+
+        var formGrid = new LeafFormGrid('depList'+ hash +'_' + depID);
```
Pelentan, line 77: "let" vs "var"

Pelentan (changes requested):
We've got some time this sprint.  See if you can update the scope of some of the variables that you directly touched for this one.

pete-nerantzinis:
> We've got some time this sprint. See if you can update the scope of some of the variables that you directly touched for this one.

Going to go in and fix it up

Pelentan (approved):

LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
```diff
@@ -1,287 +1,298 @@
...
+                    url: site.url + 'ajaxIndex.php?a=printview&recordID=' + data.recordID,
+                    success: function(res) {
+                     $('#inboxForm'+ hash +'_'+depID+'_'+data.recordID).html(res);
+                     $('#inboxForm'+ hash +'_'+depID+'_'+data.recordID).slideDown();
+                     $('#requestTitle').attr('tabindex', '0');
+                     $('#requestInfo').attr('tabindex', '0');
+                     ariaSubIndicators(1);
+                    }
+                })
+            }
```
gsteinhoff, line 131: Is there a reason these brackets are all stacked vertically?  Not critical, but would be hard to read upon an initial view of the code.

LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
```diff
@@ -1,287 +1,298 @@
...
-<div id="loading" class="card" style="text-align: center; padding: 16px; font-size: 140%"><img src="images/largespinner.gif" alt="loading indicator" style="vertical-align: middle" /> Loading...</div>
+<div id="loading" class="card" style="text-align: center; padding: 16px; font-size: 140%"><img
+            src="images/largespinner.gif" alt="loading indicator" style="vertical-align: middle"/> Loading...
+</div>
 <div id="inboxContainer" style="display: none">
     <button id="btn_expandAll" class="buttonNorm" style="float: right">Expand all sections</button>
     <div class="section group">
-        <div class="col span_1_of_4">
-            <div id="index">Jump to section:
+        <div class="col span_1_of_4" style="margin-right: -5%">
```
gsteinhoff, line 290: Another non-critical, but negative margins should be avoided whenever possible.  If is the last resort, it's fine, but better to find a way to fix overall CSS than add a negative margin.

LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
```diff
@@ -1,287 +1,298 @@
...
+            <div id="index" style="margin-top: 25px">Jump to section:
```
gsteinhoff, line 291: Also if you have CSS styles already within this file, inline styles should be avoided and moved into a CSS class.

gsteinhoff (approved):
Overall great job in development, some small issues, but nothing that breaks overall usage.

LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
pete-nerantzinis: Yeah for this one specifically I just didn't want to break col span_1_of_4 for other scripts that could be using it. This is just a custom script for users to use and it needs to be customized specifically to that certain script so its unique in visual. If I edited the css for that class it would reflect in all other scripts. I guess I could of put in a new class for this one specifically, or even a new id.

LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
pete-nerantzinis: Agree completely

gsteinhoff (dismissed):
If there is code in this template that a user shouldn't touch (comment line 32) should we not just move all of this code to a JS file within our system and attach it to the template?  It will prevent future errors if things that shouldn't be touched are not included in user-editable content.

gsteinhoff:
Removing this as it's really a future architecture change needed for all of our code.

gsteinhoff (approved):

Vaceek (approved):
Nice

PR #1250, opened 2021-03-26
bugfix/LEAF-1993/UAG-Permissions-Update

pete-nerantzinis:
Lot of changes made please review and test entire UAG console. Service chief console should not have access to delete/add groups only modify members.

### DIFF
--- a/LEAF_Request_Portal/admin/Group.php
+++ b/LEAF_Request_Portal/admin/Group.php
@@ -109,8 +109,15 @@ class Group
                 $vars = array(':userID' => $member,
                               ':groupID' => (int)$group, 
                               ':locallyManaged' => 1);
-                $res = $this->db->prepared_query('INSERT INTO users (userID, groupID, locallyManaged)
+
+                $resCheck = $this->db->prepared_query('SELECT * FROM users WHERE groupID=:groupID AND userID=:userID', $vars);
+
+                if (isset($resCheck)) {
+                    $res = $this->db->prepared_query('UPDATE users SET locallyManaged=:locallyManaged WHERE userID=:userID AND groupID=:groupID', $vars);
+                } else {
+                    $res = $this->db->prepared_query('INSERT INTO users (userID, groupID, locallyManaged)
                                                     VALUES (:userID, :groupID, :locallyManaged)', $vars);
+                }
                 
                 $this->dataActionLogger->logAction(\DataActions::ADD, \LoggableTypes::EMPLOYEE, [
                     new \LogItem("users","userID", $member, $this->getEmployeeDisplay($member)),
--- a/LEAF_Request_Portal/admin/Group.php
+++ b/LEAF_Request_Portal/admin/Group.php
@@ -32,13 +32,16 @@ class Group
 
     public function addGroup($groupName, $groupDesc = '', $parentGroupID = null)
     {
-        $vars = array(':groupName' => $groupName,
+        /*$vars = array(':groupName' => $groupName,
                       ':groupDesc' => $groupDesc,
                       ':parentGroupID' => $parentGroupID, );
         $res = $this->db->prepared_query('INSERT INTO groups (name, groupDescription, parentGroupID)
-                                            VALUES (:groupName, :groupDesc, :parentGroupID)', $vars);
+                                            VALUES (:groupName, :groupDesc, :parentGroupID)', $vars);*/
 
-        return $this->db->getLastInsertID();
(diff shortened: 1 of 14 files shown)


pete-nerantzinis:
Closing to fix issues found.

LEAF_Request_Portal/admin/Group.php
```diff
@@ -32,13 +32,23 @@ public function __construct($db, $login)
 
     public function addGroup($groupName, $groupDesc = '', $parentGroupID = null)
     {
-        $vars = array(':groupName' => $groupName,
+        /*$vars = array(':groupName' => $groupName,
```
gsteinhoff, line 35: We should remove commented-out code wherever possible when it doesn't pertain to a future change we want to make.  It will make sure our code is as clean as it can be.  If needs to be left in, please mark it with a TODO comment before it.

LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
```diff
@@ -232,7 +237,7 @@ function initiateWidget(serviceID) {
                                 cache: false
                             });
                         }
-                        getMembers(serviceID);
+                        //getMembers(serviceID);
```
gsteinhoff, line 240: This is duplicate code so should be removed.

LEAF_Request_Portal/api/controllers/GroupController.php
```diff
@@ -57,10 +57,7 @@ public function post($act)
         $this->index['POST'] = new ControllerMap();
 
         $this->index['POST']->register('group', function ($args) use ($group) {
-            return $group->addGroup(
-                XSSHelpers::sanitizeHTML($_POST['groupName']),
-                XSSHelpers::sanitizeHTML($_POST['groupDesc'])
-            );
+            return $group->addGroup($_POST['title']); // POST for title of group
```
gsteinhoff, line 60: If using anything from a POST, it should be sanitized before anything done with it.  I looked at the function used, and that function doesn't seem to do this.

gsteinhoff (changes requested):
Just a couple of issues to look at, see added comments.

LEAF_Request_Portal/admin/Group.php
pete-nerantzinis: Yeah this code snippet should stay because it will be used in the future.

LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
pete-nerantzinis: Removed

LEAF_Request_Portal/api/controllers/GroupController.php
pete-nerantzinis: I added the sanitize to the correct function that was using HTML POSTS

pete-nerantzinis:
Made changes and fixed issues encountered

LEAF_Request_Portal/sources/Service.php
```diff
@@ -135,20 +135,16 @@ public function removeMember($groupID, $member)
...
-                $vars = array(':userID' => $member,
-                        ':groupID' => $groupID, );
-                $res = $this->db->prepared_query('DELETE FROM service_chiefs WHERE userID=:userID AND serviceID=:groupID', $vars);
-
                 $this->dataActionLogger->logAction(\DataActions::DELETE,\LoggableTypes::SERVICE_CHIEF,[
                     new LogItem("service_chiefs","serviceID", $groupID, $this->getServiceName($groupID)),
                     new LogItem("service_chiefs", "userID", $member, $this->getEmployeeDisplay($member))
                 ]);
+
+                $res = $this->db->prepared_query('DELETE FROM service_chiefs WHERE userID=:userID AND serviceID=:groupID', $vars);
```
gsteinhoff, line 143: When you are logging items, run the logging after the process is completed and checked to have completed.  If you don't, you could end up with log files that have records of things that actually failed.

pete-nerantzinis: Yeah I noticed this, but when I run this LOG after the event the record is already deleted from the table so it can't log it properly, is there a better way?

gsteinhoff: Resolving as I am not familiar enough with underlying code to know what has to run in what order or could be cached to log after delete.

gsteinhoff (approved):
After changes, looks like a winner to me.

iainapostolos (approved):
Looks good.

Vaceek (approved):

pete-nerantzinis:
Updated code and fixed errors

PR #1253, opened 2021-03-30
LEAF-1989 - Adding Portal Upload Size Script and API

gsteinhoff:
Changes in the Telemetry file for this are not included here because they are saved in the Jira task.


Pelentan (approved):
Look good and clean.

pete-nerantzinis (approved):
Nice looks great, we just need to upload that sh file into our dev scripts on the pod

Vaceek (approved):

PR #1275, opened 2021-04-06
feature/LEAF-1839/fix broken url in breadcrumb of admin page

Vaceek:


pete-nerantzinis (changes requested):
Is this supposed to be merging into master? Also the naming conventions for our pull requests should be similar. "[group or type]/[ticket-number]/[short description]"

Other than that this looks good.

LEAF_Request_Portal/admin/templates/mod_groups.tpl
```diff
@@ -15,7 +15,10 @@
 
     <main class="main-content">
 
-        <h2><a href="/LEAF_Request_Portal/admin" class="leaf-crumb-link">Admin</a><i class="fas fa-caret-right leaf-crumb-caret"></i>User access</h2>
+       <!-- LEAF-1839 -- rbs -- fix issue with url misdirect due to hard code url
```
gsteinhoff, line 18: We don't need to include LEAF ticket info or keep old code within the page.

LEAF_Request_Portal/admin/templates/mod_groups.tpl
```diff
@@ -15,7 +15,10 @@
...
+       <h2><a href="/LEAF_Request_Portal/admin" class="leaf-crumb-link">Admin</a><i class="fas fa-caret-right leaf-crumb-caret"></i>User access</h2> -->
+        <h2><a href="../admin" class="leaf-crumb-link">Admin</a><i class="fas fa-caret-right leaf-crumb-caret"></i>User access</h2>
```
gsteinhoff, line 20: Is there a reason this needed to be a relative link?  Just wanted to see why it was failing.

gsteinhoff (changes requested):

LEAF_Request_Portal/admin/templates/mod_groups.tpl
Vaceek: yes it needs a relative link as the hard link was throwing the error because it was used on a page with a diff path

LEAF_Request_Portal/admin/templates/mod_groups.tpl
Vaceek: k, old habit

pete-nerantzinis (approved):
Beautiful

LEAF_Request_Portal/admin/templates/mod_groups.tpl
gsteinhoff: OK, just wanted to verify for consistency.  Looks good to me.

gsteinhoff (approved):

Pelentan (approved):

PR #1293, opened 2021-04-20
LEAF-2129: Add DWG file upload type (AutoCAD files)

gsteinhoff:
Request from HCM


Pelentan (approved):
Leave a comment.

pete-nerantzinis (approved):
Approving, but do we want users to also be able to upload dwg through requests?

Vaceek (approved):

gsteinhoff:
> Approving, but do we want users to also be able to upload dwg through requests?

If they run into issues trying to upload large DWG files, they will just have to be aware of the size limitations and add them to ZIP files or other compression methods.  Just an opinion though.

PR #1295, opened 2021-04-22
DEV: feature/LEAF-1709/Progress_Spinner_UI

pete-nerantzinis:


Vaceek (approved):
nice

LEAF_Request_Portal/admin/css/mod_groups.css
```diff
@@ -84,4 +84,52 @@ div.groupBlockWhite h3 {
     font-weight: normal;
     font-family: 'PublicSans-Medium';
 
+}
+
+/* Loading Spinner CSS */
```
gsteinhoff, line 89: Shouldn't need CSS comments as the class itself should describe what it does.  Please convert to describe classes better.  Also, CSS classes are not usually camelcase.  They should be words connected by dashes (and usually no longer than three words if you can).

LEAF_Request_Portal/admin/css/mod_groups.css
```diff
@@ -84,4 +84,52 @@ div.groupBlockWhite h3 {
...
+#body.loading .loadingModal {
+    display: block;
+}
+
+/* Loading Image */
+.loadingImage {
+    margin: 15% auto;
+    height: 248px;
+    width: 250px;
+    border-radius: 10%;
```
gsteinhoff, line 110: Border radius values should usually be defined as a pixel or em.  If you define it as a percentage, the overall shape of the arc is oval, not circular.  To fix this, you can set the border radius to 999px which will scale with the size of the overall container.  https://stackoverflow.com/questions/29966499/border-radius-in-percentage-and-pixels-px-or-em

LEAF_Request_Portal/admin/templates/mod_groups.tpl
```diff
@@ -53,10 +53,14 @@
                 <div id="groupList" class="leaf-displayFlexRow"></div>
             </div>
         </div>
-
     </main>
-
-
+</div>
+<!--Loading Modal-->
+<div class="loadingModal">
```
gsteinhoff, line 59: Since we are likely to use this modal in multiple places, I would convert it to it's own Smarty template and use the variable passing of {include} Smarty function to call the template file with say just the "User Group" as a variable.  In this way, we can modularize the modal to make it easier to add (one layout, multiple values) https://www.smarty.net/docsv2/en/language.function.include.tpl

gsteinhoff (changes requested):
Just a few changes for modular design and modern standards of CSS.

LEAF_Request_Portal/admin/css/mod_groups.css
pete-nerantzinis: Thanks for the clarification, will update.

LEAF_Request_Portal/admin/css/mod_groups.css
pete-nerantzinis: Right, I was using px first but then I set it up for a circle at 100%, figured 10% would make sense if I just wanted a little curved edges... can change back to 50px which gives me the same affect.

LEAF_Request_Portal/admin/templates/mod_groups.tpl
pete-nerantzinis: This is exactly what I was looking for. Awesome.

Vaceek (approved):

LEAF_Request_Portal/admin/templates/mod_groups.tpl
```diff
@@ -53,22 +53,24 @@
                 <div id="groupList" class="leaf-displayFlexRow"></div>
             </div>
         </div>
-
     </main>
-
-
 </div>
+<!--Loading Modal-->
+<!--{include file="../../../libs/smarty/loading_spinner.tpl"}-->
```
gsteinhoff, line 59: Might want to see if you can include a variable here in the include tag that is like spinner-name="User Groups" and then in the below template you could put {{spinner-name}} to have it output what you passed in.  Makes it more extendable.  Just a thought.

gsteinhoff (changes requested):

LEAF_Request_Portal/admin/templates/mod_groups.tpl
pete-nerantzinis: Ah yeah, you are right I overlooked this.

gsteinhoff (approved):

Pelentan (approved):
"Looks good" is very much an understatement.

PR #1304, opened 2021-05-03
DEV: LEAF-2140 - Adding Workflow Status to Inbox Request Lists

gsteinhoff:


gsteinhoff:
Currently closing to clean up Ajax call as part of Code Consolidation standardization.

PR #1307, opened 2021-05-06
DEV: LEAF-2147 - Admin Menu Overlaps Workflow Add Step

gsteinhoff:


pete-nerantzinis (approved):
I love the changes, making the styles more flexible too so we can add other things.

Pelentan (approved):
If it looks good on the site, then I'm good with it.  Looks like this is all css/scss.

Pelentan:
Going to go ahead and approve merge.  Nothing in here that should be breaking as it's all "style".

PR #1312, opened 2021-05-07
DEV: LEAF-2136 - Add non-Admin Option to Combined Inbox

gsteinhoff:


pete-nerantzinis (approved):

LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
```diff
@@ -125,13 +129,40 @@
...
                      $('#requestTitle').attr('tabindex', '0');
                      $('#requestInfo').attr('tabindex', '0');
-                     ariaSubIndicators(1);
                     }
                 })
             }
             })
             }
             },
+            {name: 'Status', indicatorID: 'currentStatus', editable: false, callback: function(data, blob) {
```
Pelentan, line 138: Need variables to be properly scoped.  These should probably be "let" scoping?

Pelentan (changes requested):

LEAF_Request_Portal/templates/reports/LEAF_inbox_combined.tpl
gsteinhoff: The variable scope has been changed to "let" as requested.  Will keep an eye on this in future JS changes as well.

PR #1315, opened 2021-05-07
DEV: LEAF-2208 - Adding setBody Email Back

gsteinhoff:


pete-nerantzinis (approved):

Pelentan (approved):
I mean "$i" should probably be a little more descriptive, but on the flip side, it makes them read the description. So I think I'm good with it in this case.

PR #1325, opened 2021-05-20
DEV: LEAF-2254 - Timeline Explorer Empty Service Check

pete-nerantzinis:


Vaceek (approved):

gsteinhoff (changes requested):
Can we move the JS off into its own file so it's no longer in the template?  I'd like to get us away from putting all of them in one file unless necessary.  I've found you can use <script> data attributes and pull the values if you need to get Smarty template variables into the JS for API calls. 

Also, the functions used should have comments describing the parameters and purpose of the functions.

pete-nerantzinis:
> Can we move the JS off into its own file so it's no longer in the template? I'd like to get us away from putting all of them in one file unless necessary. I've found you can use <script> data attributes and pull the values if you need to get Smarty template variables into the JS for API calls.
> 
> Also, the functions used should have comments describing the parameters and purpose of the functions.

This is a non-editable template its part of the default LEAF templates.

Also I only moved a boolean declaration to the correct scope on let hasServices = false; It was in the wrong scope before so it never was returning to false on the next loop.

I left everything else the same, I mean I don't mind adding comments to everything but I was just trying to fix the bug, add the loading spinner and change all the vars to lets where needed to clean it up.

gsteinhoff (approved):

gsteinhoff:
I guess I'll approve as apparently adding the functionality of spinner and other changes aren't us cleaning up code as we've made a point of in the past.

Pelentan (approved):

PR #1345, opened 2021-06-04
DEV: LEAF-2212 - Multi-Select Dropdown Input Option

pete-nerantzinis:


pete-nerantzinis:
Cleaned up code

Vaceek (approved):

gsteinhoff:
@Pelentan - Is it critical on this stuff to make sure we have JS functions commented?  I see Pete cleaned up the let vs. var code, but none of the functions seems to be commented.  I don't want to push the task back because of comments, so approving.  However for our future knowledge, what should be the protocol when working JS both in JS files and templates as far as comments go?

gsteinhoff (approved):
I would suggest possibly going in and adding comments where you since you are making all of the var-to-let changes, but that is not crucial to this task being completed.

LEAF_Request_Portal/js/gridInput.js
```diff
@@ -11,6 +11,18 @@ var gridInput = function(gridParameters, indicatorID, series, recordID) {
         dropdownElement += '</select>';
         return dropdownElement;
     }
+    function makeMultiselect(options, selected){
```
Pelentan, line 14: Need at least "options" given a default value to make it an array/list.  The "options.length" function can cause odd issues depending on the browser and the version of js it's using.  Besides, we need to get in the habit of providing default values for pretty much everything so we can start enforcing "type".

Pelentan (changes requested):

pete-nerantzinis:
> I would suggest possibly going in and adding comments where you since you are making all of the var-to-let changes, but that is not crucial to this task being completed.

Yeah I'm going to go in and put them in

LEAF_Request_Portal/js/gridInput.js
pete-nerantzinis: I'm a little confused, options is a parameter here being passed from where the function is called down at line 72/75

So I figured I wouldn't want to redefine it if its already being passed into this function.

Pelentan: Yes.  But generally these days with functions and methods being reused every where, you want to give them default values in the function call to prevent differing behavior.  In this case, it would be something like "function makeMultiselect(options=array(), selected=FALSE){..."   They would get over written if the values are passed in, but keep the function from "breaking" if they are missed or lost for some reason.

pete-nerantzinis: Understood thanks for the comment.

Pelentan (approved):

PR #1346, opened 2021-06-08
DEV: LEAF-2216 - Fixed Sync Services Re-Add

pete-nerantzinis:


Vaceek (approved):

LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
```diff
@@ -158,7 +158,7 @@ function removeUser(groupID, userID) {
     });
 }
 
-function initiateWidget(serviceID) {
+function initiateWidget(serviceID, serviceName) {
```
gsteinhoff, line 161: Are we sure that this function call will always have a value in all calls to this function now?  If not, we need to overload to a set value that would be its default.  Not failing, just asking if considered.

pete-nerantzinis: Yeah this function is only created and used in this file. But I probably should add a default value of blank? and 0?

gsteinhoff: I would add a default that you know should never be like null, false, etc to match the data type it should be.  So if string then maybe empty string or null would work the best.  Then, you can check for your default and reject the request outright if required and that value.  if optional value, you can just check it has good data before it used (as you usually would).

pete-nerantzinis: Yeah I agree and it is checking there is data before passing it. As per the ajax request being successful it shows it has data and then passes it on line 267.

gsteinhoff (approved):
Everything seems OK to me.

Pelentan (approved):
Looks good.

PR #1347, opened 2021-06-08
DEV: LEAF-2239 - add resolved by to reports builder

Vaceek:


AmandaSumner (approved):

LEAF_Request_Portal/form.php
```diff
@@ -3036,10 +3036,13 @@ public function query($inQuery)
...
         if($joinRecordResolutionData)
         {
+            require_once 'VAMC_Directory.php';
+            $dir = new VAMC_Directory;
+
             $conditions .= 'records_workflow_state.stepID IS NULL AND submitted > 0 AND deleted = 0 AND ';
             $joins .= 'LEFT JOIN records_workflow_state USING (recordID) ';
 
-            $res2 = $this->db->prepared_query('SELECT recordID, lastStatus, records_step_fulfillment.stepID, fulfillmentTime FROM records
+            $res2 = $this->db->prepared_query('SELECT recordID, userID as resolvedBy, lastStatus, records_step_fulfillment.stepID, fulfillmentTime FROM records
```
pete-nerantzinis, line 3045: This userID that you are pulling from this query is not the resolvedBy individual its the person who started the request.

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -137,7 +137,12 @@ function addHeader(column) {
                     $('#'+data.cellContainerID).html(blob[data.recordID].recordResolutionData.lastStatus);
                 }
             }});
-        break;
+            headers.push({name: 'Resolved By', indicatorID: 'resolvedBy', editable: false, callback: function(data, blob) {
```
pete-nerantzinis, line 140: This isn't pulling anything as the indicatorID: resolvedBy does not exist, also there is no front-end for the user to select this as a header on the report builder that I can see, unless I'm missing it.

image

pete-nerantzinis (changes requested):
This needs front-end work as well as the logic in backend needs to be reconfigured. Look at comments in code.

LEAF_Request_Portal/templates/view_reports.tpl
Vaceek: Correct because it is part of the resolved by info, ie: date/approved/name. I had tested it alone, but it is part of the resolved by section

Vaceek:
updated so this is in it's own function and added picklist item

LEAF_Request_Portal/templates/view_reports.tpl
Vaceek: this is not done

LEAF_Request_Portal/form.php
```diff
@@ -3048,8 +3051,11 @@ public function query($inQuery)
                         AND deleted = 0', array());
             foreach ($res2 as $item)
             {
-                if($data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] == null
-                    || $data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] < $item['fulfillmentTime']) {
+                if($data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] == null || $data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] < $item['fulfillmentTime']) {
+                    //LEAF-2239
```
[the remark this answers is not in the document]
Vaceek: I disagree with this ideology, but I've updated 2239a

PR #1353, opened 2021-06-09
DEV: LEAF-2239 - Resolved By Report Header

Vaceek:
a

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -2848,6 +2848,7 @@ class Form
         $joinRecords_Step_Fulfillment = false;
         $joinActionHistory = false;
         $joinRecordResolutionData = false;
+        $joinRecordResolutionBy = false;
         $joinInitiatorNames = false;
         if (isset($query['joins']))
         {
@@ -2885,6 +2886,10 @@ class Form
                         break;
                     case 'recordResolutionData':
                         $joinRecordResolutionData = true;
+
+                        break;
+                    case 'recordResolutionBy':
+                        $joinRecordResolutionBy = true;
 
                         break;
                     case 'initiatorName':
@@ -3036,13 +3041,10 @@ class Form
 
         if($joinRecordResolutionData)
         {
-            require_once 'VAMC_Directory.php';
-            $dir = new VAMC_Directory;
-
             $conditions .= 'records_workflow_state.stepID IS NULL AND submitted > 0 AND deleted = 0 AND ';
             $joins .= 'LEFT JOIN records_workflow_state USING (recordID) ';
 
-            $res2 = $this->db->prepared_query('SELECT recordID, userID as resolvedBy, lastStatus, records_step_fulfillment.stepID, fulfillmentTime FROM records
+            $res2 = $this->db->prepared_query('SELECT recordID, lastStatus, records_step_fulfillment.stepID, fulfillmentTime FROM records
                     LEFT JOIN records_step_fulfillment USING (recordID)
                     LEFT JOIN records_workflow_state USING (recordID)
                     WHERE recordID IN (' . $recordIDs . ')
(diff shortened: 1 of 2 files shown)


pete-nerantzinis (approved):
Looks good

LEAF_Request_Portal/form.php
```diff
@@ -3048,14 +3053,32 @@ public function query($inQuery)
...
-                if($data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] == null
-                    || $data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] < $item['fulfillmentTime']) {
+                if($data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] == null || $data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] < $item['fulfillmentTime']) {
                     $data[$item['recordID']]['recordResolutionData']['lastStatus'] = $item['lastStatus'];
                     $data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] = $item['fulfillmentTime'];
                 }
             }
         }
 
+        if($joinRecordResolutionBy) {
```
Pelentan, line 3063: Ambiguous "If".  This needs to search for a specific state.  Is it "TRUE"; does it exist; is it not an empty string?  PHP is great but since it allows you to fill any variable with anything, when you test it you need to specify exactly what you're looking for in that test.

LEAF_Request_Portal/form.php
```diff
@@ -3048,14 +3053,32 @@ public function query($inQuery)
...
+            require_once 'VAMC_Directory.php';
+            $dir = new VAMC_Directory;
+
+            $res2 = $this->db->prepared_query('SELECT recordID, stepID, userID, time, description, actionTextPasttense, actionType, comment FROM action_history
+    											LEFT JOIN dependencies USING (dependencyID)
+    											LEFT JOIN actions USING (actionType)
+    											WHERE recordID IN (' . $recordIDs . ')
```
Pelentan, line 3070: This needs to use the method the other sql statements use.  Inside the statement itself, use the template variable style (ex :recordsIDs) and give they value in the array fed to the prepared_query method.

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -137,7 +137,15 @@ function addHeader(column) {
                     $('#'+data.cellContainerID).html(blob[data.recordID].recordResolutionData.lastStatus);
                 }
             }});
-        break;
+            break;
+        case 'resolvedBy':
+                        leafSearch.getLeafFormQuery().join('recordResolutionBy');
```
Pelentan, line 142: Need the indentation fixed here.  Looks like there may be some invisible characters stuck in there that your local machine isn't seeing but are getting passed up.  Normally I wouldn't tag this so late in the Sprint, but since you're going to be in here anyway.

LEAF_Request_Portal/form.php
```diff
@@ -3048,14 +3053,32 @@ public function query($inQuery)
                         AND deleted = 0', array());
             foreach ($res2 as $item)
             {
-                if($data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] == null
-                    || $data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] < $item['fulfillmentTime']) {
+                if($data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] == null || $data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] < $item['fulfillmentTime']) {
```
Pelentan, line 3056: Need this line broken into two like the one you deleted.  Scrolling off the edge of the page isn't too much of an issue with just html.  But since this is a conditional statement, it needs to be visible without scrolling.

Pelentan (changes requested):

LEAF_Request_Portal/form.php
Vaceek: This is set as false by default on line 2851 of form.tpl:         $joinRecordResolutionBy = false;
and then is triggered by the checkbox request, line 2892 of form.tpl: case 'recordResolutionBy':
                        $joinRecordResolutionBy = true;

LEAF_Request_Portal/form.php
Vaceek: this is fixed.

LEAF_Request_Portal/templates/view_reports.tpl
Vaceek: indention fixed

PR #1382, opened 2021-06-18
DEV: LEAF-2239 / add resoved by

Vaceek:


pete-nerantzinis (approved):
Looks great I'm approving but just wanted to note I would like to see the last name come first comma first name just in-case people will sort the column they can sort it by last name automatically since last name is first. 

My second concern is why there is a test json file being commited as well? This looks a bit out of place and looks like its removing a lot of things... Just a concern I guess.

Other than that it looks great.

pete-nerantzinis (changes requested):
Revised my comments above

LEAF_Request_Portal/form.php
```diff
@@ -3061,6 +3066,24 @@ public function query($inQuery)
...
+            require_once 'VAMC_Directory.php';
+            $dir = new VAMC_Directory;
+
+            $res2 = $this->db->prepared_query('SELECT recordID, userID as resolvedBy FROM action_history
+                                                LEFT JOIN dependencies USING (dependencyID)
+                                                LEFT JOIN actions USING (actionType)
+                                                WHERE recordID IN (' . $recordIDs . ')
+                                                and stepID > 0
+                                                ORDER BY time', array());
+
```
gsteinhoff, line 3079: As Pete stated, this current query only gets the last person to touch a request in the workflow.  That "touch" could be a comment, move between steps that doesn't complete the request, or a move backward in the workflow.  This will need to be redone so it's checking against workflow next Steps vs. last step taken.  If they are both zero, then it's at the end of the workflow and is resolved.

LEAF_Request_Portal/form.php
```diff
@@ -3061,6 +3066,24 @@ public function query($inQuery)
...
+            foreach ($res2 as $item) {
+                $user = $dir->lookupLogin($item['resolvedBy']);
+                $nameResolved = isset($user[0]) ? "{$user[0]['Fname']} {$user[0]['Lname']}" : $item['resolvedBy'];
```
gsteinhoff, line 3082: As per the rest of our code base, we should be listing names as "lastName, firstName" as it does when searching users and displaying them within the User Access Group admin section.

gsteinhoff (changes requested):

LEAF_Request_Portal/form.php
```diff
@@ -3061,6 +3066,24 @@ public function query($inQuery)
             }
         }
 
+        if ($joinRecordResolutionBy) {
```
Pelentan, line 3069: Need this not to be ambiguous.  "If" what?  !empty?  TRUE?

Pelentan (changes requested):

LEAF_Request_Portal/form.php
gsteinhoff: The query can even be modified slightly from the one just above this one as it deals with Resolved data and probably already has the data you need for the report (or close to it).

gsteinhoff (changes requested):

LEAF_Request_Portal/form.php
pete-nerantzinis: Yeah I think all the functions are coded in if statements like this unfortunately 🤣

LEAF_Request_Portal/form.php
Vaceek: fixed

LEAF_Request_Portal/form.php
Vaceek: fixed

LEAF_Request_Portal/form.php
Vaceek: query fixed to pull correct info

pete-nerantzinis (approved):
Looks good

LEAF_Request_Portal/form.php
```diff
@@ -3061,6 +3066,23 @@ public function query($inQuery)
...
+
+            $res2 = $this->db->prepared_query('SELECT recordID, userID as resolvedBy, stepID, time FROM action_history
+                                                LEFT JOIN workflow_routes USING (stepID)
+                                                WHERE recordID IN (' . $recordIDs . ')
+                                                and nextStepID = 0
+                                                ORDER BY time', array());
+
+            foreach ($res2 as $item) {
+                $user = $dir->lookupLogin($item['resolvedBy']);
+                $nameResolved = isset($user[0]) ? "{$user[0]['Lname']}, {$user[0]['Fname']}" : $item['resolvedBy'];
```
aerinkayne, line 3081: 3081 in form.php ternary looks like it's assigning not checking value equality

aerinkayne (changes requested):

LEAF_Request_Portal/form.php
Vaceek: ? it is assigning the first and last name if it can find it. What were you expecting that isn't there?

Vaceek: the inline evaluates if/then, not a = b, and is correct usage.

Pelentan (approved):

aerinkayne (approved):

LEAF_Request_Portal/form.php
gsteinhoff: Issue is still present.  the query on line 3052 from this file has the exact query you need and you only need to add the action_history table to get the userID.  you might even be able to remove some of the joins depending on the data you need.  I would review that query and see how it can be modified.  it should definitely at least have the non-submitted and deleted requests removed from it's search.

gsteinhoff (changes requested):

gsteinhoff (changes requested):
With query change, it looks good to be tested.

gsteinhoff (approved):

PR #1383, opened 2021-06-18
DEV: LEAF-1585 - OR Operator for Reports and Search

pete-nerantzinis:


Vaceek (approved):

gsteinhoff (changes requested):
Just a small change in logic but otherwise looks great.

LEAF_Request_Portal/form.php
```diff
@@ -2487,6 +2488,23 @@ public function query($inQuery)
...
+            // Logic for AND/OR Operator's
+            $op = 'AND';
+            if ($q['op']) {
+                $op = $q['op'];
+            }
+            if ($count === 0) {
+                $op = '';
+                $conditions = '(';
+            }
+            if ($count != 0) {
```
gsteinhoff, line 2500: The logic here can be consolidated into count is zero or else.  don't need to check it again if it's just every other number but zero.

gsteinhoff (changes requested):

LEAF_Request_Portal/form.php
pete-nerantzinis: Consolidated 👍

gsteinhoff (approved):
Looking good, nice work!

aerinkayne (approved):

Pelentan (approved):

PR #1387, opened 2021-06-23
DEV: LEAF-1585 - OR Operator for Reports and Search

pete-nerantzinis:
IE Fix
Query Fix


Vaceek (approved):

gsteinhoff (approved):
Changes look good.  Nice work to fix IE.

Pelentan (approved):
Going to need to look into getting Michael Gao to pull the plug on IE.  Don't let me forget that.

PR #1404, opened 2021-07-02
LEAF-2138 - Last Workflow Change in Report Builder

Pelentan:
Obviously this is a draft only.  Will allow us to look at the changes re master.

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -2851,6 +2851,7 @@ class Form
         $joinAllCategoryID = false;
         $joinRecords_Dependencies = false;
         $joinRecords_Step_Fulfillment = false;
+        $addJoinRecords_Step_Fulfillment_Only = false;
         $joinActionHistory = false;
         $joinRecordResolutionData = false;
         $joinInitiatorNames = false;
@@ -2886,8 +2887,12 @@ class Form
                         break;
                     case 'stepFulfillment':
                         $joinRecords_Step_Fulfillment = true;
-
                         break;
+
+                    case 'stepFulfillmentOnly':
+                        $addJoinRecords_Step_Fulfillment_Only = true;
+                        break;
+
                     case 'recordResolutionData':
                         $joinRecordResolutionData = true;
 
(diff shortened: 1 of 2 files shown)


LEAF_Request_Portal/form.php
```diff
@@ -3063,15 +3068,31 @@ public function query($inQuery)
 
         if ($joinRecords_Step_Fulfillment)
         {
-            $res2 = $this->db->prepared_query('SELECT * FROM records_step_fulfillment
-    											LEFT JOIN workflow_steps USING (stepID)
-    											WHERE recordID IN (' . $recordIDs . ')', array());
+            $strSQL = 'SELECT * FROM records_step_fulfillment LEFT JOIN workflow_steps USING (stepID) '.
```
Pelentan, line 3071: Please change this so it's not "select *".

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -187,22 +187,51 @@ function addHeader(column) {
...
+            let headerName = (column == 'days_since_last_action') ? 'Days Since Last Action' : 'Days Since Last Workflow Change';
+            let indicatorIDName = (column == 'days_since_last_action') ? 'daysSinceLastAction' : 'daysSinceLastStepMovement';
+            headers.push({
+                name: headerName,
+                indicatorID: indicatorIDName,
+                editable: false,
+                callback: function(data, blob) {
+                    let daysSinceAction;
+                    let recordBlob = blob[data.recordID];
+                    if(recordBlob.action_history != undefined) {
```
Pelentan, line 203: I know this is likely legacy, but please look at pulling it out into its own function.  Since we're not going to overload the QA team this time around, we have some time for cleaning up old code.  Just make sure to log the time into the "cleanup" Jira and not the one for this issue.

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -286,7 +315,9 @@ function loadSearchPrereqs() {
             buffer += '<div class="indicatorOption"><input type="checkbox" class="icheck" id="indicators_approval_history" name="indicators[approval_history]" value="approval_history" />';
             buffer += '<label class="checkable" style="width: 100px" for="indicators_approval_history"> Approval History</label></div>';
             buffer += '<div class="indicatorOption"><input type="checkbox" class="icheck" id="indicators_days_since_last_action" name="indicators[days_since_last_action]" value="days_since_last_action" />';
-            buffer += '<label class="checkable" style="width: 100px" for="indicators_days_since_last_action"> Days since last action</label></div>';
+            buffer += '<label class="checkable" style="width: 100px" for="indicators_days_since_last_action"> Days Since Last Action</label></div>';
```
Pelentan, line 318: Nothing to change here.  Just something to put into everyone's brain.  Movement onto the new UI is going to mean rows like this are going to be going away.  Variables are going to be supplied to a calling api entrance point.  So we may want to start breaking those variables out of lines like this.  The ones defined by the switch operation in this case, and then calling a different function that creates the html.  Then the return on the outer function will return a json that can have both/either just the variables and/or the html block.

LEAF_Request_Portal/form.php
Vaceek: is this an update query??

pete-nerantzinis:
Looks great aside from what Michael mentioned although I believe this SELECT needs to be * because all columns are being used I believe.

PR #1409, opened 2021-07-09
Add files via upload

maxrichardjr:
Test to determine which local 'repositories' are not needed

### DIFF
--- a/test/TestFile.txt
+++ b/test/TestFile.txt
@@ -1,0 +1,1 @@
+This is file to test uploading a .txt. file.


AmandaSumner (approved):

aerinkayne (approved):

maxrichardjr:
Had it committing to Master apparently

maxrichardjr:
I’d already cancelled the request. I’ll try again tomorrow. I really need it to go to master, and because it is only a textfile (which I will remove afterwards), I don’t see the harm. Trying to determine which of my 4 or 5 repositories are actually needed, and which one contains MY code. I know I can just go to the site and select -> open files in explorer, but for some reason I don’t completely trust that because of the way Greg set it up.

maxrichardjr:
Max Richard would like to recall the message, "[VHAINNOVATIONS/LEAF] Add files via upload (#1409)".

PR #1415, opened 2021-07-13
QA: LEAF-2194 - Add Delete to Disabled Form Indicators

Vaceek:


LEAF_Request_Portal/admin/templates/mod_form.tpl
```diff
@@ -1095,9 +1095,16 @@ function getForm(indicatorID, series) {
                     </tr>\
                     <tr>\
                         <td>Disabled</td>\
-                        <td colspan="1"><input id="disabled" name="disabled" type="checkbox" /></td>\
+                        <td colspan="1"><input id="disabled" name="disable_or_delete" type="radio" /></td>\
```
pete-nerantzinis, line 1098: My only concern for these buttons user wise is if they accidently click it they can't unclick either or because its a radio, so one always has to be selected. I would like to see it as a grouped checkbox system, only one of them can be checked at a time or make it disable the other checkbox if one is selected so users can unselect if they change their mind and made changes.

pete-nerantzinis (changes requested):

LEAF_Request_Portal/admin/templates/mod_form.tpl
raphaelfretter: This might help: http://wastedpotential.com/html-multi-checkbox-set-the-correct-way-to-group-checkboxes/

raphaelfretter (changes requested):
This might help: http://wastedpotential.com/html-multi-checkbox-set-the-correct-way-to-group-checkboxes/

aerinkayne (commented):
Updating so that users can uncheck the selections if needed.  We prefer to keep the way the info is handled consistent with the other checkboxes for now.

aerinkayne:
Updates have been made for 2194, but it has been pushed to a new branch https://github.com/VHAINNOVATIONS/LEAF/tree/feature/LEAF-2194%2Fform_archival

PR #1422, opened 2021-07-20
QA: LEAF-2335 - create requests from report builder

aerinkayne:


AmandaSumner (approved):

pete-nerantzinis (approved):
Looks good, functions properly.

gsteinhoff (approved):

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -688,6 +692,58 @@ function showJSONendpoint() {
...
+    dialog.setTitle('New Request Title');
+    dialog.setContent('<div><h4>Title of Request</h4>'
+                    + '<p>Enter a title and save changes to create new requests.<br/>They will appear at the top of the table.</p>'
+                    + '<span class="text">'
+                    + '<input id="newTitle" type="text" name="newTitle" required placeholder="Request Title" style="font-size: 1.3em; font-family: monospace"/>'
+                    + '</span>'
+                    + '</div>');
+    dialog.show();
+    let titleInput;
+    $('#button_save').on('click', function() {
```
Pelentan, line 713: Couple things on this.  

titleInput needs verified.  The "if" will test true on " " and "0".  Neither of which I'm guessing are useful.

catID is unchecked before it gets used.  I know IE doesn't allow default values in function attributes, much less type-forcing.  This can be gotten around by modifying the "if" statement that uses it to something like this:  
if (typeof(Number(jj)) === "number" && Number(jj) > 0) {}

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -3,7 +3,7 @@
 .group:after,.section{clear:both}.section{padding:0;margin:0}.col{display:block;float:left;margin:1% 0 1% 1.6%}.col:first-child{margin-left:0}.group:after,.group:before{content:"";display:table}.group{zoom:1}.span_3_of_3{width:100%}.span_2_of_3{width:66.13%}.span_1_of_3{width:32.26%}@media only screen and (max-width:480px){.col{margin:1% 0}.span_1_of_3,.span_2_of_3,.span_3_of_3{width:100%}}
 </style>
 
-<div id="step_1" style="<!--{if $query != '' && $indicators != ''}-->display: none; <!--{/if}-->width: 70%; background-color: white; border: 1px solid black; margin: auto; padding: 0px">
+<div id="step_1" style="<!--{if $query != '' && $indicators != ''}-->display: none; <!--{/if}-->width: 600px; background-color: white; border: 1px solid black; margin: 2em auto; padding: 0px">
```
Pelentan, line 6: Why was this changes from % to fixed?  This will cause issues when using smaller screens like phones or 'pads.

Pelentan (changes requested):

LEAF_Request_Portal/templates/view_reports.tpl
pete-nerantzinis: This change actually make it better visually on phones and pads. The previous styling was horrendous haha :)

LEAF_Request_Portal/templates/view_reports.tpl
pete-nerantzinis: Its ok if it tests true on numbers people tend to put only numbers, I don't know if theres only a space maybe some users do that? they like to have blank titles so they just put a space. I can do a check for that. And doing a check for catID I completely agree.

Pelentan (approved):
Not perfect, but a workable iteration.

PR #1437, opened 2021-08-03
FEATURE: LEAF-2405 - Dedicated Session Timeout Page

gsteinhoff:


LEAF_Request_Portal/session_expire.php
```diff
@@ -0,0 +1,22 @@
+<?php
+// Include XSSHelpers
+if (!class_exists('XSSHelpers'))
+{
+    include_once __DIR__ . '/../libs/php-commons/XSSHelpers.php';
+}
+
+include '../libs/smarty/Smarty.class.php';
+
+function customTemplate($tpl)
```
Pelentan, line 10: Need default value for $tpl attribute.  Additionally type hinting.  https://www.php.net/manual/en/language.types.declarations.php.  Would like to start seeing this both for the arguments and return type.  Exp:

function customTemplate(string $tpl='default_tpl_name')  ?string { ... }

LEAF_Request_Portal/templates/view_session_expire.tpl
```diff
@@ -0,0 +1,33 @@
...
+        </div>
+        <span id="headerTab"></span>
+        <span id="headerTabImg"><img src="images/tab.png" alt="tab" /></span>
+    </div>
+    <div id="body">
+        <div id="content">
+            <div id="bodyarea">
+                <div class="card" style="width: 50%; margin: auto; padding: 16px">
+                    <h2>Your Session Has Expired</h2>
+                    <p>Return to your <a href="{$previousPage}">LEAF site</a></p>
```
Pelentan, line 25: Should this route to the "/login" page?  I don't think so, but probably ought to be asked.

Pelentan (changes requested):

LEAF_Request_Portal/session_expire.php
gsteinhoff: This has been updated in the latest commit.

LEAF_Request_Portal/templates/view_session_expire.tpl
gsteinhoff: This has been changed in the latest commit as not really need for the previous page if next page should be login anyway, not back to the old page.

pete-nerantzinis (approved):
Looks good

raphaelfretter (approved):
looks good

maxrichardjr (approved):
Thanks for the IDs!

AmandaSumner (approved):

PR #1439, opened 2021-08-06
hotfix: LEAF-2416 fix export button, verbiage

aerinkayne:


Pelentan (approved):
Looks good.

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -950,8 +950,6 @@ $(function() {
 
     	// create save link once
     	if(!extendedToolbar) {
-    	    //adding the below line to overwrite the grid's export button instead of changing it in the grid file, since I'm not sure it's not used for anything else
```
raphaelfretter, line 953: Was the export button not used for anything else?

raphaelfretter (approved):

LEAF_Request_Portal/templates/view_reports.tpl
pete-nerantzinis: Export button is in formGrid.js, we added it here for the reasons listed in the comment but it caused errors since the functionality of the button was in formGrid.js, so we just changed it there and it works everywhere now

AmandaSumner (approved):

Pelentan:
Looks good.

PR #1442, opened 2021-08-10
Bugfix: LEAF-1987 - Currency/Validation/Required checks and Auto-Scroll to Input fields

pete-nerantzinis:


maxrichardjr (approved):
Looks good.

AmandaSumner (approved):

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -116,7 +116,13 @@
                         return !valid;
                     },
                     setRequiredError: function() {
+                        $([document.documentElement, document.body]).animate({
```
gsteinhoff, line 119: With all of these animations happening on the page at once, it should obey on the first animate, stop, and clear the queue of all future animations until the "Next Question" button is clicked.  If you don't, a page with multiple animates will scroll, pause, scroll, pause all the way down the page to the last required error.  

Look into the use of stop(), finish() or clearQueue() jquery functions to find out the combination you need.

gsteinhoff (changes requested):

LEAF_Request_Portal/js/parallelProcessing.js
```diff
@@ -46,17 +46,19 @@ function parallelProcessing(recordID, orgChartPath, CSRFToken)
...
+                            .attr('value', indicatorObject[i].indicatorID)
+                            .html(indicatorObject[i].name)
+                            .appendTo($("select#indicator_selector"));
+                    }
                 }
-                if(obj.length == 0) {
+                else {
                     $('#submitControl').css('display', 'none');
-                    $('#selectDiv').html('<span style="font-size: 120%">Error: The form must contain a field of type "orgchart group" or "orgchart employee" to begin Parallel Processing.</span>');
+                    $('#selectDiv').html('<span style="font-size: 120%">Error: The form/workflow must contain a field of type "orgchart group" or "orgchart employee" to begin Parallel Processing.</span>');
```
Pelentan, line 61: Let's start pulling text out of the line that uses them.  Eventually they are going to be supplied from a completely different resource based on using a response template.  So in this case, inside the "else" clause create a "failResponse" variable and assign the response text to it.  And then an "htmlResponse" variable that has the rest of what is currently in the .html() and uses the failResponse variable.  Then with the $('#selectDiv').html(htmlResponse).  I know that seems like a lot of work for something that is simple, but it's going to start to pay off big time when we go to move things out.

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -116,7 +116,13 @@
...
+                            scrollTop: $('#<!--{$indicator.indicatorID|strip_tags}-->_required').offset().top
+                        }, 700);
                         $('#<!--{$indicator.indicatorID|strip_tags}-->_required').css({"background-color": "red", "color": "white", "padding": "4px", "font-weight": "bold"});
+                    },
+                    setRequiredOk: function() {
+                        $('#<!--{$indicator.indicatorID|strip_tags}-->_required').css({"background-color": "white", "color": "red", "padding": "0", "font-weight": "normal"});
```
Pelentan, line 125: The same for the .css values, please.  Separate json variable.

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -151,7 +157,13 @@
                     return ($('#<!--{$indicator.indicatorID|strip_tags}-->').val().trim() == '');
                 },
                 setRequiredError: function() {
+                    $([document.documentElement, document.body]).animate({
+                        scrollTop: $('#<!--{$indicator.indicatorID|strip_tags}-->_required').offset().top
+                    }, 700);
```
Pelentan, line 162: Unless I'm wrong, these are all using the same css values.  Definitely need a properly scoped variable to handle them all.

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -375,25 +426,46 @@
...
             <!--{if $indicator.required == 1}-->
             formRequired["id<!--{$indicator.indicatorID}-->"] = {
                 setRequired: function() {
                     return ($('#<!--{$indicator.indicatorID|strip_tags}-->').val() == '');
                 },
                 setRequiredError: function() {
+                    $([document.documentElement, document.body]).animate({
+                        scrollTop: $('#<!--{$indicator.indicatorID|strip_tags}-->_required').offset().top
+                    }, 700);
                 	$('#<!--{$indicator.indicatorID|strip_tags}-->_required').css({"background-color": "red", "color": "white", "padding": "4px", "font-weight": "bold"});
```
Pelentan, line 465: Hey look!  Same css  values!  ;-)

Pelentan (changes requested):

LEAF_Request_Portal/templates/subindicators.tpl
pete-nerantzinis: 👍 Changed

LEAF_Request_Portal/js/dialogController.js
```diff
@@ -94,17 +94,17 @@ dialogController.prototype.showButtons = function() {
 };
 
 dialogController.prototype.enableLiveValidation = function() {
-	var t = this;
-	$('input[type="text"]').on('keyup', function() {
+	let t = this;
+	$('body').on('change', function() {
```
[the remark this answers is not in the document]
pete-nerantzinis: 👍 Changed to .mainform

LEAF_Request_Portal/js/parallelProcessing.js
pete-nerantzinis: 👍 Changed

LEAF_Request_Portal/templates/subindicators.tpl
pete-nerantzinis: 👍 Changed the css to an actual class on the style.css and implemented the style to toggle.

LEAF_Request_Portal/templates/subindicators.tpl
pete-nerantzinis: 👍 Changed per last conversation

LEAF_Request_Portal/templates/subindicators.tpl
pete-nerantzinis: 👍 Changed per last conversation

LEAF_Request_Portal/js/dialogController.js
```diff
@@ -94,24 +96,32 @@ dialogController.prototype.showButtons = function() {
 };
 
 dialogController.prototype.enableLiveValidation = function() {
-	var t = this;
-	$('input[type="text"]').on('keyup', function() {
+	let t = this;
```
Pelentan, line 99: Sorry Pete. Missed this the first time.  I know this wasn't you.  Please get rid of all of them (there are at least 3) and fix all the times it was used.  I'm surprised this hasn't bitten up in the behind yet.

LEAF_Request_Portal/js/dialogController.js
```diff
@@ -125,17 +135,25 @@ dialogController.prototype.isValid = function() {
 	return 1;
 };
 
-dialogController.prototype.isComplete = function() {
+dialogController.prototype.isComplete = function(isSubmit) {
+	isSubmit = isSubmit || false;
 	this.incomplete = 0;
 	for(var item in this.requirements) {
```
Pelentan, line 141: While you're in here, if you please.  ;-)

LEAF_Request_Portal/js/dialogController.js
```diff
@@ -153,7 +171,7 @@ dialogController.prototype.setSaveHandler = function(funct) {
 	$('#' + this.btnSaveID).off();
 	var t = this;
```
Pelentan, line 172: My mind comes up with uncivilized thoughts seeing this.

LEAF_Request_Portal/js/parallelProcessing.js
```diff
@@ -46,17 +46,20 @@ function parallelProcessing(recordID, orgChartPath, CSRFToken)
...
             url: 'api/?a=form/'+recordID+'/workflow/indicator/assigned',
-            success: function(obj) {
-                indicatorObject = obj;
-                for (var i = 0; i < indicatorObject.length; i++) {
-                    $(document.createElement('option'))
-                        .attr('value', indicatorObject[i].indicatorID)
-                        .html(indicatorObject[i].name)
-                        .appendTo($("select#indicator_selector"));
+            success: function(arr) {
+                indicatorObject = arr;
```
Pelentan, line 50: Why is this line here?  "arr" is not used otherwise.  Why not just make the argument that's passed "indicatorObject"?

LEAF_Request_Portal/js/parallelProcessing.js
```diff
@@ -46,17 +46,20 @@ function parallelProcessing(recordID, orgChartPath, CSRFToken)
...
+                    for (let i = 0; i < indicatorObject.length; i++) {
+                        $(document.createElement('option'))
+                            .attr('value', indicatorObject[i].indicatorID)
+                            .html(indicatorObject[i].name)
+                            .appendTo($("select#indicator_selector"));
+                    }
                 }
-                if(obj.length == 0) {
+                else {
+                    let failResponse = '<span style="font-size: 120%">Error: The form/workflow must contain a field of type "orgchart group" or "orgchart employee" to begin Parallel Processing.</span>';
```
Pelentan, line 60: Since you're in here...  This is almost what I'm looking for.

let failResponse = 'Error: The form/workflow must contain a field of type "orgchart group" or "orgchart employee" to begin Parallel Processing.';

let responseHTML = '<span style="font-size: 120%">' + failResponse + '</span>';

The idea is to keep types clean.  "failResponse" is data.  And should be able to be changed without changing the "code".  "responseHTML" is presentation-code.  It will be handled with Vue in a future iteration.  And the Vue function will take the "failResponse" as an argument and return formatted html.

Pelentan: Since this is outside of the ticket specifically, I'll bring the "Code Grooming ticket in to this sprint and you can put your hours there.  Please make sure you hit this at every point in the files you touch for the ticket.

Pelentan (changes requested):

LEAF_Request_Portal/js/parallelProcessing.js
pete-nerantzinis: That was Carrie and myself first thought haha but we went along with it... Definitely will change

aerinkayne: I had found an error log and at first I thought it was because of the loop, due to the param name (the type of an array is still object, but getting and utilizing the length property is different), so I changed the param name and where it had been used to be more specific (and overlooked that it was otherwise weird to begin with).  Will update to be more consistent.

Pelentan (approved):
Good to go for now. ;-)

LEAF_Request_Portal/js/dialogController.js
```diff
@@ -151,9 +169,9 @@ dialogController.prototype.isComplete = function() {
 
 dialogController.prototype.setSaveHandler = function(funct) {
 	$('#' + this.btnSaveID).off();
-	var t = this;
+	let t = this;
     this.dialogControllerXhrEvent = $('#' + this.btnSaveID).on('click', function() {
-        if(t.isValid() == 1 && t.isComplete() == 1) {        	
+        if(t.isValid(true) == 1 && t.isComplete(true) == 1) {
```
gsteinhoff, line 174: this syntax is weird.  from first glance it looks like this is setting isValid to true.  it's ok, just maybe in the future define a variable like "submitted = true/false" and then send to the isValid and isComplete functions.

gsteinhoff (approved):

LEAF_Request_Portal/js/dialogController.js
pete-nerantzinis: Yeah that would make more sense haha :D

PR #1446, opened 2021-08-10
FEATURE: LEAF-2374 - Site Caching

gsteinhoff:
Uses environment variable to check to invalidate the cache if modified dates don't match. Works for only GET and HEAD page requests.

### DIFF
--- a/LEAF_Request_Portal/db_mysql.php
+++ b/LEAF_Request_Portal/db_mysql.php
@@ -58,6 +58,7 @@ class DB
             }
             $this->isConnected = false;
         }
+        $this->checkLastModified();
         unset($pass);
     }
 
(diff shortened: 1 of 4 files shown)


LEAF_Request_Portal/db_mysql.php
```diff
@@ -290,4 +291,23 @@ public function enableDryRun()
     {
         $this->dryRun = true;
     }
+
+    private function checkLastModified() {
+        //get the last build time
+        $lastBuildTime =  getenv('LAST_BUILD_DATE', true) ?: getenv('LAST_BUILD_DATE');
```
Pelentan, line 297: Sorry, Greg, I messed you up.  You probably haven't pulled the prototype in a while.  These need to be two different values.  I originally had them the same because I wasn't sure where I wanted to put the default variable.  ;-)  The line construct it to make sure the variable is filled with something usable that won't break things further in the code.

gsteinhoff: Changed code to add default Unix epoch date if LAST_BUILD_DATE not found

Pelentan (approved):
Looks good

pete-nerantzinis (approved):

PR #1448, opened 2021-08-11
propagate change events

Pelentan:
This is from Michael Gao

### DIFF
--- a/LEAF_Request_Portal/templates/subindicators.tpl
+++ b/LEAF_Request_Portal/templates/subindicators.tpl
@@ -822,6 +822,7 @@
                                 data: {CSRFToken: '<!--{$CSRFToken}-->'},
                                 success: function(res) {
                                     $('#<!--{$indicator.indicatorID|strip_tags}-->').val(res);
+                                    $('#<!--{$indicator.indicatorID|strip_tags}-->').trigger('change');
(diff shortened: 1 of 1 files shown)


AmandaSumner (approved):

maxrichardjr (approved):
Looks good.

gsteinhoff (approved):

pete-nerantzinis (changes requested):
I dont see the reason for this as there are no indicator ids that hold a change event besides multiselect. This woulnd't trigger on anything I think, I even went through the entire tpls I see nothing that has a change event.

aerinkayne (approved):

mgaoVA:
> I dont see the reason for this as there are no indicator ids that hold a change event besides multiselect. This woulnd't trigger on anything I think, I even went through the entire tpls I see nothing that has a change event.
> 
> I guess its more so me wanting to know what exactly is this triggering. @mgaoVA

@pete-nerantzinis Sorry missed your message here. This resolves a consistency issue for custom UI devs that rely on expected behavior of input fields. Standard input fields such as text/textarea/number automatically trigger changes on their respective field's ID #.

Since the underlying DOM object is not directly controlled by users for the orgchart_employee widget, change events would never fire as expected.

PR #1449, opened 2021-08-13
Feature: LEAF 883 table header color options

aerinkayne:


AmandaSumner (approved):

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -586,20 +558,20 @@ var LeafFormGrid = function(containerID, options) {
...
-         function announceResults(){
-             var term = $('[name="searchtxt"]').val();
-
-             if(currentData.length == 0) {
-                 $('.status').text('No results found for term ' + term);
-         }else{
-                 $('.status').text('Search results found for term ' + term + ' listed below');
-            }
+     function announceResults(){
+         var term = $('[name="searchtxt"]').val();
```
Pelentan, line 565: var or let?

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -536,6 +580,17 @@ function editLabels() {
...
+        gridColorData = Object.assign({}, tempColorData);
+        if(gridColorData != {}) {
+            updateHeaderColors();
+            urlColorData = LZString.compressToBase64(JSON.stringify(gridColorData));
+            url = baseURL + '&v=' + version + '&query=' + encodeURIComponent(urlQuery) + '&indicators=' + encodeURIComponent(urlIndicators) + '&colors=' + encodeURIComponent(urlColorData);
+            if ($('#reportTitle').val() != '') {
+                url += '&title=' + encodeURIComponent(btoa($('#reportTitle').val()));
+            }
+            window.history.pushState('', '', url);
+        }
```
Pelentan, line 592: Does this need an "else"?  What happens if gridColorData _is_ {}.  Even if nothing, would things look better/smoother if we had the option to define values outside of the code?

Pelentan (changes requested):

maxrichardjr (approved):
Looks good to me

LEAF_Request_Portal/js/formGrid.js
aerinkayne: I must have touched the spacing and now it's mine.  I can get this one though.

LEAF_Request_Portal/templates/view_reports.tpl
aerinkayne: Title and colors are situations where there might not be any info to add to the url (so if they are empty nothing).  There already is a keyup listener on the report title input field that changes the url so I needed to rebuild it here too (the Edit Labels save btn).  Open to ways to make it look cleaner though.

aerinkayne:
replacing with new revised branch

PR #1454, opened 2021-08-19
Feature LEAF 883-header color options (new)

aerinkayne:


AmandaSumner (approved):

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -586,20 +558,20 @@ var LeafFormGrid = function(containerID, options) {
...
-                 $('.status').text('No results found for term ' + term);
-         }else{
-                 $('.status').text('Search results found for term ' + term + ' listed below');
-            }
+     function announceResults(){
+         let term = $('[name="searchtxt"]').val();
 
+         if(currentData.length == 0) {
+             $('.status').text('No results found for term ' + term);
+        }else{
```
raphaelfretter, line 569:
```suggestion
         } else {
```

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -248,19 +300,22 @@ function addHeader(column) {
             });
             break;
 	    default:
-	    	if(column.substr(0, 6) == 'depID_') { // backwards compatibility for LEAF workflow requirement based approval dates
-	    		depID = column.substr(6);
```
raphaelfretter, line 252: Was this not working before?

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -296,10 +354,10 @@ function addHeader(column) {
 	}
 }
 
-var resIndicatorList = {};
+var resIndicatorList = { };
```
raphaelfretter, line 357:
```suggestion
var resIndicatorList = {};
```

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -979,41 +1103,46 @@ $(function() {
     	else {
     		$('#editLabels').css('display', 'inline');
     	}
-
     	urlQuery = LZString.compressToBase64(JSON.stringify(leafSearch.getLeafFormQuery().getQuery()));
     	urlIndicators = LZString.compressToBase64(JSON.stringify(selectedIndicators));
 
+
```
raphaelfretter, line 1109:
```suggestion
```

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -979,41 +1103,46 @@ $(function() {
...
             $('#reportTitle').on('keyup', function() {
-                url = baseURL + '&v='+ version + '&query=' + encodeURIComponent(urlQuery) + '&indicators=' + encodeURIComponent(urlIndicators) + '&title=' + encodeURIComponent(btoa($('#reportTitle').val()));
+                url = baseURL + '&v='+ version + '&query=' + encodeURIComponent(urlQuery) + '&indicators=' + encodeURIComponent(urlIndicators);
+                if (urlColorData !== 'str'){
+                    url += '&colors=' + urlColorData;
+                }
+                url += '&title=' + encodeURIComponent(btoa($('#reportTitle').val()));
                 window.history.pushState('', '', url);
             });
+
```
raphaelfretter, line 1128:
```suggestion
```

raphaelfretter (changes requested):
Some minor formatting issues; fix for consistency

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -248,19 +300,22 @@ function addHeader(column) {
             });
             break;
 	    default:
-	    	if(column.substr(0, 6) == 'depID_') { // backwards compatibility for LEAF workflow requirement based approval dates
-	    		depID = column.substr(6);
+	    	if(column.substr(0, 6) === 'depID_') { // backwards compatibility for LEAF workflow requirement based approval dates
+	    		let depID = column.substr(6);
```
pete-nerantzinis, line 304: depID should be passed through at line 326, so I dont see the reason for this let? This might cause issues changing how its previously working. Also the next if statement for stepID is set the same way before as well. On line 329.
```suggestion
	    		depID = column.substr(6);
```

pete-nerantzinis (approved):
It looks very good just a minor thing I was unsure about with the let depID

LEAF_Request_Portal/templates/view_reports.tpl
Pelentan: Not sure which part you have a question about.  If it's the addition of "let" in front of the variable, it's to properly scope it.  Looks like it might have been depending on it being instantiated else where globally.  Always and iffy prospect.

Pelentan (approved):
In some of the format conversions, you missed some changing of some "var"s to "let"s. but considering the number of them you would then have had to test out completely, I'm good with that for now.  Let it ride.

LEAF_Request_Portal/templates/view_reports.tpl
aerinkayne: That one does not appear to be used outside of that block.  However, I've also discussed with Pete and Greg and we all agree that it's better not to touch anything that could affect a workflow unless we are 100% sure on it, so I was going to revert this change unless that's objected to.

LEAF_Request_Portal/templates/view_reports.tpl
aerinkayne: Going to change this just is case per our discussion.  stepID I hadn't touched, because there are some variables defined in the file without first being declared to suggest that they are used globally, and I wasn't confident that this value wasn't also being used elsewhere.

gsteinhoff (approved):

PR #1456, opened 2021-08-22
LEAF-883 most comparison changes back to == or !=

aerinkayne:
Discovered that some comparisons had been using conversion (eg '0' == 0).  Changed most comparisons back to double == to avoid potential problems.


pete-nerantzinis (approved):
Dont think this would cause many issues but understandable.

raphaelfretter (approved):
looks good

gsteinhoff (approved):

PR #1462, opened 2021-09-07
Feature: LEAF-2441 - expand report builder for grids

raphaelfretter:


AmandaSumner (approved):

maxrichardjr (approved):
Looks good.

Pelentan (approved):
Looks good.

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -45,21 +45,30 @@ var LeafFormGrid = function(containerID, options) {
...
+            return columnValues.includes(value.id);
+        });
+
+
         //finds and displays column names
         for(var i = 0; i < columns; i++){
-            tDelim = (i == columns-1) ? '' : delim;
-            gridHeadBuffer +='<td style="width: 100px;">' + values.format[i].name + tDelim + '</td>';
-            columnOrder.push(values.format[i].id)
+            if (columnValues.includes(values.format[i].id)) {
```
pete-nerantzinis, line 67: Getting error here during testing, 
image

pete-nerantzinis (changes requested):
Found some errors during testing that caused the report builder to crash. Line 67 of formGrid.js and a few others I could not determine.

LEAF_Request_Portal/js/formGrid.js
aerinkayne: I got the same.  I sent him an image of the checkbox conditions that seemed to be associated with it.

pete-nerantzinis: This looks to be working now

pete-nerantzinis (changes requested):
I was just about to approve but then came across this issue.

image

Seems when you create a report then refresh the page to get the cached version of the report from sharing the url, it does not save the grid selections. It will refresh properly at first but if you go to modify and regenerate it will give the error above, and closer looking on modify on step 2 you can see only the main checkbox is checked and not the sub checkboxes.

To Rrecreate:
1. Go to Report builder and setup report for something that shows grids checkbox the main grid so it selects all grid items
2. Generate Report
3. Refresh the page so its on the cached version where you cannot edit the main title
4. Modify again and go to step 2, notice that the checkboxes for the sub grid items are not checked anymore, then when generate report it bugs out

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -462,6 +541,13 @@ function loadSearchPrereqs() {
                             if(t_inIndicators != undefined) {
                                 for(var i in t_inIndicators) {
                                     $('#indicators_' + t_inIndicators[i].indicatorID).prop('checked', true);
+
+                                    if (t_inIndicators[i].cols !== undefined) {
+                                        for (var j in t_inIndicators[i].cols) {
+                                            console.log('#indicators_' + t_inIndicators[i].indicatorID + '_columns_' + t_inIndicators[i].cols[j]);
```
pete-nerantzinis, line 547: So it mostly looks good now, just clean up this last console log unless its necessary for users. 

Other than that, I was doing some backwards compatibility with reports created before this feature was implemented with a grid field type. It doesn't work it gives me this error:
image

Just make sure its also backwards compatible by moving back to master branch and creating a report with grids and saving the link then checking out the new branch and testing that link for backwards compatibility.

I think this is good to go after that is done.

pete-nerantzinis (changes requested):
Sorry Raphael, I caught one last thing!

beingforthebenefit (approved):

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -45,7 +45,14 @@ var LeafFormGrid = function(containerID, options) {
...
      */
-    function printTableReportBuilder(values) {
+    function printTableReportBuilder(values, columnValues) {
+        // remove unused columns
+        if (columnValues !== null) {
+            values.format = values.format.filter(function (value) {
+                return columnValues.includes(value.id);
+            });
+        }
+
```
aerinkayne, line 55: I get an error here if I try to edit/save the grid section.  Adding && !== undefined will fix it.  After editing, the grid resets to the full grid (I don't know if that's an issue or not.  The dialog boxes are their own little monsters).

aerinkayne (approved):

PR #1463, opened 2021-09-07
Bugfix: LEAF-2473 - New Row logic update

aerinkayne:
Addresses the situation of the New Row button not appearing if no rows are returned from the search.

### DIFF
--- a/LEAF_Request_Portal/templates/view_reports.tpl
+++ b/LEAF_Request_Portal/templates/view_reports.tpl
@@ -893,29 +893,33 @@ let isOneFormType = false;
 /**
  * Purpose: Check if only one type of form could logically be returned and,
  * if so, update global variables isOneFormType (bool) and categoryID (string).
- * @param searchQueryTerms - variable with result of leafSearch.getLeafFormQuery().getQuery().terms
+ * @param searchQueryTerms - variable with result of leafSearch.getLeafFormQuery().getQuery().terms (array)
  */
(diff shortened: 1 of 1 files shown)


AmandaSumner (approved):

maxrichardjr (approved):
Looks good

raphaelfretter (approved):
Looks good

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -888,6 +888,37 @@ var indicatorSort = {}; // object = indicatorID : sortID
...
 let gridColorData = {}; //object updated with id: color
 let tempColorData = {}; //object updated with id: color
+let isOneFormType = false;
+
+/**
+ * Purpose: Check if only one type of form could logically be returned and,
+ * if so, update global variables isOneFormType (bool) and categoryID (string).
+ * @param searchQueryTerms - variable with result of leafSearch.getLeafFormQuery().getQuery().terms
+ */
+function checkIfOneTypeSearchedAndUpdate(searchQueryTerms) {
```
Pelentan, line 898: Need some type checking on searchQueryTerms.  I know IE doesn't like it in the call, but inside the function we need to ensure it's actually and object so later calls don't fail badly.

Pelentan (changes requested):

pete-nerantzinis (approved):
Looks good

Pelentan (approved):

beingforthebenefit (approved):

PR #1476, opened 2021-09-23
bug: LEAF-2506 fix required currency bug

aerinkayne:
Moved an empty string check prior to regex and Math methods.  
Please see test suggestions in the associated Jira ticket.

### DIFF
--- a/LEAF_Request_Portal/templates/subindicators.tpl
+++ b/LEAF_Request_Portal/templates/subindicators.tpl
@@ -475,14 +475,18 @@
             formValidator["id<!--{$indicator.indicatorID}-->"] = {
                 setValidator: function() {
                     let value = $('#<!--{$indicator.indicatorID|strip_tags}-->').val().trim();
+                    if (value === ''){
+                        return true
+                    }
                     value = value.replace(/,/ig, '');
-                    if (!/^(\d+)(\.\d+)?$/.test(value)){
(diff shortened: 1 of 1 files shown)


AmandaSumner (approved):

maxrichardjr (approved):
Looks good.

gerald-todd (approved):

gsteinhoff (approved):

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -475,14 +475,18 @@
             formValidator["id<!--{$indicator.indicatorID}-->"] = {
                 setValidator: function() {
                     let value = $('#<!--{$indicator.indicatorID|strip_tags}-->').val().trim();
+                    if (value === ''){
```
Pelentan, line 478: Please get rid of all the "returns".  LINT screams at that.  The accepted method is to declare a return variable at the top of the function (ex $isNumeric = false;) and then fill/modify it as the code goes.  This keeps it to a single return and also helps maintain the return _type_.  Very important when we start tightening up our function calls.  ;-)

Pelentan (changes requested):

raphaelfretter (approved):

Pelentan (approved):
Much better.  Thank you.

PR #1506, opened 2021-10-18
Bug/leaf 2534: fix employee refresh for users with apostrophe

aerinkayne:
This should resolve employee refresh issue in the nexus for users with an apostrophe in their username.

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -2542,6 +2542,9 @@ class Form
                     return 0;
             }
 
+	    if ($q['id'] === 'userID') {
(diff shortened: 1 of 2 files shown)


AmandaSumner (approved):

LEAF_Request_Portal/form.php
```diff
@@ -2542,6 +2542,9 @@ public function query($inQuery)
                     return 0;
             }
 
+	    if ($q['id'] === 'userID') {
+            	$q['match'] = htmlspecialchars_decode($q['match'], ENT_QUOTES);
```
gsteinhoff, line 2546: This should be standardized with XSSHelpers::xssafe function

gsteinhoff (changes requested):

PR #1511, opened 2021-10-25
LEAF-2557 newRow null fix

aerinkayne:
Resolves an error caused by a potential null value when using the newRow button in the Report Builder.
Also added a message that will display in the scenario where the query used does not return the created request, and a variable to disable the button for the duration of report generation to reduce multiple clicks.


gsteinhoff (approved):

maxrichardjr (approved):
I must've missed this one the first time through. Looks good.

maxrichardjr (approved):
Missed viewing one file on previous review. Looks good.

PR #1512, opened 2021-10-27
Rewrite method to compute user upload directory size using only PHP.

gerald-todd:
Rewrite a method to accurately compute user uploads directory size.

A permissions modification is required to fully enable this feature. The following command must be run:

`chmod u+r vaccination_data_reporting/` to allow the script to read files in that directory.

### DIFF
--- a/LEAF_Request_Portal/sources/Telemetry.php
+++ b/LEAF_Request_Portal/sources/Telemetry.php
@@ -8,8 +8,6 @@
     Date Created: March 3, 2016
 
 */
-// For testing purposes
-error_reporting(E_ALL);
 
 $currDir = __DIR__;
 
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/sources/Telemetry.php
```diff
@@ -142,23 +143,15 @@ public function getRequestsPerMonth()
...
-
-        $command = 'du -sb ' . Config::$uploadDir;
-        $output = shell_exec($command);
-        if ($output) {
-            $sizeOutput = explode("\t", $output);
-            return $sizeOutput[0];
-
-        } else {
-            return '';
+    public function getRequestUploadStorage():String
```
mgaoVA, line 149:
```suggestion
    public function getRequestUploadStorage():string
```

Should this be ":string"?

gsteinhoff (approved):

PR #1515, opened 2021-11-06
Simplify query debugging process

mgaoVA:
This introduces a parameter &debugQuery, which has the same effect as adding &debug to any form/query call, except it also shows information about the actual SQL query.

This simplifies debugging of form/query by returning EXPLAIN output. The EXPLAIN output is restricted to admins as it can contain information about of the size of the database. Non-admins will only be able to see the query and standard &debug output.

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -3014,7 +3014,14 @@ class Form
             $vars = array_map(fn($val):string => is_numeric($val) ? $val : '"'.$val.'"', $vars);
             header('X-LEAF-Query: '. str_replace(array_keys($vars), $vars, $debugQuery));
             
-            return XSSHelpers::scrubObjectOrArray(json_decode(html_entity_decode(html_entity_decode($_GET['q'])), true));
+            if($this->login->checkGroup(1)) {
+                return $res = $this->db->prepared_query('EXPLAIN SELECT * FROM records
+                                                        ' . $joins . '
+                                                        WHERE ' . $conditions . $sort . $limit, $vars);
+            }
+            else {
+                return XSSHelpers::scrubObjectOrArray(json_decode(html_entity_decode(html_entity_decode($_GET['q'])), true));
+            }
         }
         $res = $this->db->prepared_query('SELECT * FROM records
     										' . $joins . '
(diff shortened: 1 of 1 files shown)


AmandaSumner (approved):

Pelentan (dismissed):
The only issue I have with this is if the security team finds this, they will have an aneurysm.  While this may be locked down as who can see it and what table, it's still transmitting fairly detailed info over the web on how our database is set up.  Perhaps instead of returning it, it would be better to put this into a log file.  That way it never leaves the server.

maxrichardjr (approved):

mgaoVA:
> 
> 
> The only issue I have with this is if the security team finds this, they will have an aneurysm. While this may be locked down as who can see it and what table, it's still transmitting fairly detailed info over the web on how our database is set up. Perhaps instead of returning it, it would be better to put this into a log file. That way it never leaves the server.

If an issue is raised, I think this would be defensible since the EXPLAIN output is guarded behind an admin check.

LEAF_Request_Portal/form.php
```diff
@@ -3009,6 +3009,20 @@ public function query($inQuery)
             $joins .= "LEFT JOIN (SELECT userName, lastName, firstName FROM {$this->oc_dbName}.employee) lj_OCinitiatorNames ON records.userID = lj_OCinitiatorNames.userName ";
         }
 
+        if(isset($_GET['debugQuery'])) {
+            $debugQuery = str_replace(["\r", "\n"], ' ', 'SELECT * FROM records ' . $joins . 'WHERE ' . $conditions . $sort . $limit);
+            $debugVars = array_map(fn($val):string => is_numeric($val) ? $val : '"'.$val.'"', $vars);
```
raphaelfretter, line 3014: This ternary operator seems backwards. If you want to return a string, shouldn't it convert to string if it's numeric and leave it alone otherwise?

raphaelfretter (changes requested):
https://github.com/department-of-veterans-affairs/LEAF/pull/1515/files#r745587711

LEAF_Request_Portal/form.php
mgaoVA: The purpose of line 3014 is to provide a more realistic looking query by wrapping strings in quotes, since SQL expects strings to be quoted while numeric values aren't.

While looking at this, additional logic needs to be added to wrap numbers in quotes in relation to the "data" table, since data is being stored as strings.

raphaelfretter (approved):

gerald-todd (approved):

gsteinhoff:
The Fortify scan done had about 1/3 of it's CRITICAL issues caused by code like this which was submitted to OIS.  I had to put in that we'd make sure to universally scrub these SQL inputs throughout our database classes, LEAF-2625, so we didn't get dinged in our ATO for having them (or at least that's what I had to put because it technically is still a security risk even after scrubbing).

PR #1517, opened 2021-12-08
LEAF 2575 disable local users if not found on refresh

aerinkayne:
Adds a query to disable users locally if they are not found in national when they are refreshed in the nexus.
Also associated with remaining issues for ticket 2534:  
Users with apostrophes in their name should now be found in national if present (and not disabled) and correctly updated on refresh.


AmandaSumner (approved):

maxrichardjr (approved):
Looks good

gerald-todd (approved):

gsteinhoff (approved):

Pelentan (approved):
I'm not a fan of the blanket decode of htmlspecialchars.   However, I'm not coming up with a good way to get around it right now.

raphaelfretter (approved):

PR #1520, opened 2021-12-15
Resolve DOM ID collision issue

mgaoVA:
When a workflow has multiple steps, each with a Form Field set, it generates multiple "workflowIndicator" IDs and prevents the UI from populating correctly.

This fix appends the stepID to avoid collisions.


AmandaSumner (approved):

maxrichardjr (approved):
Looks good

gerald-todd (approved):

gsteinhoff (approved):

raphaelfretter (approved):

PR #1528, opened 2022-02-02
LEAF-2698: add condition to distinguish username with space from first last format

aerinkayne:
This was the hotfix to prod


AmandaSumner (approved):

gerald-todd (commented):
Looks good

PR #1534, opened 2022-03-08
Group Backup Sync Fix

gerald-todd:
Changed the behavior of the 'Sync Services' button in the admin panel. It now syncs members of groups that are not marked as 'active'. Previously, it would ignore 'inactive' members of the group. The problem in ticket LEAF-2668 can now be resolved with the 'Sync Services' button.


AmandaSumner (approved):

maxrichardjr (approved):
Nice.

Pelentan (approved):

pete-nerantzinis (approved):
I am approving but there was a reason I put active != 0 and I can't remember now for the life of me since this was a long time ago. But as long as testers really test it out frontwards and backwards and there are no issues then I see no reason not to push this.

PR #1535, opened 2022-03-10
Clarify API

mgaoVA:
This deprecates the /nonadmin endpoint. Using the masquerade modifier clarifies the intent of the request. This is left as a modifer since it communicates the the Login system: "I'm an admin and want to masquerade as a non-admin".

The /nonadmin endpoint should be removed in the future, after implementations to it are updated in prod. Implementations of need to be replaced with inbox/dependency/[text]?masquerade=nonAdmin


AmandaSumner (approved):

maxrichardjr (approved):
Looks good.

gerald-todd (approved):

Pelentan (approved):

PR #1548, opened 2022-04-18
LEAF 2795 alert IE users

aerinkayne:
Displays an alert if a user accesses the site with Internet Explorer that advises them to use Chrome or Edge.
Also displays a more prominent message on the portal homepage.

### DIFF
--- a/LEAF_Nexus/admin/templates/main.tpl
+++ b/LEAF_Nexus/admin/templates/main.tpl
@@ -1,6 +1,11 @@
 <!DOCTYPE html>
 <html>
 <head>
+    <script>
+    if(navigator.userAgent.indexOf("Trident") != -1) {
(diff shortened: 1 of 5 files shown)


AmandaSumner (approved):

gerald-todd (approved):

maxrichardjr (approved):

LEAF_Nexus/admin/templates/main.tpl
```diff
@@ -1,6 +1,11 @@
 <!DOCTYPE html>
 <html>
 <head>
+    <script>
+    if(navigator.userAgent.indexOf("Trident") != -1) {
```
raphaelfretter, line 5: Should we check for older versions of IE as well i.e. checking 'MSIE ' as well as 'Trident'?

Pelentan (approved):

PR #1573, opened 2022-05-26
LEAF 2879 handle empty unsaved 'checkboxes' values

aerinkayne:
If a form with 'checkboxes' format items is immediately viewed as a single page (without entering data / clicking next question), validation to enter the 'no' values does not run.  The respective 'checkboxes' input element's value is an empty string instead of an array, which in turn causes the code associated with the 'print to PDF' feature to error.

Adds an or operator to ensure that the returned value for checkboxes format questions is an array if there are unentered/unsaved data.


AmandaSumner (approved):

maxrichardjr (approved):

raphaelfretter (approved):
Looks good.

PR #1575, opened 2022-05-31
LEAF 2833/2834 - multiselect format: comma issues, data serialization, and checkboxes interchangeability

aerinkayne:
**edited:**  sanitized multiselect string prior to split, minor associated edits to subindicators.tpl and formPrint.js for comparisons.

Multiselect format data are currently stored as comma separated strings directly via the 'chosen' plugin (eg option A,option B,option C) rather than as serialized data.  This causes issues if users create options with commas in them and also introduces some data ambiguity.

These updates are intended to resolve issues caused by commas present in options (eg 'gas, lodging, and other travel expenses' should be interpreted as one option and not three).  The chosen plugin _selected value is also further formatted so that the data are stored as serialized arrays (**requires minor updates to customizable subindicators.tpl and print_subindicators_ajax**).   A separate script on LEAF-Scripts (or an update file) would need to be run to serialize previously entered multiselect data.

other:  
- resolves an issue where multiselect data could not be completely de-selected.
- resolves failure to check selected items in print to PDF for radio and dropdown formats if the option contains encoded characters


gerald-todd (approved):

maxrichardjr (approved):
Also Looks good.

LEAF_Request_Portal/form.php
```diff
@@ -1138,7 +1138,8 @@ public function doModify($recordID)
         {
             // If form has _selected key use over initial key (Multi-Select Dropdown)
             if (is_numeric($key) && $_POST[$key . '_selected']) {
-                $_POST[$key] = $_POST[$key . '_selected'];
+                $strval = $_POST[$key . '_selected'];
```
Pelentan, line 1141: Are the $_POST variables getting sanitized somewhere?

Pelentan (changes requested):

LEAF_Request_Portal/form.php
aerinkayne: I made some updates to sanitize the string that contains the multiselect options.  Also noted the possible issue with checkboxes format (checkboxes format is not related to this ticket)

AmandaSumner (approved):

maxrichardjr (approved):
Nice!

raphaelfretter (approved):

PR #1585, opened 2022-06-14
LEAF-2958: Include full field name in PDF export

mgaoVA:
The original behavior would exclude the full field name if a short label exists. Since the full field name often includes instructions, it should be included in the PDF.


maxrichardjr (approved):
Looks good

raphaelfretter (approved):

aerinkayne (approved):

gerald-todd (approved):

pete-nerantzinis (approved):

PR #1587, opened 2022-06-16
LEAF-2915: KILL CONNECTION_ID() line is causing warnings

raphaelfretter:


LEAF_Request_Portal/db_mysql.php
```diff
@@ -90,7 +90,13 @@ public function __destruct()
             echo '<hr />';
             echo "</pre><br />Time: {$this->time} sec<br />";
         }
-        $this->db->query('KILL CONNECTION_ID()');
+        
+        try {
+            $this->db->query('KILL CONNECTION_ID()');
```
Pelentan, line 95: Need logging for both succeed and fail on the try/catch.  Would like to see if this is working at all.  When we added it last year, it did seem to clear up some things, but that could have been do to other things.

Pelentan (changes requested):

Pelentan (approved):

pete-nerantzinis (approved):

aerinkayne (approved):

gerald-todd (approved):

PR #1594, opened 2022-07-08
LEAF-2958: Include full field name in PDF export

pete-nerantzinis:
The original behavior would exclude the full field name if a short label exists. Since the full field name often includes instructions, it should be included in the PDF.

### DIFF
--- a/LEAF_Request_Portal/js/formPrint.js
+++ b/LEAF_Request_Portal/js/formPrint.js
@@ -120,12 +120,12 @@ var printer = function() {
                 var title = typeof (indicator.description) !== "undefined"
                 && indicator.description !== null
                 && indicator.description.length > 0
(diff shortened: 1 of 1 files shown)


gerald-todd (approved):

raphaelfretter (approved):

AmandaSumner:
While testing really long field names (like a page long), I found that the second page is blurry due to the double print of the words and the bottom of the first page is crowded together at the bottom. Here are the screenshots: 

d39a1b04-e37a-4d6a-a5ee-03b0e4822107

98d48c56-f5c8-4f8c-83cb-1323cfd987fd

aerinkayne (approved):

aerinkayne (commented):
Updating my review selection after seeing Amanda's comment.  Long name issue should be looked at further

Pelentan (approved):

jampaul3 (approved):

PR #1598, opened 2022-07-13
LEAF-2907 Time filters were not working as expected when other filters were cha

shaneodd:
Time filters were not working as expected when other filters were changed as well as when changing the time filters

The date time pickers were not refreshing the data, and if other filters were called the time pickers did nothing.

Things of note look at Resolved Requests and Resolution time per service values here should change when setting and dragging the filters around within Resolution time per month.


AmandaSumner (approved):

maxrichardjr (approved):
Great Comment/Description!!

aerinkayne (approved):

Pelentan (approved):

PR #1599, opened 2022-07-13
LEAF 2920 - Create a process to run a script on an interval between t…

shaneodd:
Create a process to run a script on an interval between two times

This will allow a process to be run on a set interval between two times, this is an example of a class with hopefully all of the fixins that Michael S was looking for.


pete-nerantzinis (approved):

shaneodd:
Please note this will required the php 8 change to go live before this goes live.

pete-nerantzinis (approved):
It looks good, approving for now but wanted to note that the readme file will be pushed out to portals as well and don't know if this is something we want on the front end accessibility wise, same thing with the batch scripts being in frontend areas.

raphaelfretter (approved):

Pelentan (approved):

PR #1600, opened 2022-07-18
Rc/2022 07 18/sprint 35 c1

Pelentan:
This is for these tickets:

LEAF-2833
LEAF-2896
LEAF-2907 
LEAF-2961 
Prevent race conndition
Stop Edge prompting for language translation


shaneodd (commented):
Still cannot get into Demo1 however the RC running locally appears to be good, that being said I do not feel comfortable approving so will just comment.. I primarily looked at 2907 and that appeared good. 2961 I was not really able to test locally since there appears to be a single entry for testing purposes locally. All others I think I was able to test fully locally and appeared good.

PR #1602, opened 2022-07-19
LEAF 2971 Report Builder exports: implement format check

aerinkayne:
Adds the 'data-format' data property to report builder table cells.
A format check is now used in combination with a date-pattern regex so that date format questions are not affected by an earlier changed intended to prevent Excel from auto-formatting non-date input as dates.

### DIFF
--- a/LEAF_Request_Portal/js/formGrid.js
+++ b/LEAF_Request_Portal/js/formGrid.js
@@ -721,7 +721,7 @@ var LeafFormGrid = function(containerID, options) {
 
                 var trimmedText = val.innerText.trim();
                 line[i] = trimmedText;
-                //add \t to data that could be interpretted as dates by excel
+                //add quotes to data that could be interpretted as dates by excel
                 const dataFormat = val.getAttribute('data-format');
                 const testDateFormat = /^\d+[\/-]\d+([\/-]\d+)?$/;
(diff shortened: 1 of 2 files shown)


LEAF_Request_Portal/js/formGrid.js
```diff
@@ -716,9 +721,10 @@ var LeafFormGrid = function(containerID, options) {
 
                 var trimmedText = val.innerText.trim();
                 line[i] = trimmedText;
-                //wrap text that could be interpretted as dates by excel in quotes
+                //add \t to data that could be interpretted as dates by excel
+                const dataFormat = val.getAttribute('data-format');
                 const testDateFormat = /^\d+[\/-]\d+([\/-]\d+)?$/;
-                line[i] = testDateFormat.test(line[i]) ? line[i] + '\t' : line[i];
+                line[i] = dataFormat !== 'date' && testDateFormat.test(line[i]) ? line[i] + '\t' : line[i];
```
pete-nerantzinis, line 727: Would change this back to `"'" + line[i] + "'": line[i];` and get rid of the \t tab that is being metioned

pete-nerantzinis (changes requested):
Made some requests

pete-nerantzinis (approved):

PR #1606, opened 2022-07-21
LEAF-2921: Nexus primary admin Implementation

pete-nerantzinis:


LEAF_Nexus/admin/templates/menu.tpl
```diff
@@ -1,4 +1,34 @@
...
+                            $('#help-primary-admin').html(res['firstName'] + " " + res['lastName'] + emailString);
+                        }
+                        else if(res["userName"] !== undefined)
+                        {
+                            $('#help-primary-admin').html(res['userName']);
+                        }
+                        else
+                        {
+                            $('#help-primary-admin').html('Primary Admin has not been set.');
+                        }
```
Pelentan, line 24: Need fail catch.  Might even need it to be an "error" response to have them reload the page.

LEAF_Nexus/admin/templates/mod_system.tpl
```diff
@@ -83,15 +88,71 @@ function saveSettings()
...
+                url: '../api/system/setPrimaryadmin',
+                data: {userID: $('#primaryAdmin').val(),
+                CSRFToken: CSRFToken},
+                success: function(res) {
+                    if(res['success'] !== true) {
+                        alert('Primary Admin must be a System Administrator');
+                    } else {
+                        location.reload();
+                    }
+                }
```
Pelentan, line 102: fail catch.

LEAF_Nexus/admin/templates/mod_system.tpl
```diff
@@ -83,15 +88,71 @@ function saveSettings()
...
     dialog_confirm = new dialogController('confirm_xhrDialog', 'confirm_xhr', 'confirm_loadIndicator', 'confirm_button_save', 'confirm_button_cancelchange');
 
+    $.ajax({
+        type: 'GET',
+        async: false,
+        url: '../api/system/primaryadmin',
+        success: function(res) {
+            $('.employeeSelectorInput').val('userName:'+res['userName']);
+            primarySet = res;
+        }
```
Pelentan, line 126: fail catch

LEAF_Nexus/admin/templates/setup_medical_center.tpl
```diff
@@ -563,6 +563,9 @@ $(function() {
 	    case 'services':
             setupServices();
             break;
+        case 'primaryAdmin':
```
Pelentan, line 566: need to get rid of need to be case sensitive.

LEAF_Nexus/sources/System.php
```diff
@@ -242,4 +256,90 @@ public function refreshOrgchartEmployees()
 
         return $ret;
     }
+
+    /**
+     * Get primary admin.
+     *
+     * @return array|false array with primary admin's info
+     */
+    public function getPrimaryAdmin()
```
Pelentan, line 265: Please add return type.  public function getPrimaryAdmin(): string

You'll need to change the else return to ""

LEAF_Nexus/sources/System.php
```diff
@@ -242,4 +256,90 @@ public function refreshOrgchartEmployees()
...
+
+        return $resultArray;
+    }
+
+    /**
+     * Unset primary admin.
+     *
+     * @return array array with query response
+     */
+    public function unsetPrimaryAdmin()
```
Pelentan, line 331: return type

LEAF_Nexus/sources/System.php
```diff
@@ -242,4 +256,90 @@ public function refreshOrgchartEmployees()
...
+     */
+    public function getPrimaryAdmin()
+    {
+        $strSQL = "SELECT `data` FROM `settings` WHERE `setting` = 'primaryAdmin'";
+        $primaryAdminRes = $this->db->query($strSQL);
+
+        include_once 'Employee.php';
+        $employee = new Orgchart\Employee($this->db, $this->login);
+
+        if(count($primaryAdminRes))
```
Pelentan, line 273: This needs to a comparison.  if(count($primaryAdminRes) > 0)

LEAF_Nexus/sources/System.php
```diff
@@ -242,4 +256,90 @@ public function refreshOrgchartEmployees()
...
+        {
+            $user = $employee->lookupLogin($primaryAdminRes[0]['data']);
+            $user = $employee->lookupEmpUID($user[0]['empUID']);
```
Pelentan, line 276: Am I missing something?  You're assigning $user and then reassigning it.  Looks like maybe you meant $user[]?

LEAF_Nexus/sources/System.php
```diff
@@ -242,4 +256,90 @@ public function refreshOrgchartEmployees()
...
+        include_once 'Employee.php';
+        $employee = new Orgchart\Employee($this->db, $this->login);
+        $user = $employee->lookupLogin($userID);
+
+        $sqlVars = array(':empUID' => $user[0]['empUID']);
+        $strSQL = "SELECT `empUID` FROM `relation_group_employee`".
+                    "WHERE `groupID` = 1 AND empUID = :empUID";
+        $res = $this->db->prepared_query($strSQL, $sqlVars);
+
+        if(count($res))
```
Pelentan, line 302: Need comparison

shaneodd (changes requested):
After adding a primary admin save triggers an alert of "Primary Admin must be an Administrator" appears that the userid is not being sent/process probably shouldnt run again. Details were discussed with Pete.

LEAF_Nexus/sources/System.php
```diff
@@ -242,4 +256,90 @@ public function refreshOrgchartEmployees()
...
+            $primary = $this->getPrimaryAdmin();
+
+            $this->dataActionLogger->logAction(\DataActions::ADD, \LoggableTypes::PRIMARY_ADMIN, [
+                new \LogItem("settings", "setting", 'primaryAdmin'),
+                new \LogItem("settings", "data", $primary["empUID"], $primary["firstName"].' '.$primary["lastName"]),
+            ]);
+        }
+        else
+        {
+            $resultArray = array('success' => false, 'response' => $res);
```
Pelentan, line 320: I know I harp on "else" statements.  But in this case, any variable you are going to return, you want defined in your "base" code block.  Sooo... I would put this right after you run the select query.  Then just drop the else statement.  Inside the main "if", you assign the different values to the different elements of the array.

LEAF_Nexus/templates/menu.tpl
```diff
@@ -1,6 +1,36 @@
...
+                            }
+                            else if(res["userName"] !== undefined)
+                            {
+                                $('#help-primary-admin').html(res['userName']);
+                            }
+                            else
+                            {
+                                $('#help-primary-admin').html('Primary Admin has not been set.');
+                            }
+
```
Pelentan, line 27: fail catch

Pelentan (changes requested):
We're hitting the brakes anyway.  We should move toward better practices.

LEAF_Nexus/admin/templates/menu.tpl
pete-nerantzinis: Added fail catch, this would annoy users if they were prompted for that on a thing most users dont even hover over. Since its just a menu hover for help.

LEAF_Nexus/admin/templates/mod_system.tpl
pete-nerantzinis: Added fail catch

LEAF_Nexus/admin/templates/setup_medical_center.tpl
pete-nerantzinis: Added case sensitivity to the switch variable

LEAF_Nexus/sources/System.php
pete-nerantzinis: added return type

pete-nerantzinis: Also changed the returns so it stays standard array

LEAF_Nexus/sources/System.php
pete-nerantzinis: added return type

LEAF_Nexus/sources/System.php
pete-nerantzinis: Added comparison

LEAF_Nexus/sources/System.php
pete-nerantzinis: Added comparison

LEAF_Nexus/sources/System.php
pete-nerantzinis: I just implemented the user search changes to include email so we don't have to do this double method calling anymore.

LEAF_Nexus/sources/System.php
pete-nerantzinis: Removed the else and defined the result after the query

LEAF_Nexus/templates/menu.tpl
pete-nerantzinis: added fail catch

Pelentan (approved):
Looks good

shaneodd (approved):
Looks okay to me.

aerinkayne (approved):

PR #1620, opened 2022-08-02
LEAF 3004, IFTHEN: handle conditions with encoded characters

aerinkayne:
Resolves display and behavior issues that occur if a condition's prefill value, or its parent trigger options contain encoded characters (eg, quotes, apostrophes, &).


Pelentan (approved):

maxrichardjr (approved):

pete-nerantzinis (approved):
Just needs thorough testing

PR #1622, opened 2022-08-03
LEAF-2975 Queries WHERE IN  are null/empty

shaneodd:
Add checks to make sure values for in () are actually set before trying to run query. Also updated a few of the select statements to be more inline with coding standards.

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -2212,6 +2212,8 @@ class Form
                     }
 
                     $out[$item['recordID']]['s' . $item['series']]['id' . $item['indicatorID']] = isset($indicatorMasks[$item['indicatorID']]) && $indicatorMasks[$item['indicatorID']] == 1 ? '[protected data]' : $item['data'];
+                    $indFormat = explode("\n", $indicators[$item['indicatorID']]['format'])[0];
+                    $out[$item['recordID']]['s' . $item['series']]['id' . $item['indicatorID'] . '_format'] = $indFormat;
                     if (isset($item['dataOrgchart']))
                     {
                         $out[$item['recordID']]['s' . $item['series']]['id' . $item['indicatorID'] . '_orgchart'] = $item['dataOrgchart'];
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -1821,23 +1821,24 @@ class Form
             }
             $uniqueCategoryIDs = trim($uniqueCategoryIDs, ',');
 
-            $catsInGroups = $this->db->prepared_query(
-                "SELECT * FROM category_privs WHERE categoryID IN ({$uniqueCategoryIDs}) AND readable = 1",
-                array()
-            );
-            if (count($catsInGroups) > 0)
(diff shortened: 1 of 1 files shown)


pete-nerantzinis (approved):
Nice simple way to check, can we get some use cases for this on the dynatrace side?

shaneodd:
> Nice simple way to check, can we get some use cases for this on the dynatrace side?

I am not sure if this is possible. The queries data are obfuscated and all I can really get out of the logs are queries and the stack of files that resulted in this error. I do not see the url that caused it and I do not see any way to look at what data was posted or used in the queries. Which is how is should be since we do not want to leak data out.

LEAF_Request_Portal/form.php
```diff
@@ -3051,135 +3057,142 @@ public function query($inQuery)
...
+                $res2 = $this->db->prepared_query('SELECT recordID, lastStatus, records_step_fulfillment.stepID, fulfillmentTime FROM records
+                        LEFT JOIN records_step_fulfillment USING (recordID)
+                        LEFT JOIN records_workflow_state USING (recordID)
+                        WHERE recordID IN (' . $recordIDs . ')
+                            AND records_workflow_state.stepID IS NULL
+                            AND submitted > 0
+                            AND deleted = 0', array());
+                foreach ($res2 as $item)
+                {
+                    if($data[$item['recordID']]['recordResolutionData']['fulfillmentTime'] == null
```
Pelentan, line 3135: $dfTime = $data[$item['recordID']]['recordResolutionData']['fulfillmentTime'];
if (empty($dfTime) || $dfTime < $item['fulfillmentTime'] {
   $dfTime = $item['fulfillmentTime'];
   ...
   
Basic rule of thumb, if you use an interpreted variable more than once, you generally want to create a separate variable for it.  Makes it not only "cheaper" but easier to read/change by those who follow on.

LEAF_Request_Portal/form.php
```diff
@@ -3051,135 +3057,142 @@ public function query($inQuery)
...
-                      "LEFT JOIN records USING (recordID) ".
-                      "INNER JOIN workflow_routes USING (stepID) ".
-                      "LEFT JOIN records_workflow_state USING (recordID) ".
-                      "WHERE recordID IN ($recordIDs) ".
-                        "AND action_history.actionType = workflow_routes.actionType ".
-                        "AND records_workflow_state.stepID IS NULL ".
-                        "AND nextStepID = 0 ".
-                        "AND submitted > 0 ".
-                        "AND deleted = 0";
+                $strSQL = "SELECT recordID, action_history.userID as resolvedBy, action_history.stepID, action_history.actionType FROM action_history ".
```
Pelentan, line 3147: I had been of two minds going through the code on this.  Since your going to be making change anyway, please go back and set up all the sql calls you touched to this method:  define the call in a string then use it in the db call like here.

Pelentan (changes requested):

Pelentan (approved):

raphaelfretter (approved):

aerinkayne (approved):

PR #1627, opened 2022-08-17
Speed up admin/?a=mod_groups page load

mgaoVA:
On sites with large numbers of groups and members, it takes too long to open the User Access Groups page.

Part of the reason for this seems to be that api/group/members is returning redundant data: it returns a list of all groups + all members + **all groups for each member**.

Since "all groups for each member" can be derived from all groups + all members, this patch removes the redundancy.

### DIFF
--- a/LEAF_Request_Portal/admin/Group.php
+++ b/LEAF_Request_Portal/admin/Group.php
@@ -71,7 +71,7 @@ class Group
         return 'Cannot remove group.';
     }
 
-    // return array of userIDs
+    // getMembers returns a list of members associated with $groupID
     public function getMembers($groupID)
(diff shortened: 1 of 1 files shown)


maxrichardjr (approved):

AmandaSumner (approved):

raphaelfretter (approved):

aerinkayne (approved):

shaneodd (approved):
Since we do not have a ticket for this one, having the number of people that it takes to slow down might be useful. Or having which portal you notice this on would help, testing team could focus on that portal to make sure there are no un indented consequences.

PR #1649, opened 2022-09-06
Bug leaf 3066 xsshelpers

aerinkayne:
Assigns default values to parameters used in xsshelpers, and wraps calls to methods like htmlspecialchars with a check for string type


AmandaSumner (approved):

shaneodd (approved):

raphaelfretter (approved):

libs/php-commons/XSSHelpers.php
```diff
@@ -57,179 +65,180 @@ public static function xscrub($data)
...
-                //echo "examining: {$matches[1][$i]}{$matches[2][$i]}\n";
-                // proper closure
-                if ($matches[1][$i] == '/' && isset($openTags[$matches[2][$i]]) && $openTags[$matches[2][$i]] > 0)
-                {
-                    $openTags[$matches[2][$i]]--;
-                // echo "proper\n";
-                }
-                // new open tag
-                else
+                if ($matches[2][$i] != 'br'
```
Pelentan, line 193: Might want to consider changing this to an in-array conditional check.  You're checking the same value against three different strings.  a list that could conceivably grow.  And do you need to check/ensure case here?

PR #1665, opened 2022-09-14
LEAF-3097 Report builder export memory

shaneodd:
use what I used for the download zip in LEAF-2465, I think the append to the body is what is causing the issue here which is why I went with what I did in the zip file download. It does appear to be when the browser appends the data to the body the tab checks out.


Pelentan (approved):

pete-nerantzinis (approved):

aerinkayne (approved):

shaneodd:
This did not fully fix what is going on, tested with another portal and this did not take care of the issue. Added notes to task.

PR #1667, opened 2022-09-15
LEAF 2465 - National Download All Files

shaneodd:
Feature/leaf 2465/download all files clean

This is for a national, this does not have permission checks required to operate on normal portals. This is set specifically for this specific national.

### DIFF
--- a/LEAF_Request_Portal/api/controllers/FormController.php
+++ b/LEAF_Request_Portal/api/controllers/FormController.php
@@ -32,6 +32,12 @@ class FormController extends RESTfulResponse
 
         $this->index['GET'] = new ControllerMap();
         $cm = $this->index['GET'];
+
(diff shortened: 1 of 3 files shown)


mgaoVA:
> Feature/leaf 2465/download all files clean
> 
> This is for a national, this does not have permission checks required to operate on normal portals. This is set specifically for this specific national.

To reduce maintenance complexity, "non-emergency" sites cannot have server side customizations. If a server-side change is needed, it must be developed in a way that works with all portals. @Pelentan 

Permissions must always be checked on processes that enable users to read data. See file.php for an example of how $form->getIndicator() is used to check permissions when downloading an attachment.

Pelentan (approved):

aerinkayne (approved):

PR #1669, opened 2022-09-20
Implement batch loading in view_search

mgaoVA:
This implements a similar batch loading technique in PR#1657 to divide large queries and provide a better UX.


raphaelfretter (approved):

raphaelfretter (commented):
If the variable value is not changing, please use const instead of let.

maxrichardjr (approved):

aerinkayne (approved):

AmandaSumner (approved):

Pelentan (approved):

PR #1687, opened 2022-10-12
LEAF-3166/2450/3159 Added notes to View History and Report Builder

jampaul3:


LEAF_Request_Portal/View.php
```diff
@@ -21,88 +21,101 @@ public function __construct($db, $login)
...
+
+            foreach ($res as $tmp)
             {
-                $user = $dir->lookupLogin($tmp['userID']);
-                $name = isset($user[0]) ? "{$user[0]['Fname']} {$user[0]['Lname']}" : $tmp['userID'];
-                $packet['userName'] = $name;
+                $packet = [];
+                $packet['time'] = $tmp['time'];
+
+                if ($tmp['description'] == 'Note Added') {
```
Pelentan, line 60: If we are comparing strings, the comparison must be between like cases.  In this case you will want to tolower() the variable and make sure the value compared to is all lower case.  Not likely too important here, but this needs to become the default.

LEAF_Request_Portal/View.php
```diff
@@ -21,88 +21,101 @@ public function __construct($db, $login)
...
+                    $packet['description'] = 'Note Added: ';
+                } else if ($tmp['description'] == '' && $tmp['actionText'] == '' ) {
```
Pelentan, line 62: Instead of looking for empty strings, '',  you'll want to use the empty() function.  PHP8 is much more sensitive to the difference between '', 0, and null.

LEAF_Request_Portal/View.php
```diff
@@ -21,88 +21,101 @@ public function __construct($db, $login)
...
-                $user = $dir->lookupLogin($tmp['userID']);
-                $name = isset($user[0]) ? "{$user[0]['Fname']} {$user[0]['Lname']}" : $tmp['userID'];
-                $packet['userName'] = $name;
+            foreach ($res as $tmp) {
+                $packet = [];
+                $packet['time'] = $tmp['timestamp'];
+                $packet['description'] = $tmp['stepTitle'] . ': Digitally Signed';
+                $packet['comment'] = 'Signature Hash: ' . $tmp['signature'];
+
+                if ($tmp['userID'] != '') {
```
Pelentan, line 95: empty()

LEAF_Request_Portal/View.php
```diff
@@ -21,88 +21,101 @@ public function __construct($db, $login)
...
+                    $user = $dir->lookupLogin($tmp['userID']);
+                    $name = isset($user[0]) ? "{$user[0]['Fname']} {$user[0]['Lname']}" : $tmp['userID'];
+                    $packet['userName'] = $name;
+                }
+
+                $result[] = $packet;
             }
-            $result[] = $packet;
+
+            usort($result, function($a, $b) {
```
Pelentan, line 104: Two things.  1) No nested functions.  2) It doesn't look like it even gets called?

LEAF_Request_Portal/admin/templates/mod_form.tpl
```diff
@@ -1404,7 +1404,7 @@ function getForm(indicatorID, series) {
         }
 
         if(sensitiveChanged){
-            calls.push(            
+            calls.push(
```
Pelentan, line 1407: Guessing all of the below need fail catches.

LEAF_Request_Portal/form.php
```diff
@@ -1187,159 +1187,156 @@ public function doModify($recordID)
...
-                $res = $this->db->prepared_query('SELECT * FROM workflow_routes
-            										WHERE workflowID=:workflowID
-            											AND stepID=-1
-            											AND actionType="submit"', $vars);
-                if (count($res) > 0)
-                {
-                    $hasInitialStep = true;
+            // write new workflow states
+            $vars = array(':recordID' => $recordID);
+            $res = $this->db->prepared_query('SELECT * FROM category_count
```
Pelentan, line 1219: No select *

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -106,70 +123,107 @@
...
-		type: 'POST',
-		url: "./api/form/" + recordID + "/submit",
-		data: {CSRFToken: '<!--{$CSRFToken}-->'},
-		success: function(response) {
+    $('#submitControl').empty().html('<img src="./images/indicator.gif" />Submitting...');
+    $.ajax({
+        type: 'POST',
+        url: "./api/form/" + recordID + "/submit",
+        data: {CSRFToken: '<!--{$CSRFToken}-->'},
+        success: function(response) {
```
Pelentan, line 147: fail catch

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -106,70 +123,107 @@
...
-		}
-	});
+        }
+    });
+}
+
+function submitNote(recordID){
+    var form = $("#note_form").serialize();
+
+    $.ajax({
```
Pelentan, line 169: fail catch

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -106,70 +123,107 @@
...
+        CSRFToken: '<!--{$CSRFToken}-->'},
+        success: function(response) {
+            $("#note").val('');
+
+            addNote(response);
+        }
+    });
+}
+
+function addNote(response) {
```
Pelentan, line 182: default value

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -106,70 +123,107 @@
...
-			if(res.length > 0) {
-				buffer = res.length + ' Bookmarks'
-			}
-
-			$('#tags').empty().html(buffer);
-			$('#tags').fadeIn(250);
-		},
-		cache: false
-	});
+    $('#tags').fadeOut(250);
```
Pelentan, line 195: fail catch

Pelentan (changes requested):

LEAF_Request_Portal/View.php
jampaul3: I changed it here. Might just as well with the other changes.

LEAF_Request_Portal/View.php
jampaul3: I wasn't sure what this was even. I just left it because I didn't want to mess anything up.

LEAF_Request_Portal/admin/templates/mod_form.tpl
shaneodd: I think this file probably should not even be in this commit. But this is something I have to look at for my code and something to consider for the code around what I am working currently.

LEAF_Request_Portal/View.php
```diff
@@ -21,88 +21,101 @@ public function __construct($db, $login)
         $this->login = $login;
     }
 
-    public function buildViewStatus($recordID)
+    public function buildViewStatus(int $recordID): array
```
shaneodd, line 24: PHPdocs?

LEAF_Request_Portal/admin/templates/mod_form.tpl
jampaul3: You are probably right. Looks like the only changes are for spaces. I must have opened it and made a small change then changed it back or something and the space changes stuck. I'm sure VSCode removed the extra spaces when I saved.

LEAF_Request_Portal/View.php
jampaul3: I think this is sorting the result array. I pulled it out into a separate static function

Pelentan (approved):

pete-nerantzinis (approved):

aerinkayne (approved):

AmandaSumner (approved):

PR #1699, opened 2022-10-26
LEAF-3198 auto loading classes

jampaul3:
Removed all includes and requires for classes.
Added one require for the LEAF_autoloader.


shaneodd (changes requested):

LEAF_Nexus/scripts/convertFileNames.php
```diff
@@ -1,106 +0,0 @@
-<?php
```
raphaelfretter, line 1: Is this file no longer needed?

LEAF_Request_Portal/db_mysql.php
```diff
@@ -1,347 +0,0 @@
-<?php
```
raphaelfretter, line 1: Not sure about this either.

raphaelfretter (commented):
It's finally done. Just a couple questions about some files.

LEAF_Request_Portal/db_mysql.php
jampaul3: This one should be deleted. When I was going through every file, and noticing an issue with what db file was being used and then noticing there really wasn't much of a difference between the two files (Nexus and Portal) I decided to combine them. That made the two obsolete so yes I will delete this one.

LEAF_Nexus/scripts/convertFileNames.php
jampaul3: You can double check but when I search for the file name it does not come up. So best as I can tell it is not used anywhere.

raphaelfretter (approved):

shaneodd (approved):

PR #1700, opened 2022-10-26
LEAF-2452 - copy existing record by

shaneodd:
On a record, there is now a new button under tools called Duplicate From. This will have a popup that will allow a user to choose a new title, service, adjust priority and choose which sections of the record to be copied over.


LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -461,42 +484,296 @@ function cancelRequest() {
...
         });
     });
 }
 
+/**
+ *
+ * @param {object} indicators
+ * @returns {array}
+ */
+function getChildrenIndicatorIDs(indicators) {
```
Pelentan, line 519: You check to see if it's null, but you don't verify that it's an object with the needed key-values.  How will javascript react if "indicators" is "plaid"?

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -461,42 +484,296 @@ function cancelRequest() {
...
+                success: function (res) {
+                    Object.values(res).forEach(function (resultValue) {
+
+                        if (pickAndChooseValues.includes(resultValue[series].indicatorID)) {
+
+                            // uploaded files will need to have a special case done to them to copy them over to the new record
+                            if(resultValue[series].format == 'fileupload' || resultValue[series].format == 'image'){
+                                if(resultValue[series].value.length > 0){
+                                    resultValue[series].value.forEach(function(currentFile){
+                                        var fileDat = {
```
Pelentan, line 669: Why is this a var instead of let?  Looks like it is only used here.

Pelentan (changes requested):

jampaul3 (approved):

LEAF_Request_Portal/templates/print_form.tpl
shaneodd: This was something I missed, was testing something and after I realized what I was doing I forgot to change it back.

aerinkayne (approved):

pete-nerantzinis (approved):

Pelentan (approved):

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -461,42 +484,302 @@ function cancelRequest() {
...
+
+            // make sure indicatorID exists
+            if(indicator.indicatorID !== undefined){
+                children.push(indicator.indicatorID);
+            }
+
+            // make sure child exists
+            if (indicator.child !== undefined) {
+                let subchildren = getChildrenIndicatorIDs(indicator.child);
+                // well this took me a bit to realize concat returns the value of results
```
raphaelfretter, line 533: I don't know if this is a necessary comment lol

raphaelfretter (changes requested):

AmandaSumner (approved):

maxrichardjr (approved):

PR #1712, opened 2022-11-03
LEAF 3119 - automated email notifications

shaneodd:
This will require a database update for storage of data, timestamp for determining when to trigger reminders and a new db entry for the templates required for this reminder. 

Setup:
- Go to the workflow and click on a step. 
- There is a new button on the bottom right called "Email Reminder"
- In this new popup you can enable or disable the reminder using the checkbox.
- This checkbox will give you some options on days/weeks/months
- The users that will be notified will be the same as choosing "notify next approver" in the event triggers

Testing:
/LEAF_Request_Portal/scripts/automated_email.php will run the process as expected, waiting for  the time to expire
/LEAF_Request_Portal/scripts/automated_email.php?current=current will trigger the email process on items that are setup for reminders instead of the time set.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_workflow.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_workflow.tpl
@@ -187,6 +187,305 @@ function groupListContent(groups) {
     return content;
 }
 
+// Automated Emails Section
+function emptyAlert(idElement) {
+    const elementId = document.getElementById(idElement).value;
+    if (elementId == "") {
+        alert("Please make a choice");
+    }
+}
+
+function createElement(typeOfElement, elementId, parentDiv) {
+    const newElement = document.createElement(typeOfElement);
+    newElement.setAttribute("id", elementId);
+    document.getElementById(parentDiv).appendChild(newElement);
+}
+
+function removeChild(removeChildId) {
+    const childId = document.getElementById(removeChildId);
+    if (childId) {
+        childId.removeChild(childId.firstElementChild);
+    }
+}
+
+function removeAllChildren(containerName) {
+    const container = document.getElementById(containerName);
+    while (container.firstChild) {
+        container.removeChild(container.firstChild);
+    }
+}
+
+function insertDOM(elementId, content) {
+    document.getElementById(elementId).innerHTML = content;
+}
+
+/**
+ * Email reminder dialog that will be triggered via the step popup
+ * @param stepID
+ */
+function addEmailReminderDialog(stepID){
+    $('.workflowStepInfo').css('display', 'none');
+    let workflowStep = null;
+    $.ajax({
+        type: 'GET',
+        data: {
+            CSRFToken: CSRFToken,
+
+        },
+        url: '../api/workflow/step/' + stepID,
+        async: false,
+        success: function (res) {
+            workflowStep = res
+        }
+    });
+
+    dialog.setTitle('Email Reminder');
+    let output = '<label for="edit_email_check">Automated Emails:</label> <input type="checkbox" id="edit_email_check" onclick="editEmailChecked()"><div id="edit_email_container"></div><br>';
+    dialog.setContent(output);
+    dialog.setSaveHandler(function() {
+
+        let seriesData = {
+            'Automate Email Group': $('#edit_email_check').prop('checked'),
+            'Date Selected': $('#edit_dates_selected').val(),
+            'Days Selected': $('#edit_dates_days').val()
+        }
+
+        $.ajax({
+            type: 'POST',
+            data: {
+                CSRFToken: CSRFToken,
+                seriesData: seriesData
+            },
+            url: '../api/workflow/stepdata/' + stepID,
+            success: function (res) {
+                if (res == 1) {
+                    //loadWorkflow(currentWorkflow);
+                    dialog.hide();
+                }
+                else {
+                    alert(res);
+                }
+            },
+            error: function(){ console.log('Failed to save automated email reminder data'); }
+        });
+    });
+
+    dialog.show();
+
+    // dialog changes if data is setup properly
+    if(workflowStep?.stepData !== null) {
+        let stepParse = JSON.parse(workflowStep.stepData);
+        let automateEmailGroup = stepParse.AutomateEmailGroup;
+        let dateSelected = stepParse.DateSelected;
+        let daysSelected = stepParse.DaysSelected;
+
+        if (automateEmailGroup == "true") {
+            $('#edit_email_check').prop('checked', true);
+            editEmailChecked();
+            $("#edit_dates_selected").val(dateSelected);
+            editEmailDateSelected();
+            $("#edit_dates_days").val(daysSelected);
+
+        } else {
+            $('#edit_email_check').prop('checked', false);
+        }
+    }
+}
+
+///// Creating Automated Emails
+function emailChecked() {
+    let emailChecked = document.getElementById("email_check");
+    if (emailChecked.checked) {
+        let selectdatesString = "";
+        selectdatesString += "<br>";
+        selectdatesString += '<select id="dates_selected" onchange="emailDateSelected()">';
+        selectdatesString += '<option value="">Select Option</option>';
+        selectdatesString += '<option value="1">Day/Days</option>';
+        selectdatesString += '<option value="7">Week/Weeks</option>';
+        selectdatesString += '<option value="30">Month/Months</option>';
+        selectdatesString += "</select>";
+        selectdatesString += "<br>";
+        createElement("div", "date_select", "email_container");
+        document.getElementById("date_select").innerHTML = selectdatesString;
+    } else {
+        removeAllChildren("email_container");
+    }
+}
+function emailDateSelected() {
+    let dates_selected = document.getElementById("dates_selected").value;
+    let days_display = "";
+    switch (dates_selected) {
+        case "1":
+            let days = 31;
+            days_display += "<br>";
+            days_display += '<select id="dates_days" onchange="dateResult()">';
+            days_display += '<option value="">Select Option</option>';
+            for (let index = 1; index < days; index++) {
+                days_display += '<option value="' + index + '">' + index + "</option>";
+            }
+            days_display += "</select>";
+            days_display += "<br>";
+            break;
+        case "7":
+            let weeks = 53;
+            days_display += "<br>";
+            days_display += '<select id="dates_days" onchange="dateResult()">';
+            days_display += '<option value="">Select Option</option>';
+            for (let index = 1; index < weeks; index++) {
+                days_display += '<option value="' + index + '">' + index + "</option>";
+            }
+            days_display += "</select>";
+            days_display += "<br>";
+            break;
+        case "30":
+            let months = 13;
+            days_display += "<br>";
+            days_display += '<select id="dates_days" onchange="dateResult()">';
+            days_display += '<option value="">Select Option</option>';
+            for (let index = 1; index < months; index++) {
+                days_display += '<option value="' + index + '">' + index + "</option>";
+            }
+            days_display += "</select>";
+            days_display += "<br>";
+            break;
+        default:
+            emptyAlert("dates_selected");
+            break;
+    }
+    const daySelect = document.createElement("div");
+    daySelect.setAttribute("id", "date_days");
+    document.getElementById("email_container").appendChild(daySelect);
+    document.getElementById("date_days").innerHTML = days_display;
+}
+function dateResult() {
+    let date_selected = document.getElementById("dates_selected").value;
+    let date_days = document.getElementById("dates_days").value;
+    switch (date_selected) {
+        case "1":
+            let dateResultDays = parseInt(date_selected, 10) + parseInt(date_days, 10);
+            console.log(dateResultDays);
+            const date = new Date();
+            const totalDays = document.createElement("div");
+            totalDays.setAttribute("id", "days_total");
+            document.getElementById("email_container").appendChild(totalDays);
+            document.getElementById("days_total").innerHTML = "<br>Group will be notified every: " + dateResultDays + " days from " + date.getMonth() + "/" + date.getDate() + "/" + date.getFullYear() + ".";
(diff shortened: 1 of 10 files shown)


LEAF_Request_Portal/sources/Workflow.php
```diff
@@ -249,36 +268,36 @@ public function deleteAction($stepID, $nextStepID, $action)
 
         // clear out route events
         $vars = array(':workflowID' => $this->workflowID,
-                      ':stepID' => $stepID,
-                      ':action' => $action, );
+            ':stepID' => $stepID,
+            ':action' => $action, );
         $res = $this->db->prepared_query('DELETE FROM route_events
     										WHERE workflowID=:workflowID
```
raphaelfretter, line 274: Spacing

docker/mysql/db/db_upgrade/portal/Update_RMC_DB_2022050300-2022110300.sql
```diff
@@ -0,0 +1,34 @@
+START TRANSACTION;
```
raphaelfretter, line 1: Do we want this file in the repo?

raphaelfretter (changes requested):

docker/mysql/db/db_upgrade/portal/Update_RMC_DB_2022050300-2022110300.sql
pete-nerantzinis: Yeah we put them here now so they end up in the backend on server.

pete-nerantzinis: Yeah we put them here now so they end up in the backend on server.

pete-nerantzinis (approved):
Looks good on my end I would like to see some use cases of it working if possible

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -187,6 +187,305 @@ function groupListContent(groups) {
...
 
+// Automated Emails Section
+function emptyAlert(idElement) {
+    const elementId = document.getElementById(idElement).value;
+    if (elementId == "") {
+        alert("Please make a choice");
+    }
+}
+
+function createElement(typeOfElement, elementId, parentDiv) {
```
Pelentan, line 198: What kind of error would be generated if typeOfElement is null?

Pelentan: Going to ask you to look over all these functions and make sure not having a default value isn't going to cause unintended consequences.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -187,6 +187,305 @@ function groupListContent(groups) {
...
+    dialog.show();
+
+    // dialog changes if data is setup properly
+    if(workflowStep?.stepData !== null) {
+        let stepParse = JSON.parse(workflowStep.stepData);
+        let automateEmailGroup = stepParse.AutomateEmailGroup;
+        let dateSelected = stepParse.DateSelected;
+        let daysSelected = stepParse.DaysSelected;
+
+        if (automateEmailGroup == "true") {
```
Pelentan, line 282: Remember toLower()

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -187,6 +187,305 @@ function groupListContent(groups) {
...
+        createElement("div", "date_select", "email_container");
+        document.getElementById("date_select").innerHTML = selectdatesString;
+    } else {
+        removeAllChildren("email_container");
+    }
+}
+function emailDateSelected() {
+    let dates_selected = document.getElementById("dates_selected").value;
+    let days_display = "";
+    switch (dates_selected) {
```
Pelentan, line 317: Might want to consider converting this to an int and then all your cases would just be ints.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -187,6 +187,305 @@ function groupListContent(groups) {
...
+    }
+    const daySelect = document.createElement("div");
+    daySelect.setAttribute("id", "date_days");
+    document.getElementById("email_container").appendChild(daySelect);
+    document.getElementById("date_days").innerHTML = days_display;
+}
+function dateResult() {
+    let date_selected = document.getElementById("dates_selected").value;
+    let date_days = document.getElementById("dates_days").value;
+    switch (date_selected) {
```
Pelentan, line 363: Same for here.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -187,6 +187,305 @@ function groupListContent(groups) {
...
+        createElement("div", "edit_date_select", "edit_email_container");
+        document.getElementById("edit_date_select").innerHTML = editSelectdatesString;
+    } else {
+        removeAllChildren("edit_email_container");
+    }
+}
+function editEmailDateSelected() {
+    let dates_selected = document.getElementById("edit_dates_selected").value;
+    let days_display = "";
+    switch (dates_selected) {
```
Pelentan, line 414: And here.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -187,6 +187,305 @@ function groupListContent(groups) {
...
+    const daySelect = document.createElement("div");
+    daySelect.setAttribute("id", "edit_date_days");
+    document.getElementById("edit_email_container").appendChild(daySelect);
+    document.getElementById("edit_date_days").innerHTML = days_display;
+}
+function editDateResult() {
+    let date_selected = document.getElementById("edit_dates_selected").value;
+    let date_days = document.getElementById("edit_dates_days").value;
+    const date = new Date();
+    switch (date_selected) {
```
Pelentan, line 461: Ditto

LEAF_Request_Portal/sources/Workflow.php
```diff
@@ -103,15 +103,34 @@ public function deleteStep($stepID)
...
-        ]);  
+        ]);
 
         return 1;
     }
 
+    public function getStep(int $stepID): array
+    {
+        $vars = array(':stepID' => $stepID);
+        $workflowStepsRes = $this->db->prepared_query('SELECT
```
Pelentan, line 118: The query itself should be pulled out and assigned to a variable which is used here.

LEAF_Request_Portal/sources/Workflow.php
```diff
@@ -529,20 +551,49 @@ public function updateStep($stepID, $stepTitle, $bgColor = '', $fontColor = '')
...
             new LogItem("workflows", "stepTitle",  $stepTitle),
             new LogItem("workflows", "jsSrc",  "", "empty"),
             new LogItem("workflow_steps", "workflowID", $this->getWorkflowIDFromStep($stepID))
-        ]);    
+        ]);
+
+        return 1;
+    }
+
+    public function saveStepData($stepID, $data){
```
Pelentan, line 570: Need this function cleaned up.  Best practices including only a single return.  Would prefer a single return type as well.  Letting the calling function determine what to do with the int result.

Pelentan (changes requested):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -216,7 +218,9 @@ function removeAllChildren(containerName) {
 }
 
 function insertDOM(elementId, content) {
-    document.getElementById(elementId).innerHTML = content;
+    if(elementId){
```
Pelentan, line 221: Pretty sure this isn't boolean.

Pelentan (changes requested):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -196,9 +196,11 @@ function emptyAlert(idElement) {
 }
 
 function createElement(typeOfElement, elementId, parentDiv) {
-    const newElement = document.createElement(typeOfElement);
-    newElement.setAttribute("id", elementId);
-    document.getElementById(parentDiv).appendChild(newElement);
+    if(typeOfElement && elementId && parentDiv) {
```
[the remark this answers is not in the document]
shaneodd: Changed this to make sure types are strings when coming in, this is what everything requires.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
shaneodd: added check to make sure this is the proper type.

jrodcode (approved):

aerinkayne (approved):

raphaelfretter (approved):

Pelentan (approved):

raphaelfretter (approved):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -187,6 +187,305 @@ function groupListContent(groups) {
...
+                CSRFToken: CSRFToken,
+                seriesData: seriesData
+            },
+            url: '../api/workflow/stepdata/' + stepID,
+            success: function (res) {
+                if (res == 1) {
+                    //loadWorkflow(currentWorkflow);
+                    dialog.hide();
+                }
+                else {
```
jampaul3, line 275: Are we following PSR-12? if so, that standard requires the } else { or } elseif { to be on the same line.

LEAF_Request_Portal/scripts/automated_email.php
```diff
@@ -0,0 +1,121 @@
...
+
+// this was what the random function I found used.
+$comment = '';
+
+$db_config = new DB_Config();
+$db = new DB($db_config->dbHost, $db_config->dbUser, $db_config->dbPass, $db_config->dbName);
+
+$getWorkflowStepsRes = $db->prepared_query('SELECT workflowID, stepID,stepTitle,stepData
+FROM workflow_steps WHERE stepData LIKE \'%"AutomateEmailGroup":"true"%\';', []);
+
```
jampaul3, line 27: Aren't we pulling the sql statement out of the prepared_query function into a variable and just using the variable inside the function?

LEAF_Request_Portal/sources/Workflow.php
```diff
@@ -567,32 +568,36 @@ public function updateStep($stepID, $stepTitle, $bgColor = '', $fontColor = '')
...
 
-    public function saveStepData($stepID, $data){
+    public function saveStepData(int $stepID, array $data) : bool{
+
+        $validSaveStep = TRUE;
 
         // everything that seems to modify this stuff is run through here.
         if (!$this->login->checkGroup(1))
         {
-            return 'Admin access required.';
```
jampaul3: PSR12 requires the { to be on the same line as the conditional statement. Unless, the conditions are on separate lines in which case the ) { need to be on there own line following the conditions.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -1877,7 +1823,18 @@ function loadWorkflow(workflowID) {
             	if(posY < minY) {
             		posY = minY;
             	}
-            	$('#workflow').append('<div class="workflowStep" id="step_'+ res[i].stepID +'">'+ res[i].stepTitle +'</div><div class="workflowStepInfo" id="stepInfo_'+ res[i].stepID +'"></div>');
+
+                let emailNotificationIcon = '';
+                if(typeof res[i].stepData == 'string') {
+                    let stepParse = JSON.parse(res[i].stepData);
+                    if(stepParse.AutomateEmailGroup === 'true'){
+                        let dayCount = stepParse.DateSelected * stepParse.DaysSelected;
```
jampaul3, line 1831: PSR12 requires a space between if ( and a space between ) {

LEAF_Request_Portal/sources/Workflow.php
```diff
@@ -567,32 +568,36 @@ public function updateStep($stepID, $stepTitle, $bgColor = '', $fontColor = '')
         return 1;
     }
 
-    public function saveStepData($stepID, $data){
+    public function saveStepData(int $stepID, array $data) : bool{
+
```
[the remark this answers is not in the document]
Pelentan: In this specific instance PSR12 is wrong.  For functions the opening bracket needs to be on the same line as the declarative.   Else it starts to separate in the human mind from the control.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
Pelentan: And the Linters scream bloody murder if there aren't spaces.  You would think you kicked there dog.   But in this case I agree:  Helps readability.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
shaneodd: Changed this and everything else I could find.

LEAF_Request_Portal/scripts/automated_email.php
shaneodd: Moved this around a bit, the variable was a bit tricky since we are dealing with quotes. This seemed like the only way to format it the way I needed it. This string is moved into a separate variable for use here, variable in the query stayed in that string.

LEAF_Request_Portal/sources/Workflow.php
shaneodd: Updated this and any other entries I could find.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
shaneodd: I was half tempted to just use the auto formatter on this and let it eat. But I also do not want to blow up the change log.

Pelentan (approved):

jampaul3 (approved):

PR #1713, opened 2022-11-03
LEAF-3231 Update orgchart_* widget customization

mgaoVA:
This implements JS promises in orgchart_* widgets, intended to be used within customizations. Promises must be used since the widget can potentially load asynchronously.

Example to disable the standard orgchart_group widget:
```js
    let fieldID = 5;
    leaf_groupSelector[fieldID].then(grpSel => {
    	grpSel.disableSearch();
    });
```


maxrichardjr (approved):

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -858,8 +858,8 @@
                 }
             };
             <!--{/if}-->
-            if(grpSel != undefined) {
-                var grpSel = {};
+            if(typeof leaf_groupSelector == 'undefined') {
+                var leaf_groupSelector = {};
```
raphaelfretter, line 862: Did we want lets here? Don't know if this could cause issues in scoping

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -948,8 +949,8 @@
             <div id="posSel_<!--{$indicator.indicatorID|strip_tags}-->"></div>
             <input id="<!--{$indicator.indicatorID|strip_tags}-->" name="<!--{$indicator.indicatorID|strip_tags}-->" style="visibility: hidden" />
             <script>
-            if(posSel == undefined) {
-                var posSel = {};
+            if(typeof leaf_positionSelector == 'undefined') {
+                var leaf_positionSelector = {};
```
raphaelfretter, line 953: Same here

LEAF_Request_Portal/templates/subindicators.tpl
```diff
@@ -1015,8 +1017,8 @@
             <input id="<!--{$indicator.indicatorID|strip_tags}-->" name="<!--{$indicator.indicatorID|strip_tags}-->" value="<!--{$indicator.value|sanitize}-->" style="display: none" />
 
             <script>
-            if(empSel == undefined) {
-                var empSel = {};
+            if(typeof leaf_employeeSelector == 'undefined') {
+                var leaf_employeeSelector = {};
```
raphaelfretter, line 1021: Here too

raphaelfretter (changes requested):

LEAF_Request_Portal/templates/subindicators.tpl
mgaoVA: The scoping rule was a bit confusing, and I'm not sure of the best way to express this. Since this block could potentially be included in the DOM many times, "let" would flag a variable redeclaration error.

The intent here is to make leaf_groupSelector available as a global variable. And when the block is included again, it shouldn't wipe the contents of the variable.

Would this syntax make sense? (Just committed in this PR as [6a88fb5](https://github.com/department-of-veterans-affairs/LEAF/pull/1713/commits/6a88fb5d2874afb0fb645f1c532b40de0ccd0c79))
```js
var leaf_groupSelector = leaf_groupSelector ?? {};
```

shaneodd (approved):

AmandaSumner (approved):

LEAF_Request_Portal/templates/subindicators.tpl
raphaelfretter: Yes this would work better.

raphaelfretter (approved):

Pelentan (approved):

pete-nerantzinis (approved):

aerinkayne (approved):

PR #1724, opened 2022-11-14
LEAF_2601 Delete Step from workflow

jrodcode:
Ticket 2601 - This will help delete the initial step from a workflow, and detach it from the records_workflow_state.

### DIFF
--- a/LEAF_Request_Portal/form.php
+++ b/LEAF_Request_Portal/form.php
@@ -789,56 +789,57 @@ class Form
 
     /**
      * Delete file/image attachment
+     *
      * @param int $recordID
      * @param int $indicatorID
      * @param int $series
-     * @return int 1 for success, 0 for fail
+     * @param string $fileName
+     *
+     * @return bool
+     *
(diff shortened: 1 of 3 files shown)


LEAF_Request_Portal/form.php
```diff
@@ -1357,48 +1358,131 @@ public function doSubmit(int $recordID): int|array
...
-            $resRecord[strtolower($record['categoryID'])] = $record['count'];
+                                                    WHERE recordID=:recordID', $vars);                         
+        $isSubmitted = false;
+        foreach ($resRecordInfoEachForm as $request) {
+            if ($request['submitted'] > 0) {
+                $isSubmitted = true;
+                break;
+            } 
         }
+        if ($isSubmitted) {
```
raphaelfretter, line 1382: Not sure if we want this logic in here. Just throwing it out there in case anyone else can add anything.

jrodcode:
For some reason the code did not update, so that's why I did a new PR.

raphaelfretter (approved):

PR #1725, opened 2022-11-14
LEAF-2601 Delete Step from workflow

jrodcode:
Ticket 2601 - This will help delete the initial step from a workflow, and detach it from the records_workflow_state.

### DIFF
--- a/LEAF_Request_Portal/sources/Workflow.php
+++ b/LEAF_Request_Portal/sources/Workflow.php
@@ -71,10 +71,8 @@ class Workflow
         return $out;
     }
 
-    public function deleteStep($stepID)
-    {
-        if (!$this->login->checkGroup(1))
-        {
+    public function deleteStep($stepID){
+        if (!$this->login->checkGroup(1)){
             return 'Admin access required.';
         }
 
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/sources/Workflow.php
```diff
@@ -71,10 +71,8 @@ public function getSteps()
         return $out;
     }
 
-    public function deleteStep($stepID)
-    {
-        if (!$this->login->checkGroup(1))
-        {
+    public function deleteStep($stepID){
```
raphaelfretter, line 74:
```suggestion
    public function deleteStep(int $stepID): int|string {
```

raphaelfretter (changes requested):

LEAF_Request_Portal/sources/Workflow.php
jrodcode: Great addition, I tested it and it works well! thank you.

shaneodd (approved):

raphaelfretter (approved):

AmandaSumner (approved):

aerinkayne (commented):
From the Jira description, it looks like workflow step deletion should only succeed when there are records at the step if all those records are in a deleted state.  My understanding is that if there are any active requests at the step that those would still need to be manually deleted by initiators before the step can be removed.

Pelentan (approved):

aerinkayne (changes requested):
We should discuss some questions about this in more detail

PR #1757, opened 2022-11-18
LEAF-3296 - fix input type

mgaoVA:
Fixes an issue with non-selfclosing textarea tag


aerinkayne (approved):

pete-nerantzinis (approved):

maxrichardjr (approved):

raphaelfretter (approved):

jampaul3 (approved):

Pelentan (approved):

shaneodd (changes requested):
Would this also need to update the edit form as well? That would still have textareas as well.

AmandaSumner (approved):

shaneodd (approved):
Discussed with @mgaoVA if we should also update the edit side of this as well and did the update there as well.

PR #1767, opened 2022-12-01
Leaf-3198 step 1 arrange classes in class folders

jampaul3:
Step 1 was to get all the class files into class folders, rather than have them placed willy nilly. All classes are now in sources and in the api folders. I didn't really change the api and controller folders. In moving the files I had to make sure they were being included correctly as well. So all files that uses a moved class file has changed as well.


maxrichardjr (approved):
I realize it was largely changes to path for mySQL and config.php, but it was a ton of changes, 83 files!!

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -153,6 +153,7 @@ function doSubmit(recordID) {
 		data: {CSRFToken: '<!--{$CSRFToken}-->'},
 		success: function(response) {
             if(response.errors.length == 0) {
+                console.log('it was a success with NO errors');
```
raphaelfretter, line 156: Should not need logs here for production.

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -161,6 +162,7 @@ function doSubmit(recordID) {
                 workflow.getWorkflow(recordID);
             }
             else {
+                console.log('it was a succes with errors');
```
raphaelfretter, line 165: Here too.

raphaelfretter (changes requested):

shaneodd (approved):

LEAF_Request_Portal/templates/print_form.tpl
jampaul3: removed

raphaelfretter (approved):

aerinkayne (approved):

PR #1769, opened 2022-12-02
Leaf-3311 - Autoloader Step 2 - All classes have their own file

jampaul3:
Step 2 - this took all the files that had multiple class declarations in them and separated them into their own file. It also changed some names of files to match their class name. Also made sure that the newly created file is included where it needs to be so that it is available when needed.


maxrichardjr (approved):

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -310,17 +310,22 @@ var LeafFormGrid = function(containerID, options) {
         var idKey = 'id' + key;
         var tDate;
         for(let i in currentData) {
+            console.log(i);
```
raphaelfretter, line 313: Probably don't need these console logs.

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -310,17 +310,22 @@ var LeafFormGrid = function(containerID, options) {
...
+            console.log(currentData);
```
raphaelfretter, line 314: Same here.

LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -519,8 +519,6 @@ public function handleAction($dependencyID, $actionType, $comment)
                     break;
                 default:
                     return array('status' => 0, 'errors' => array('Invalid Operation'));
-
-                    break;
```
raphaelfretter, line 523: Do we want to leave breaks on defaults or can we remove them?

raphaelfretter (changes requested):

LEAF_Request_Portal/sources/FormWorkflow.php
jampaul3: This is removed because the code can never get to it. I stops with the return. If the return isn't there I would say you'd want he break.

maxrichardjr (approved):

shaneodd (approved):

LEAF_Request_Portal/js/formGrid.js
jampaul3: removed

LEAF_Request_Portal/js/formGrid.js
jampaul3: removed

raphaelfretter (approved):

AmandaSumner (approved):

aerinkayne (approved):

PR #1773, opened 2022-12-06
Grooming/leaf 3310/autoloader step 3 adding namespaces

jampaul3:
Step 3 in the autoloader update places namespaces in all LEAF originating files. Their are 3 namespaces used, Orgchart, which is original meaning it existed before these changes occurred, however, there were a few files that did not include this namespace. Portal was added for portal side classes and Leaf was added to Leaf wide class, for example Db.php has replaced the the db_mysql.php files in the portal and nexus folders. Db.php is located in the libs/php-commons folder and has the Leaf namespace. There are a few more classes that use this namespace as well, most, if not all of which are located in the libs folder somewhere.

All references to these classes have been updated to include the namespace where needed.


LEAF_Nexus/api/controllers/Group/GET.php
```diff
@@ -47,7 +47,7 @@
         $offset = (int)$_GET['offset'];
     }
 
-    return $group->listGroupEmployeesDetailed($args[0], $searchText, $offset, $limit);
+    //return $group->listGroupEmployeesDetailed($args[0], $searchText, $offset, $limit);
```
shaneodd, line 50: Is there a reason to comment this line out?

LEAF_Nexus/api/controllers/NationalEmployeeController.php
```diff
@@ -24,11 +27,12 @@ public function get($act)
...
         $this->index['GET'] = new ControllerMap();
         $this->index['GET']->register('national/employee/version', function () {
-            return $this->API_VERSION;
+            return self::API_VERSION;
         });
 
         $this->index['GET']->register('national/employee/[digit]', function ($args) use ($employee) {
-            return $employee->getSummary($args[0]);
+            // getSummary does not exist in the NationalEmployee class
+            // return $employee->getSummary($args[0]);
```
shaneodd, line 35: Also here is another that was commented out on the return.

shaneodd (changes requested):
There are quite a few return statements that are commented out. One of the comments talks about something not existing. A bit concerning since we do not want to remove functionality.

LEAF_Nexus/api/controllers/Group/GET.php
jampaul3: I'm not sure why I commented this out. I uncommented it.

LEAF_Nexus/api/controllers/NationalEmployeeController.php
jampaul3: Where I put this particular comment, I observed that the method does not exist in class it is being called from. I can uncomment it but it is going to throw an error if someone is trying to use this endpoint.

jampaul3: I believe I commented it out because VSCode shows it as a critical error, meaning that it would throw a fatal if not corrected. So to make this code work we would need to add the method to the NationalEmployee class.

LEAF_Nexus/sources/Group.php
```diff
@@ -522,13 +522,14 @@ public function listGroupPositions($groupID)
     /**
      * Get all employees explicitly associated with a group
      * @param int $groupID
-     * @return array
+     *
```
raphaelfretter, line 525: Need these two blank lines?

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1690,8 +1692,8 @@ public function hasReadAccess($recordID)
 
     /**
      * Checks if the current user has access to a particular dependency
-     * @param dependencyID
-     * @param details - Associative Array containing dependency-specific details, eg: $details['groupID']
+     * @param int $dependencyID
+     * @param array $details - Associative Array containing dependency-specific details, eg: $details['groupID']
      * @return boolean
      */
     public function hasDependencyAccess($dependencyID, $details)
```
raphaelfretter, line 1699: This should have param datatypes and return datatype

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1782,9 +1784,7 @@ public function hasDependencyAccess($dependencyID, $details)
 
                     return $this->checkIfBackup($empUID);
                  }
-
-
-                break;
+                 // unreachable code so no break
```
raphaelfretter, line 1787: Maybe the whole case should be commented out if it's not reachable.

LEAF_Request_Portal/sources/Form.php
```diff
@@ -3900,9 +3900,10 @@ private function isIndicatorOrphan($indicator, &$indicatorList)
      * @param int $recordID
      * @param int $newRecordID
      * @param int $series
-     * @return int 1 for success, errors for failure
+     * @return array|int 1 for success, errors for failure
      */
-    public function copyAttachment($indicatorID, $fileName, $recordID, $newRecordID, $series) {
+    public function copyAttachment($indicatorID, $fileName, $recordID, $newRecordID, $series): array|int
```
raphaelfretter, line 3905: datatypes for params

LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -65,7 +67,7 @@ public function isActive()
     /**
      * Retrieves current steps of a form's workflow, controls access to steps
      * @return array database result
-     * @return null if no database result
+     * @return array|int|null if no database result
      */
     public function getCurrentSteps()
```
raphaelfretter, line 72: Return params

raphaelfretter (changes requested):

AmandaSumner (approved):

LEAF_Nexus/utils/jsonExport_PDL.php
```diff
@@ -108,7 +108,7 @@
     if (is_numeric($data[17]['data']) // current fte
         && is_numeric($data[19]['data']))
     { // total headcount
-        $output[$pos['positionID']]['data']['Current FTE'] = $data[19]['data'] == 0 ? 0 : round($data[17]['data'] / $data[19]['data'], 5);
+        $output[$pos['positionID']]['data']['Current FTE'] = $data[19]['data'] == 0 ? 0 : round($data[17]['daXSSHelpersta'] / $data[19]['data'], 5);
```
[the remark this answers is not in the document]
jampaul3: Thanks for this catch. I saw it and it didn't make sense to me but didn't want to mess anything up so left it alone. But I'm glad you saw it.

LEAF_Request_Portal/scripts/VAMC_directory_maintenance_AD.php
```diff
@@ -119,7 +121,7 @@ public function query($sql):void
             $res = $this->db->prepared_query($sql, array());
             if (is_object($res))
             {
-                return $res->fetchAll(PDO::FETCH_ASSOC);
+                return $res->fetchAll(\PDO::FETCH_ASSOC);
```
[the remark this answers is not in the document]
jampaul3: I really wasn't cleaning anything up with this ticket. If I had I could very easily have gone down a bunch of rabbit holes and basically re-written the entire system. If I remove this return then have to add an else or a blank return, as I assume the code that follows this if statement should not be run if this is true. I'm going to leave it for now as I feel it is out of the scope of this ticket.

LEAF_Request_Portal/sources/Form.php
jampaul3: the break was the only thing not reachable. in the if statement just before this there is a return statement in both the true and false sections. So you would never get to the break. I put the comment here so that in the future when we fix these files and only have one return per method we wouldn't forget to put the break back in.

raphaelfretter (approved):

LEAF_Request_Portal/sources/Form.php
jampaul3: I know I added some here and there but it really is outside the scope of this ticket.

LEAF_Request_Portal/sources/FormWorkflow.php
jampaul3: Again, scope. I probably added it here because of VSCode not really liking just null in this case.

LEAF_Nexus/sources/Group.php
jampaul3: Removed one of them

aerinkayne (approved):

LEAF_Nexus/api/RESTfulResponse.php
```diff
@@ -79,9 +87,14 @@ public function handler($action)
...
-     * @param string $out
+     *
+     * @param string|array $out
+     *
+     * @return void
+     *
+     * Created at: 12/2/2022, 1:27:24 PM (America/New_York)
      */
-    public function output($out = '')
+    public function output(string|array $out = ''): void
```
aerinkayne, line 97: I didn't notice this initially, but this actually seems to sometimes get null.  This causes an issue when editing/saving a position title (not sure where else).

LEAF_Request_Portal/sources/FormEditor.php
```diff
@@ -831,8 +832,16 @@ public function getFormName($categoryID){
...
-    public function getHistory($filterById)
+    /**
+     *
+     * @param int|null $filterById
+     *
+     * @return array
+     *
+     * Created at: 12/5/2022, 10:45:38 AM (America/New_York)
+     */
+    public function getHistory(?int $filterById): array
```
aerinkayne, line 843: From the log, this was taking a form (category) ID, which should be a string

aerinkayne (commented):
updating this comment, as both the parameter issues that I noticed while navigating on step 4 had been introduced on this branch

shaneodd (approved):

PR #1780, opened 2022-12-08
Leaf 3312 Step 4 of Autoloading

jampaul3:
In this final step of the autoloading feature the autoloader class was added as well as an autoloading file. The autoloading file does exactly what you might expect in autoloading classes. But it also does some pre loading of classes that are used on almost every page. Most notably the Config classes and login classes. It also does some session handling that was previously done in the Login class file. By having these classes instantiated here and then included on all pages this variable will be available to the page it was included on.

Also formerly completed in the Login class was sanitizing of the $_GET variable. This has been moved to the autoloader file as well. 

One more point, the XSSHelper class was added to the Leaf namespace and subsequently updated throughout the  system to look for it in that namespace.


maxrichardjr (approved):

shaneodd (approved):

libs/qrcode/index.php
```diff
@@ -14,15 +14,9 @@
 
     // TODO: Replace this with centrally managed server config variable
```
raphaelfretter, line 15: Should move this to the autoloader class.

raphaelfretter (commented):
Look at previous comment.

AmandaSumner (approved):

aerinkayne (commented):
Noticed a couple issues while navigating on the step 4 branch (as mentioned over teams).  Specific comments for the locations are on step 3

maxrichardjr (approved):

jrodcode (approved):

PR #1788, opened 2022-12-15
LEAF 3350  form entry scroll behavior

aerinkayne:
Adds a call to window.scrollTo after the progress check (form entry page only) to ensure that it is at the top.

Also focuses the top next question button if the form had been progressed through the prev or next buttons (does not affect the form nav menu), so that the tab focus also will be at the beginning of form entry.

### DIFF
--- a/LEAF_Request_Portal/templates/form.tpl
+++ b/LEAF_Request_Portal/templates/form.tpl
@@ -133,6 +133,9 @@ function updateProgress(focusNext=false) {
                 $('#nextQuestion').focus();
             }
         },
+        error: function(err) {
+            console.log('an error occurred during form progress checking', err);
+        },
         cache: false
     });
 }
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/templates/form.tpl
```diff
@@ -114,7 +114,7 @@ function onKeyPressClick(event){
     }
 }
 
-function updateProgress() {
+function updateProgress(focusNext=false) {
     $.ajax({
```
shaneodd, line 118: error handling? I know this is something that we are trying to at least get console log when errors happen.

shaneodd (changes requested):

LEAF_Request_Portal/templates/form.tpl
aerinkayne: agree, I will update later with an error callback

jampaul3 (approved):

maxrichardjr (approved):

pete-nerantzinis (approved):

raphaelfretter (approved):

raphaelfretter (approved):

jrodcode (approved):

PR #1790, opened 2022-12-16
Leaf 3348 - autoloader step 5 prepare for launchpad upgrade

jampaul3:
Step 5 included changes that will be affected by the Launchpad upgrade. In this release config and db_config will no longer be needed. The variables in these three locations were already in the db or they will be with the Launchpad upgrade.

### DIFF
--- a/LEAF_Request_Portal/open.php
+++ b/LEAF_Request_Portal/open.php
@@ -5,7 +5,7 @@
 
 error_reporting(E_ERROR);
 
-include '../libs/loaders/Leaf_autoloader.php';
+require_once '../libs/loaders/Leaf_autoloader.php';
 
 $report = isset($_GET['report']) ? Leaf\XSSHelpers::xscrub($_GET['report']) : '';
 if($report != '') {
--- a/LEAF_Request_Portal/sources/Shortener.php
+++ b/LEAF_Request_Portal/sources/Shortener.php
@@ -54,7 +54,7 @@ class Shortener
         return $res - $this->offset;
     }
 
-    public function getFormQuery($shortUID) {
+    public function getFormQuery($shortUID, $form) {
         $shortID = $this->decodeShortUID($shortUID);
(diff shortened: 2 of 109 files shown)


maxrichardjr (approved):

LEAF_Request_Portal/open.php
```diff
@@ -5,7 +5,7 @@
 
 error_reporting(E_ERROR);
 
-include '../libs/loaders/Leaf_autoloader.php';
+require_once '../libs/loaders/Leaf_autoloader.php';
 
 $report = isset($_GET['report']) ? Leaf\XSSHelpers::xscrub($_GET['report']) : '';
 if($report != '') {
```
aerinkayne, line 11: I am not sure when this was actually introduced, but it looks like at some point the shortener class got shortened...

jampaul3: Found and fixed in step 4, merged into this step also.

jrodcode (approved):

LEAF_Request_Portal/sources/Shortener.php
```diff
@@ -54,7 +54,7 @@ private function decodeShortUID($in)
         return $res - $this->offset;
     }
 
-    public function getFormQuery($shortUID) {
+    public function getFormQuery($shortUID, $form) {
```
shaneodd, line 57: Wait, why is form here?

shaneodd (changes requested):

LEAF_Request_Portal/sources/Shortener.php
jampaul3: Because before Form was instantiated in the method. for various reasons I couldn't do that anymore so now I'm passing it in so it can be used. This was done in several places in this step. It's a start to getting instantiation of classes out of other classes. If a class is needed in a different class it really should be passed in not instantiated on the fly.

PR #1801, opened 2023-01-06
LEAF-3351 - Development env nginx settings

shaneodd:
adjust the nginx settings for dev so api calls work correctly.  

Testing area that showed this issue was deleting files within the file manager, other areas should be tested as well to make sure this change did not interfere.

### DIFF
--- a/docker/nginx/leaf_nginx.conf.template
+++ b/docker/nginx/leaf_nginx.conf.template
@@ -35,25 +35,6 @@ server {
         gzip on;
     }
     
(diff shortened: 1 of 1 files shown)


raphaelfretter (approved):
Wild that the config is just in the wrong spot.

AmandaSumner (approved):

jampaul3 (approved):

maxrichardjr (approved):

Pelentan (approved):

PR #1806, opened 2023-01-11
LEAF 3392 ELT access issue

aerinkayne:
This addresses an issue that can affect the 'hasAccess' value for users who are members of more than one ELT group, and prevent the action buttons from appearing during the request workflow.  This also resolves an error caused by a return value data type

### DIFF
--- a/LEAF_Request_Portal/FormWorkflow.php
+++ b/LEAF_Request_Portal/FormWorkflow.php
@@ -66,7 +66,7 @@ class FormWorkflow
      * @return array database result
      * @return null if no database result
      */
-    public function getCurrentSteps(): ?array
+    public function getCurrentSteps(): array|int|null
     {
         // check privileges
         require_once 'form.php';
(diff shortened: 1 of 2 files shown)


raphaelfretter (approved):

LEAF_Request_Portal/FormWorkflow.php
```diff
@@ -120,7 +120,7 @@ public function getCurrentSteps(): ?array
                         ':quadGroupIDs' => $quadGroupIDs
                     );
                     $strSQL = 'SELECT * FROM services
-                        WHERE groupID IN (:quadGroupIDs)
+                        WHERE find_in_set(groupID, :quadGroupIDs)
```
Pelentan, line 123: Not seeing the need for this change and it adds cycles to the computation.  "Where/In" just returns a boolean.  "find_in_set" has to compute the location and add 1.  And you're not using the return except as a boolean.

maxrichardjr (approved):

LEAF_Request_Portal/FormWorkflow.php
aerinkayne: If there is a more efficient way to do this, I can update it, but groupID IN was not consistently getting query results when the user was in more than one group.

shaneodd: Do you have an example of a trouble query you could share? (Probably not in this thread.)

AmandaSumner (approved):

shaneodd (approved):

PR #1810, opened 2023-01-13
LEAF 3405 remaining ELT group access issues

aerinkayne:
updates queries to the services table to correct behavior for members of multiple ELT groups

other: updates method return type for getLactAction to avoid a php error

### DIFF
--- a/LEAF_Request_Portal/FormWorkflow.php
+++ b/LEAF_Request_Portal/FormWorkflow.php
@@ -117,10 +117,9 @@ class FormWorkflow
                     $quadGroupIDs = $this->login->getQuadradGroupID();
                     $vars3 = array(
                         ':serviceID' => $res[$i]['serviceID'],
-                        ':quadGroupIDs' => $quadGroupIDs
                     );
                     $strSQL = 'SELECT * FROM services
-                        WHERE find_in_set(groupID, :quadGroupIDs)
+                        WHERE groupID IN ('.$quadGroupIDs.')
                         AND serviceID = :serviceID';
                     $res3 = $this->db->prepared_query($strSQL, $vars3);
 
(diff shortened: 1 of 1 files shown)


AmandaSumner (approved):

LEAF_Request_Portal/FormWorkflow.php
```diff
@@ -117,10 +117,9 @@ public function getCurrentSteps(): array|int|null
                     $quadGroupIDs = $this->login->getQuadradGroupID();
                     $vars3 = array(
                         ':serviceID' => $res[$i]['serviceID'],
-                        ':quadGroupIDs' => $quadGroupIDs
```
raphaelfretter, line 120: I feel like this should stay.

raphaelfretter (changes requested):

LEAF_Request_Portal/FormWorkflow.php
aerinkayne: I mentioned this to Michael Gao because I wanted to confirm this as well.  QuadIds is coming directly from us, and the IDs composing it from the db are enforced int types, so this should be a safe query.  There is also some additional complexity that would be associated with using parameterized variables inside WHERE ... IN

shaneodd: the original change had "WHERE groupID IN (:quadGroupIDs)" which did not work. It seemed like the data was getting escaped as a string since single ids would work but multiple did not. IE: WHERE groupID IN ('100,101') vs WHERE groupID IN (100,101). That is how it was rendering in my head.

maxrichardjr (approved):

jampaul3 (approved):

raphaelfretter (approved):

shaneodd (approved):

PR #1816, opened 2023-01-23
LEAF-3424 - Email Reminders not working on National sites

shaneodd:
Nationals do not normally use services. Add in flexibility to allow for services to exist or not.

Local replication, take a request that is currently on the step that an email reminder would go out. In the database remove the service (set it to 0 which is how nationals work). Then run the email reminder script from with in the scripts folder.

I would also suggest we get a demo national and a demo vax portal for us to specifically test.


raphaelfretter (approved):
Looks good.

maxrichardjr (dismissed):

aerinkayne (dismissed):

jampaul3 (dismissed):

Pelentan (dismissed):

AmandaSumner (approved):

PR #1830, opened 2023-02-01
LEAF 3346 - automated emails iteration after the first email

shaneodd:
I have adjusted the previous testing methodologies from just sending  the current information to augmenting the number of days to number of minutes. So for example initial email can go out in 1 day and every other notification will be 2 days. These numbers will then become 1 minute and 2 minutes when using the url below. 
https://localhost/LEAF_Request_Portal/scripts/automated_email.php?minutes=minutes

Some talking points for this would be the wording on the popup when choosing the additional notifications.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_workflow.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_workflow.tpl
@@ -252,7 +252,8 @@ function addEmailReminderDialog(stepID){
     dialog.setContent(output);
     
     dialog.setValidator('reminder_days', function() {
-        if ($('#edit_email_check').prop('checked') == true && (parseInt($('#reminder_days').val()) === NaN || parseInt($('#reminder_days').val()) < 1)) {
+        reminder_days = $('#reminder_days').val();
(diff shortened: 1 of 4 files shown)


maxrichardjr (approved):

raphaelfretter (approved):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -252,7 +252,8 @@ function addEmailReminderDialog(stepID){
     dialog.setContent(output);
     
     dialog.setValidator('reminder_days', function() {
-        if ($('#edit_email_check').prop('checked') == true && (parseInt($('#reminder_days').val()) === NaN || parseInt($('#reminder_days').val()) < 1)) {
+        reminder_days = $('#reminder_days').val();
+        if ($('#edit_email_check').prop('checked') == true && (parseInt(reminder_days) === NaN || parseInt(reminder_days) < 1) || reminder_days == '') {
```
aerinkayne, line 256: I think you want isNaN(val) .  val === NaN won't be true

maxrichardjr (approved):

jampaul3 (approved):

Pelentan (approved):

PR #1831, opened 2023-02-02
LEAF 2381 - email address can come in as null

shaneodd:
Email addresses can also come in as null, was only looking for strings since that is all I would see.

### DIFF
--- a/LEAF_Request_Portal/Email.php
+++ b/LEAF_Request_Portal/Email.php
@@ -109,9 +109,9 @@ class Email
 
     /**
      * Set email sender object variable
-     * @param $strAddress
+     * @param string|null $strAddress
      */
-    public function setSender(string $strAddress): void
+    public function setSender(string|null $strAddress): void
     {
         $this->emailSender = $strAddress;
     }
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/Email.php
```diff
@@ -109,9 +109,9 @@ public function clearRecipients(): void
 
     /**
      * Set email sender object variable
-     * @param $strAddress
+     * @param string|null $strAddress
      */
-    public function setSender(string $strAddress): void
+    public function setSender(string|null $strAddress): void
     {
```
jampaul3, line 115: Not that I want you to change it now, but this could be written as ?string $strAddress. the ? says that it can  be nullñ

jampaul3 (approved):

raphaelfretter (approved):

Pelentan (approved):

LEAF_Request_Portal/Email.php
shaneodd: Oh okay, did not know there was a shorthand way of writing that. I will look into that.

maxrichardjr (commented):
I agree - good thought

AmandaSumner (commented):
No errors or issues found during regression testing.

PR #1837, opened 2023-02-08
LEAF-2875 Mark old accounts as disabled

jrodcode:
This will make sure that if the user hasn't been updated after one week to be marked as disabled by adding the timestamp.

### DIFF
--- a/LEAF_Nexus/scripts/refreshOrgchartEmployees.php
+++ b/LEAF_Nexus/scripts/refreshOrgchartEmployees.php
@@ -48,8 +48,11 @@ if (strtolower($config->dbName) == strtolower(DIRECTORY_DB)) {
 
 /*
  *	Updates single employee information from national orgchart to local orgchart
-*/
-function updateUserInfo($userName, $empUID)
+ *
+ * @param string $userName
+ * @param int $empUID
+ */
+function updateUserInfo(string $userName, int $empUID)
 {
     global $db, $phonedb;
 
@@ -110,10 +113,7 @@ function updateUserInfo($userName, $empUID)
 
         // sets local employee_data table
         updateEmployeeData($res[0]['empUID'], $empUID);
-    }
 
-
-    if (count($res) > 0) {
         $currentTime = time();
         $oneWeekAdded = $res[0]['lastUpdated'] + 604800;
 
@@ -128,7 +128,7 @@ function updateUserInfo($userName, $empUID)
                     WHERE userName=:userName";
 
             $db->prepared_query($sql4, $vars);
-        }
+    }
 }
 
 function updateLocalOrgchartBatch()
--- a/LEAF_Nexus/scripts/refreshOrgchartEmployees.php
+++ b/LEAF_Nexus/scripts/refreshOrgchartEmployees.php
@@ -77,6 +77,10 @@ function updateUserInfo($userName, $empUID)
             SET deleted=:deleted
             WHERE userName=:userName";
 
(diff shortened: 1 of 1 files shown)


maxrichardjr (approved):

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
```diff
@@ -31,81 +31,114 @@
...
+        $oneWeekAdded = $res[0]['lastUpdated'] + 604800;
+
+        if ($currentTime > $oneWeekAdded) {
+            // user is disabled if userUpdated is lower than the current time
+            $vars = array(
+                ':deleted' => $currentTime,
+                ':userName' => $userName
+            );
+            $sql = "UPDATE employee
+                    SET userDisabled=:deleted
```
aerinkayne, line 123: Employee table field name is 'deleted'.  Also, if we don't want to process accounts that have a 'lastUpdated' value of 0, we could check that and only go forward with the rest if it isn't 0, rather than processing it and then updating it again.  Not sure if such accounts might already be deleted for other reasons - if so, we would not want to re-enable them.

shaneodd (approved):

jampaul3 (approved):

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
jrodcode: Thank you for finding the issue. Please review it again, thank you!

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
```diff
@@ -31,81 +31,104 @@
...
     }
-
 }
 
 /*
  *	Updates single employee information from national orgchart to local orgchart
 */
-function updateUserInfo($userName, $empUID){
-	global $db, $phonedb;
+function updateUserInfo($userName, $empUID)
```
Pelentan, line 52: need typehinting and return type

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
```diff
@@ -31,81 +31,104 @@
...
+
+        // sets local employee table
+        $db->prepared_query($sql2, $vars);
+
+        // sets local employee_data table
+        updateEmployeeData($res[0]['empUID'], $empUID);
+    }
+
+
+    if (count($res) > 0) {
```
Pelentan, line 116: why are there two identical if statements.  Can't everything be handled under one?

Pelentan (changes requested):

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
jrodcode: Will make the changes right now!

maxrichardjr (approved):

maxrichardjr (approved):

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
raphaelfretter: SET deleted=:deleted

raphaelfretter (changes requested):
See comments under  line 123

Pelentan (approved):

maxrichardjr (approved):

raphaelfretter (approved):

aerinkayne (approved):

maxrichardjr (approved):

AmandaSumner (approved):

PR #1849, opened 2023-02-21
LEAF 3398 file del endpoint update

aerinkayne:
Updates the registered deletion endpoint used in the file manager to take the file name as a parameter rather than through the url, which had caused some file names (eg test#file) to be undeletable through the UI.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_file_manager.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_file_manager.tpl
@@ -69,7 +69,10 @@ function deleteFile(file) {
             success: function() {
                 showFiles();
                 dialog_confirm.hide();
-            }
+            },
+            error: function(err) {
+                console.log('an error occurred during file deletion', err)
+            },
         });
     });
     dialog_confirm.show();
--- a/LEAF_Request_Portal/admin/templates/mod_file_manager.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_file_manager.tpl
@@ -64,8 +64,8 @@ function deleteFile(file) {
     dialog_confirm.setSaveHandler(function() {
         $.ajax({
             type: 'DELETE',
-            url: '../api/system/files/_'+ file + '?' +
-                $.param({'CSRFToken': CSRFToken}),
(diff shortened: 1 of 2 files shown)


maxrichardjr (approved):

jampaul3 (approved):

LEAF_Request_Portal/api/controllers/SystemController.php
```diff
@@ -152,8 +152,8 @@ public function delete($act)
         $this->index['DELETE']->register('system', function ($args) {
         });
 
-        $this->index['DELETE']->register('system/files/[text]', function ($args) use ($db, $login, $system) {
```
shaneodd, line 155: Would there be any custom code that this change would break?

LEAF_Request_Portal/admin/templates/mod_file_manager.tpl
```diff
@@ -64,8 +64,8 @@ function deleteFile(file) {
     dialog_confirm.setSaveHandler(function() {
         $.ajax({
             type: 'DELETE',
-            url: '../api/system/files/_'+ file + '?' +
-                $.param({'CSRFToken': CSRFToken}),
+            url: '../api/system/files/delete?' +
+                $.param({'CSRFToken': CSRFToken, 'file': file}),
             success: function() {
```
shaneodd, line 69: fail condition?

shaneodd (changes requested):

LEAF_Request_Portal/api/controllers/SystemController.php
aerinkayne: I don't think so.  It looked like that endpoint was used only by the file manager for files stored there.

LEAF_Request_Portal/admin/templates/mod_file_manager.tpl
aerinkayne: updated

maxrichardjr (approved):

shaneodd (approved):

Pelentan (approved):

PR #1861, opened 2023-03-01
LEAF-2390: Improve Group Form History

raphaelfretter:
vars/methods that aren't being used: 
FormatOptions::DISPLAY
dataActionLogger::findExternalValue()
any const TABLE in the formatter files
"key" and "display" columns in modify-indicator template


shaneodd (approved):

pete-nerantzinis (changes requested):
Need to merge in master and fix conflicts as the files have been renamed to capitals and not the lowercases, also there are some require_once's and instantiations that are not needed anymore since autoloader loads them for us now.

libs/logger/DataActionLogger.php
```diff
@@ -1,9 +1,8 @@
 <?php
 
 namespace Leaf;
+class DataActionLogger{
```
jampaul3, line 5: for classes the { is supposed to be on the next line by itself. According to PSR12 and for the rest of the files.

libs/logger/formatters/EmailTemplateFormatter.php
```diff
@@ -24,4 +24,4 @@ class EmailTemplateFormatter
     ];
 
     const TABLE = "email_templates";
-}
\ No newline at end of file
+}
```
jampaul3, line 27: PSR12 also wants a blank line at the end of a php only file.

jampaul3 (approved):

pete-nerantzinis (approved):

aerinkayne (approved):

shaneodd (approved):

PR #1877, opened 2023-03-10
LEAF-3573: Update National Standard Deploy

pete-nerantzinis:


aerinkayne (approved):

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -4,6 +4,8 @@
 require_once '../globals.php';
 require_once LIB_PATH . '/loaders/Leaf_autoloader.php';
 
+$db_config = new Portal\DbConfig;
```
jampaul3, line 7: This shouldn't be needed. $db_config is assigned just like this is the Leaf_autoloader.php file and because it is required here is available for use on the page. Doesn't hurt to have it here but not really necessary.

jampaul3 (approved):

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
pete-nerantzinis: I put this in because the echo under this tried calling $db_config->dbName and it wasnt working until I added this line.

PR #1884, opened 2023-03-15
LEAF 3411 add decode for excel 'from web' import

aerinkayne:
Adds decode for the Report Builder -> JSON (toolbar option) -> JSON option, use compat mode checked, endpoint to avoid encoded characters in excel 'from web' imports.

### DIFF
--- a/LEAF_Request_Portal/api/RESTfulResponse.php
+++ b/LEAF_Request_Portal/api/RESTfulResponse.php
@@ -91,22 +91,8 @@ abstract class RESTfulResponse
         $format = isset($_GET['format']) ? $_GET['format'] : '';
         switch ($format) {
             case 'json':
-            case 'jsoncompatmode':
             default:
                 header('Content-type: application/json');
-                //this is used specifically for excel imports 'from web' via Report Builder toolbar -> JSON -> compat mode
-                if($format === 'jsoncompatmode') {
-                    foreach ($out as $recordID => $record) {
-                        $out[$recordID]['title'] = html_entity_decode($record['title'], ENT_QUOTES);
-                        if (isset($record['s1'])) {
-                            foreach ($record['s1'] as $key => $entry) {
-                                if (preg_match('/^id\d+$/', $key) && is_string($entry)) {
-                                    $out[$recordID]['s1'][$key] = html_entity_decode($entry, ENT_QUOTES);
-                                }
-                            }
-                        }
-                    }
-                }
                 $jsonOut = json_encode($out);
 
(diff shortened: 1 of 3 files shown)


maxrichardjr (approved):

jampaul3 (approved):

pete-nerantzinis (approved):

LEAF_Request_Portal/api/RESTfulResponse.php
```diff
@@ -91,8 +91,22 @@ public function output($out = '')
         $format = isset($_GET['format']) ? $_GET['format'] : '';
         switch ($format) {
             case 'json':
+            case 'jsoncompatmode':
             default:
                 header('Content-type: application/json');
+                //this is used specifically for excel imports 'from web' via Report Builder toolbar -> JSON -> compat mode
+                if($format === 'jsoncompatmode') {
```
shaneodd, line 98: I know we had chatted on this, was there no way to determine if this was for the excel importer already? Just so people do not have to pull it in a different way or reimport the file. I am not sure how easy it is to just swap out a datasource in those tools.

shaneodd (changes requested):

LEAF_Request_Portal/api/RESTfulResponse.php
aerinkayne: I will touch in on this Monday.  I'd actually prefer to have it in the Shortener's getFormQuery for a few reasons.

pete-nerantzinis (approved):

LEAF_Request_Portal/sources/Shortener.php
```diff
@@ -54,6 +54,20 @@ private function decodeShortUID($in)
         return $res - $this->offset;
     }
 
+    public function decodeEntitiesForExcel($formQuery) {
+        foreach ($formQuery as $recordID => $record) {
+            $formQuery[$recordID]['title'] = html_entity_decode($record['title'], ENT_QUOTES);
+            if (isset($record['s1'])) {
```
Pelentan, line 60: Do we need to verify that this is an array here?  I.e. not a supercalifragulistic string?

Pelentan (changes requested):

LEAF_Request_Portal/sources/Shortener.php
aerinkayne: Added param and return value types.  I also updated the method name since other programs besides excel could use this.

maxrichardjr (approved):

PR #1885, opened 2023-03-15
Leaf-3348 script to move data to db

jampaul3:
This is a script to move the data from the Config files into the Db. Next step will be to refactor the Config class to pull the data from the db.

### DIFF
--- a/libs/dev_only_scripts/configToDb.php
+++ b/libs/dev_only_scripts/configToDb.php
@@ -36,33 +36,33 @@ function checkTemplate($folder) {
                     ON DUPLICATE KEY UPDATE `setting`=:setting';
 
             $vars = array(':setting' => 'heading',
-                    ':data' => $config->title);
+                    ':data' => $config['title']);
             $db->prepared_query($sql, $vars);
 
             $vars = array(':setting' => 'subHeading',
-                    ':data' => $config->city);
+                    ':data' => $config['city']);
             $db->prepared_query($sql, $vars);
 
             $data = json_encode($config->adPath, JSON_FORCE_OBJECT);
             error_log(print_r($data, true));
             $vars = array(':setting' => 'adPath',
-                    ':data' => json_encode($config->adPath, JSON_FORCE_OBJECT));
+                    ':data' => json_encode($config['adPath'], JSON_FORCE_OBJECT));
             $db->prepared_query($sql, $vars);
 
             $vars = array(':setting' => 'orgchartImportTags',
-                    ':data' => json_encode(Portal\Config::$orgchartImportTags, JSON_FORCE_OBJECT));
+                    ':data' => json_encode($config['orgchartImportTags'], JSON_FORCE_OBJECT));
             $db->prepared_query($sql, $vars);
 
             $vars = array(':setting' => 'descriptionID',
-                    ':data' => $config->descriptionID);
+                    ':data' => $config['descriptionID']);
             $db->prepared_query($sql, $vars);
 
             $vars = array(':setting' => 'emailCC',
-                    ':data' => json_encode(Portal\Config::$emailCC, JSON_FORCE_OBJECT));
+                    ':data' => json_encode($config['emailCC'], JSON_FORCE_OBJECT));
             $db->prepared_query($sql, $vars);
 
             $vars = array(':setting' => 'emailBCC',
-                    ':data' => json_encode(Portal\Config::$emailBCC, JSON_FORCE_OBJECT));
+                    ':data' => json_encode($config['emailBCC'], JSON_FORCE_OBJECT));
             $db->prepared_query($sql, $vars);
         } elseif (file_exists($folder . '/sources/Config.php')) {
             // we are in an orgchart
@@ -91,19 +91,19 @@ function checkTemplate($folder) {
                     ON DUPLICATE KEY UPDATE `setting`=:setting';
 
             $vars = array(':setting' => 'heading',
-                    ':data' => $config->title);
+                    ':data' => $config['title']);
             $db->prepared_query($sql, $vars);
 
             $vars = array(':setting' => 'subheading',
-                    ':data' => $config->city);
+                    ':data' => $config['city']);
             $db->prepared_query($sql, $vars);
 
             $vars = array(':setting' => 'adPath',
-                    ':data' => json_encode($config->adPath, JSON_FORCE_OBJECT));
+                    ':data' => json_encode($config['adPath'], JSON_FORCE_OBJECT));
             $db->prepared_query($sql, $vars);
 
             $vars = array(':setting' => 'ERM_Sites',
-                    ':data' => json_encode(Orgchart\Config::$ERM_Sites, JSON_FORCE_OBJECT));
+                    ':data' => json_encode($config['ERM_Sites'], JSON_FORCE_OBJECT));
             $db->prepared_query($sql, $vars);
         }
     } else {
--- a/libs/dev_only_scripts/configToDb.php
+++ b/libs/dev_only_scripts/configToDb.php
@@ -27,8 +27,9 @@ function checkTemplate($folder) {
 
             // instantiate a portal db
             $db =  new \Leaf\Db(getenv('DATABASE_HOST'), getenv('DATABASE_USERNAME'), getenv('DATABASE_PASSWORD'), $site_paths['portal_database']);
-
-            $config = readPortalConfig($portal);
+            $args[] = $portal;
+            $args[] = $site_paths['portal_database'];
+            $config = readPortalConfig($args);
 
             // populate db with config data
             $sql = 'INSERT INTO `settings` (`setting`, `data`)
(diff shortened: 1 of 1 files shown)


libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+
+require_once '/var/www/html/libs/php-commons/Db.php';
+
+$dir = '/var/www/html';
+
+checkTemplate($dir);
+
+function checkTemplate($folder) {
+    //if (is_dir($folder . '/.svn')) {
+    if (is_dir($folder . '/sources')) {
```
pete-nerantzinis, line 11: This should be .svn otherwise this will take a very long time since it will go even deeper.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+            require_once $folder . '/sources/Config.php';
+            $config = new Portal\Config();
+
+            // populate db with config data
+            $sql = 'INSERT INTO `settings` (`setting`, `data`)
+                    VALUES (:setting, :data)
+                    ON DUPLICATE KEY UPDATE `setting`=:setting';
+
+            $vars = array(':setting' => 'heading',
+                    ':data' => $config->title);
```
pete-nerantzinis, line 40: We don't want to do this, since alot of users have customized names already it will overwrite the names they chose.
Better off if we left the orgcharts that are already blank just let the users edit them. Either make a check for empty data to then input the data or just leave it.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+            $db->prepared_query($sql, $vars);
+
+            $vars = array(':setting' => 'subHeading',
+                    ':data' => $config->city);
```
pete-nerantzinis, line 44: Same here, we don't want to overwrite the data users have set already since this is an empty variable on the configs. Either make a check for empty data to then input the data or just leave it.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+            $db->prepared_query($sql, $vars);
+
+            $data = json_encode($config->adPath, JSON_FORCE_OBJECT);
+            error_log(print_r($data, true));
+            $vars = array(':setting' => 'adPath',
```
pete-nerantzinis, line 49: Is this variable even being used in our codebase? Do we need to add it to the DB? Also believe its an empty array.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+                    ':data' => json_encode($config->adPath, JSON_FORCE_OBJECT));
+            $db->prepared_query($sql, $vars);
+
+            $vars = array(':setting' => 'orgchartImportTags',
+                    ':data' => json_encode(Portal\Config::$orgchartImportTags, JSON_FORCE_OBJECT));
+            $db->prepared_query($sql, $vars);
+
+            $vars = array(':setting' => 'descriptionID',
```
pete-nerantzinis, line 57: Again is this variable used anywhere in the code its the same for every portal/orgchart I believe.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+                    ':data' => $config->descriptionID);
+            $db->prepared_query($sql, $vars);
+
+            $vars = array(':setting' => 'emailPrefix',
```
pete-nerantzinis, line 61: Check if this variable gets used still in the code, but this is already being set in settings so we dont need to overwrite it but we need to make sure theres a default if its being used in the code. The setting for this one is requestLabel in settings table.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+                    ':data' => Portal\Config::$emailPrefix);
+            $db->prepared_query($sql, $vars);
+
+            $vars = array(':setting' => 'emailCC',
```
pete-nerantzinis, line 65: For emailCC and emalBCC do these even get used in our codebase from this config I don't believe this is necessary since theres not a setting for these on the Site Settings page as well. And they are just empty array's on all the sites configs.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+
+            $vars = array(':setting' => 'heading',
+                    ':data' => $config->title);
+            $db->prepared_query($sql, $vars);
+
+            $vars = array(':setting' => 'subHeading',
+                    ':data' => $config->city);
+            $db->prepared_query($sql, $vars);
+
+            $vars = array(':setting' => 'adPath',
```
pete-nerantzinis, line 107: Again like above check if this is even used or necessary to track anymore.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+                    ':data' => json_encode($config->adPath, JSON_FORCE_OBJECT));
+            $db->prepared_query($sql, $vars);
+
+            $vars = array(':setting' => 'ERM_Sites',
```
pete-nerantzinis, line 111: Check if this is even used or necessary to track anymore.

libs/dev_only_scripts/configToDb.php
```diff
@@ -0,0 +1,144 @@
...
+            // instantiate a Config.php
+            require_once $folder . '/sources/Config.php';
+            $config = new Orgchart\Config();
+
+            // populate db with config data
+            $sql = 'INSERT INTO `settings` (`setting`, `data`)
+                    VALUES (:setting, :data)
+                    ON DUPLICATE KEY UPDATE `setting`=:setting';
+
+            $vars = array(':setting' => 'heading',
```
pete-nerantzinis, line 99: Like the above for the heading/subheading lets not change these unless they are blank.

pete-nerantzinis (changes requested):
Let me know when you have done these changes, we should be checking if theres already a entry and if there is not to overwrite it because any database setting that has data already was set by the user.

libs/dev_only_scripts/configToDb.php
jampaul3: if you look at the on duplicate key it's not over writing the data, it's just resetting the setting so nothing should change. This was the quickest way I found to do it.

libs/dev_only_scripts/configToDb.php
jampaul3: same as above reply.

maxrichardjr (approved):

maxrichardjr (approved):

libs/dev_only_scripts/configToDb.php
jampaul3: Best as I can tell adPath is used in the import.php which is on every orgchart. The only place I know it is used beyond that is on launchpad when a new site is created. I also did do some digging awhile ago and there are some sites that have a populated array. Most however, I agree are empty array's

libs/dev_only_scripts/configToDb.php
jampaul3: This is used once in Email.php in the method setSubject, this is how it's being used:

$prefix = isset(Config::$emailPrefix) ? Config::$emailPrefix : 'Resources: ';

libs/dev_only_scripts/configToDb.php
jampaul3: Here is how it's being used currently.

if (isset(Config::$emailCC) && count(Config::$emailCC) > 0) {
            foreach (Config::$emailCC as $recipient) {
                $this->addCcBcc($recipient);
            }
        }

libs/dev_only_scripts/configToDb.php
pete-nerantzinis: Oh you are right I missed that haha my bad.

libs/dev_only_scripts/configToDb.php
pete-nerantzinis: Alright so lets just import it for now then.

libs/dev_only_scripts/configToDb.php
pete-nerantzinis: Guess this was meant to be added to settings on a greater level than email templates at some point lets just add it for now. They most likely are all empty arrays.

libs/dev_only_scripts/configToDb.php
pete-nerantzinis: Alright if its being used then lets import it.

maxrichardjr (approved):

pete-nerantzinis (changes requested):
Script looks good but we need to make sure we refactor the codebase as well for where things are pulling from DB now

shaneodd (approved):

Pelentan (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

pete-nerantzinis (approved):

jampaul3:
This script was run several weeks ago and is no longer valid.

PR #1896, opened 2023-03-24
Leaf 3176 - refactor RESTfulResponse

jampaul3:
So the refactor of RESTfulResponse in complete. This refactor added type hinting and removed the echo's that were within the file.

Some of the echo's were moved to the individual controller classes, which isn't any better than what was there before, but I felt it was outside the scope of this ticket to go further than this. To fix those later will have to be another ticket. 

I believe the type hinting is all good. I was able to test locally with no php errors or console errors. But that is going to be the thing to watch for in testing.


maxrichardjr (approved):

maxrichardjr (approved):

raphaelfretter (commented):
Is there a reason to echo the calls in the index files? Does it go to a log of some kind?

jampaul3:
@raphaelfretter functionality should be exactly as it was. I didn't want to make any more changes than necessary. So the echo's were in the RESTfulResponse, but in order to clean the class up I moved them out the the index. Classes shouldn't have echo's in them if the are processing data which these classes are. The echo will be sent back as part of the response to the api call, as it did previously. It's just now being sent from the index rather than the class.

A good question here would be, is this the proper way to do it? That I can't answer right now, but the changes I made here was to clean up the RESTfulResponse while keeping the overall functionality the same.

raphaelfretter:
> @raphaelfretter functionality should be exactly as it was. I didn't want to make any more changes than necessary. So the echo's were in the RESTfulResponse, but in order to clean the class up I moved them out the the index. Classes shouldn't have echo's in them if the are processing data which these classes are. The echo will be sent back as part of the response to the api call, as it did previously. It's just now being sent from the index rather than the class.
> 
> A good question here would be, is this the proper way to do it? That I can't answer right now, but the changes I made here was to clean up the RESTfulResponse while keeping the overall functionality the same.

That makes sense. I don't know of an alternative off the top of my head.

raphaelfretter (approved):

shaneodd (approved):

aerinkayne (approved):

LEAF_Request_Portal/api/controllers/FormEditorController.php
```diff
@@ -63,186 +63,203 @@ public function get($act)
...
-        $login = $this->login;
 
-        $this->verifyAdminReferrer();
+        $verified = $this->verifyAdminReferrer();
 
-        $this->index['DELETE'] = new ControllerMap();
-        $this->index['DELETE']->register('formEditor', function ($args) {
-        });
+        if ($verified) {
+            echo $verified;
```
Pelentan, line 252: Is this supposed to be commented out?

LEAF_Request_Portal/api/controllers/FormStackController.php
```diff
@@ -44,35 +44,42 @@ public function get($act)
...
         $formStack = $this->formStack;
 
-        $this->verifyAdminReferrer();
+        $verified = $this->verifyAdminReferrer();
 
-        $this->index['DELETE'] = new ControllerMap();
-        $this->index['DELETE']->register('workflow', function ($args) {
-        });
+        if ($verified) {
+            echo $verified;
```
Pelentan, line 72: And this?  Might want to do a search on "echo"s

LEAF_Request_Portal/api/controllers/GroupController.php
```diff
@@ -55,50 +55,60 @@ public function get($act)
 
     public function post($act)
     {
-        $this->verifyAdminReferrer();
-        $group = $this->group;
+        $verified = $this->verifyAdminReferrer();
 
-        $this->index['POST'] = new ControllerMap();
+        if ($verified) {
+            echo $verified;
```
Pelentan, line 61: echo

LEAF_Request_Portal/api/index.php
```diff
@@ -48,40 +48,40 @@
 
 $controllerMap->register('classicphonebook', function () use ($p_db, $login, $action) {
     $controller = new Portal\ClassicPhonebookController($p_db, $login);
-    $controller->handler($action);
+    echo $controller->handler($action);
```
Pelentan, line 51: Are these echos supposed to be here?

Pelentan (changes requested):

AmandaSumner (approved):

PR #1900, opened 2023-03-28
LEAF-3630: Simplify instructions

mgaoVA:
This aims to simplify installation for developers.

1. Are the lines removed necessary? It seems to work without them.
2. Is deployment impacted if the Config.php files are stored in the repository? It would be easier if people didn't need to manually create this.

### DIFF
--- a/docs/InstallationConfiguration.md
+++ b/docs/InstallationConfiguration.md
@@ -15,35 +15,9 @@ Clone this project into a directory on your computer (example: C:\Desktop\Projec
 
 Several files need to be created/updated for LEAF to operate in your environment.
 
-In the sections below `$dbUser` and `$dbPass` are the same values used in the mysql Dockerfile and setup script.
-
 ## LEAF_Nexus
 
-Copy `globals.php` from the configuration files provided they will look like:
-Note there is a line here that can help you emulate a national db if you are pulling in data from a national however this file IS part of the repo now.
-
-```php
-<?php
-if (!defined('PRODUCT_NAME')) define('PRODUCT_NAME', getenv('PRODUCT_NAME'));
-if (!defined('VERSION_NUMBER')) define('VERSION_NUMBER', getenv('NEXUS_VERSION_NUMBER'));
-
-if (!defined('DATABASE_DB_ADMIN')) define('DATABASE_DB_ADMIN', getenv('DATABASE_DB_ADMIN'));
-
-if (!defined('DIRECTORY_HOST')) define('DIRECTORY_HOST', getenv('DATABASE_HOST'));
-// if you need to emulate a national you can create drop in a table name here instead of the getenv.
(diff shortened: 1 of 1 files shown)


maxrichardjr (approved):

shaneodd:
> This aims to simplify installation for developers.
> 
> 1. Are the lines removed necessary? It seems to work without them.
> 2. Is deployment impacted if the Config.php files are stored in the repository? It would be easier if people didn't need to manually create this.

I would suggest we leave the notes for generating configs for each side, globals I believe are now part of the repo. For the configs I think that will be the plan eventually, Jamie is working on getting everything that would be in the configs into the db since there was a few more things to adjust. I think he can maybe explain that part a bit better than I.

docs/InstallationConfiguration.md
```diff
@@ -77,47 +51,10 @@ class Config
...
-if (!defined('LEAF_NEXUS_URL')) define('LEAF_NEXUS_URL', getenv('APP_URL_NEXUS'));
-if (!defined('HTTP_HOST')) define('HTTP_HOST', getenv('APP_HTTP_HOST'));
-if (!defined('AUTH_URL')) define('AUTH_URL', getenv('APP_URL_AUTH'));
-if (!defined('AUTH_TYPE')) define('AUTH_TYPE', getenv('APP_AUTH_TYPE'));
-if (!defined('CIPHER_KEY')) define('CIPHER_KEY', getenv('APP_CIPHER_KEY'));
-if (!defined('LIB_PATH')) define('LIB_PATH', getenv('APP_LIB_PATH'));
-
-```
-
-Create the following files `sources/DbConfig.php` and `sources/Config.php` and place the following code into them
```
shaneodd, line 103: I would keep this block for now until we move over to using just the db for configs.

shaneodd (changes requested):

aerinkayne (commented):
There are currently two files portal side.  If setup will change again in the near future though, I agree it might be better to wait to update.

docs/InstallationConfiguration.md
jampaul3: I'm in agreement with Shane. Changes to the Config files in both the LEAF_Nexus and LEAF_Request_Portal are coming in the next sprint I believe. When those changes come all the data will be pulled from the db and these classes will be updated like all other classes and won't need to be protected by gitignore. The DB_Config is going to go away completely and will no longer be used.

jampaul3:
I believe the changes proposed here are going to change again in the next sprint so I feel we ought to wait until at least then to make these changes. I can make them in my branch that deals with these.

docs/InstallationConfiguration.md
shaneodd: This change will get superseded by this change https://github.com/department-of-veterans-affairs/LEAF/pull/1939

raphaelfretter (approved):

PR #1901, opened 2023-03-28
LEAF 2839 ifthen support for option loading and crosswalk

aerinkayne:
This update adds support for loading single and multiselect dropdown options from files via the conditions-editor.


maxrichardjr (approved):
For form.js, In PHP, is 'default:' the same as 'case: else'? Are there other responses that may be returned?

            switch(xhttpInds.status) {
              case 200:
                resolve(xhttpInds.responseText);
                break;
              case 404:
                let content = `The file for indicator ${iID} was not found at files/${fileName}.`
                content += `\nCheck the entered file name in setup and in the LEAF file manager.`
                reject(new Error(content));
                break;
              default:
                reject(new Error(xhttpInds.status));
                break;

maxrichardjr (approved):

shaneodd (approved):
Array.some - cool

Pelentan (approved):

jampaul3 (approved):

AmandaSumner (approved):

PR #1933, opened 2023-04-06
LEAF-2034 - Add File Tracking

jrodcode:
This feature enables users to make changes to different sections of the admin panel and track multiple versions of edited files. The system allows for seamless merging of edited files in case of content changes. This streamlines the editing process, prevents errors, and ensures that all changes are recorded and stored for future reference.


Pelentan (changes requested):
No.

PR #1961, opened 2023-05-05
LEAF-3714 - File History Limit

jrodcode:
I replaced the rand() function with the uniqid() function due to users hitting a file cap, which required a change in function.


jampaul3 (approved):

raphaelfretter (approved):

aerinkayne (approved):

shaneodd (approved):

maxrichardjr (approved):

mgaoVA:
@jrodcode Would it be easier to use time() ? No loop would be necessary.

jrodcode:
> @jrodcode Would it be easier to use time() ? No loop would be necessary.

maxrichardjr (approved):

jampaul3 (approved):

PR #1971, opened 2023-05-15
Leaf 2789 Tracking emails sent

jampaul3:


maxrichardjr (approved):

maxrichardjr (approved):

LEAF_Request_Portal/js/pages/mass_action.js
raphaelfretter, line 1: May be worth updating tabbing here to 4 spaces. 2 is harder to read.

raphaelfretter: Also am not certain if we wanted to stick with 4 spaces for the future or try to move to 2 spaces

LEAF_Request_Portal/templates/view_status.tpl
```diff
@@ -28,7 +28,11 @@ Title of request: <a href="?a=printview&amp;recordID=<!--{$recordID|strip_tags|e
     <td>
         <span><b><!--{$indicator.description|sanitize}--></b> by <!--{$indicator.userName|sanitize}-->
         <!--{if $indicator.comment != ''}-->
-        <br />Comment: <!--{$indicator.comment|sanitize}--></span>
+            <!--{if $indicator.description != 'Email Sent: '}-->
```
raphaelfretter, line 31: I don't recall if there is an option for piping to lower case for this comparison in smarty.

raphaelfretter (changes requested):

LEAF_Request_Portal/sources/Email.php
```diff
@@ -324,9 +346,22 @@ public function sendMail(): bool
             exec("php {$currDir}/../mailer/mailer.php {$emailQueueName} > /dev/null &");
         }
 
+        $this->logEmailSent($recordID);
```
pete-nerantzinis, line 349: What happens here if no $recordID is being passed through. It seems you are putting a requirement on sendMail to have a recordID now. There are alot of custom scripts out there that use sendMail and don't pass through a recordID. Make sure this has backwards compatibility as well as all the other changes elsewhere in this ticket especially when you change a parameter set on a function and don't give it a default value.

LEAF_Request_Portal/sources/Form.php
```diff
@@ -3356,6 +3360,11 @@ public function query($inQuery)
                         FROM notes
                         WHERE recordID IN (' . $recordIDs . ')
                         AND deleted IS NULL
+                        UNION
```
pete-nerantzinis, line 3363: I haven't worked with a lot of UNION's before but just want to make sure this will be more efficient on resources vs separate querying.

pete-nerantzinis (changes requested):

LEAF_Request_Portal/js/pages/mass_action.js
jampaul3: this is fixed, prettier changed it on me.

LEAF_Request_Portal/templates/view_status.tpl
jampaul3: added.

LEAF_Request_Portal/sources/Form.php
jampaul3: From my research doing a UNION is the same as calling 3 separate queries, essentially it is 3 queries run simultaneously and put together in one result.

maxrichardjr (approved):

maxrichardjr (approved):

aerinkayne (approved):

LEAF_Request_Portal/sources/Email.php
jampaul3: I added a default of null and allowed the parameter to accept null. The email won't be tracked if a recordID is not supplied so I guess we'll deal with HD tickets for this as they come in.

pete-nerantzinis (approved):

Pelentan (approved):

jrodcode (approved):

maxrichardjr (approved):

maxrichardjr (approved):

raphaelfretter (approved):

maxrichardjr (approved):

shaneodd (approved):

jrodcode (approved):

maxrichardjr (approved):

maxrichardjr (approved):

AmandaSumner (approved):

maxrichardjr (approved):

PR #1990, opened 2023-06-05
Leaf-3546  keeping the old endpoints.

jampaul3:
This is a redo to keep the old endpoints in place. There is more work to be done but wanted to at least get this out. The work to be done is some sort of notification for the old endpoints.


jampaul3:
After talking to Michael Gao I don't need this. I just fixed the old PR

PR #1998, opened 2023-06-09
LEAF 3773 portal js cleanup 1 (formQuery)

aerinkayne:
Cleanup for the formQuery js file.  This file currently exists both in portal/js and libs/js.
There should be no behavior changes.


maxrichardjr (approved):

LEAF_Request_Portal/js/formQuery.js
```diff
@@ -227,72 +183,57 @@ var LeafFormQuery = function () {
...
-    let urlParamJSONP = "";
-    if (useJSONP) {
-      dataType = "jsonp";
-      urlParamJSONP = "&format=jsonp";
-    }
 
+    const queryUrl = el.innerText;
+    const dataType = useJSONP ? "jsonp" : "json";
+    const urlParamJSONP = useJSONP ? "&format=jsonp" : "";
     return $.ajax({
```
Pelentan, line 229: Ok.  Two problems here.  I know they are inherited, but we need to fix them.  One, it's returning an ajax call.  Big no-no.  It's one thing to return a promise, as hopefully what's calling it knows what to do with a promise.  Quite another to return something that can return literally anything.  I saw the below as well.  Additionally, not seeing a fail-catch.

LEAF_Request_Portal/js/formQuery.js
```diff
@@ -316,71 +257,44 @@ var LeafFormQuery = function () {
...
-          extraParams,
-        dataType: "jsonp",
-        success: successCallback,
-      });
-    }
+    
+    const queryUrl = el.innerText;
+    const dataType = useJSONP ? "jsonp" : "json";
+    const urlParamJSONP = useJSONP ? "&format=jsonp" : "";
+    return $.ajax({
```
Pelentan, line 270: same as above.

libs/js/LEAF/formQuery.js
```diff
@@ -26,113 +27,70 @@ var LeafFormQuery = function () {
...
    * @memberOf LeafFormQuery
    */
-  function importQuery(input) {
-    for (let i in input.terms) {
-      switch (Object.keys(input.terms[i]).length) {
+  function importQuery({ terms = [], joins = [], getData: getIndData = [] } = {}) {
+    terms.forEach(t => {
+      const {id, indicatorID, operator, match, gate} = t;
+      const numTermProperties = (Object.keys(t).length);
+      switch (numTermProperties) {
```
Pelentan, line 63: toLower()?   Kidding!  Just kidding!

libs/js/LEAF/formQuery.js
```diff
@@ -227,72 +183,57 @@ var LeafFormQuery = function () {
...
-    let urlParamJSONP = "";
-    if (useJSONP) {
-      dataType = "jsonp";
-      urlParamJSONP = "&format=jsonp";
-    }
 
+    const queryUrl = el.innerText;
+    const dataType = useJSONP ? "jsonp" : "json";
+    const urlParamJSONP = useJSONP ? "&format=jsonp" : "";
     return $.ajax({
```
Pelentan, line 229: same as above

libs/js/LEAF/formQuery.js
```diff
@@ -316,71 +257,44 @@ var LeafFormQuery = function () {
...
-          extraParams,
-        dataType: "jsonp",
-        success: successCallback,
-      });
-    }
+    
+    const queryUrl = el.innerText;
+    const dataType = useJSONP ? "jsonp" : "json";
+    const urlParamJSONP = useJSONP ? "&format=jsonp" : "";
+    return $.ajax({
```
Pelentan, line 270: again

Pelentan (changes requested):

pete-nerantzinis (approved):

shaneodd (approved):

LEAF_Request_Portal/js/formQuery.js
aerinkayne: I've tried this in various ways and there doesn't seem to be an easy way to pull these out.  The query execute method is exposed for use in custom reports.  Depending on how the query is set up, the return value is expected to be a jq ajax call.  The other path execute could go is to a call to getBulkData, which returns recursive calls to itself (and its ajax call also behaves differently on success).

Pelentan (approved):

AmandaSumner (approved):

jrodcode (approved):

PR #1999, opened 2023-06-09
Leaf 3774 - refresh orgchart employee api

jampaul3:
This is complete, both the onclick in the orgchart and the batch processing. I'm not sure though if the batch will work properly, with the api call it's expecting a csrf token, so if a session isn't created when the script is run then we'll have to come up with something else to make it work.


maxrichardjr (approved):

LEAF_Nexus/sources/Employee.php
```diff
@@ -75,6 +77,460 @@ public static function cleanWildcards($input)
...
+    }
+
+    /**
+     * @param array $local_employees_data_array
+     *
+     * @return void
+     *
+     * Created at: 6/9/2023, 2:29:12 PM (America/New_York)
+     */
+    private function batchEmployeeDataUpdate(array $local_employees_data_array): void
```
Pelentan, line 267: Same here.

LEAF_Nexus/sources/Employee.php
```diff
@@ -75,6 +77,460 @@ public static function cleanWildcards($input)
...
+    }
+
+    /**
+     * @param array $user
+     *
+     * @return bool
+     *
+     * Created at: 6/9/2023, 2:30:52 PM (America/New_York)
+     */
+    private function isActiveNationally(array $user): bool
```
Pelentan, line 392: 'Course if all you're returning is a boolean, not sure there is a reason to return anything else.

LEAF_Nexus/sources/Employee.php
```diff
@@ -75,6 +77,460 @@ public static function cleanWildcards($input)
...
+    /**
+     * @param string $user_name
+     * @param array $national_user
+     * @param Db $db
+     *
+     * @return void
+     *
+     * Created at: 6/9/2023, 2:31:54 PM (America/New_York)
+     */
+    private function updateEmployeeByUserName(string $user_name, array $national_user, Db $db): void
```
Pelentan, line 459: But here as well.  I don't think something like this should be "fire and forget".  In this case, it's common for the return to include the results of a query that demonstrates all the data was correctly updated.

LEAF_Nexus/sources/Employee.php
```diff
@@ -75,6 +77,460 @@ public static function cleanWildcards($input)
...
+    /**
+     * @param int $empUID
+     * @param array $national_data
+     * @param Db $db
+     *
+     * @return void
+     *
+     * Created at: 6/9/2023, 2:32:12 PM (America/New_York)
+     */
+    private function updateEmployeeDataByEmpUID(int $empUID, array $national_data, Db $db): void
```
Pelentan, line 494: Same as above.

LEAF_Nexus/sources/Employee.php
```diff
@@ -75,6 +77,460 @@ public static function cleanWildcards($input)
...
+
+    /**
+     * @param string $user_name
+     * @param Db $db
+     *
+     * @return void
+     *
+     * Created at: 6/9/2023, 2:31:38 PM (America/New_York)
+     */
+    private function disableEmployee(string $user_name, Db $db): void
```
Pelentan, line 522: Ditto

raphaelfretter (approved):

LEAF_Nexus/sources/Employee.php
jampaul3: I am returning the data that was updated, if you look at where this method is called you will see the next line gets the data to be returned. I double checked to be sure, an update sql statement doesn't return anything so unless I query the date here instead of after the call like I am now there is nothing really to send back.

LEAF_Nexus/sources/Employee.php
jampaul3: Same as above.

LEAF_Nexus/api/controllers/EmployeeController.php
```diff
@@ -58,12 +61,46 @@ public function get($act)
...
         $employee = $this->employee;
+        $national_db = $this->national_db;
 
         $this->index['POST'] = new ControllerMap();
         $this->index['POST']->register('employee', function ($args) {
             return print_r($args, true) . print_r($_GET, true);
         });
 
+
+        $this->index['POST']->register('employee/refresh/[text]/[digit]', function ($args) use ($employee, $national_db) {
```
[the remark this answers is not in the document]
jampaul3: Not sure why I did it this way unless I was testing something. It's been fixed now.

LEAF_Nexus/sources/Employee.php
```diff
@@ -75,6 +77,460 @@ public static function cleanWildcards($input)
...
+    /**
+     * @param array $local_employees_array
+     *
+     * @return void
+     *
+     * Created at: 6/9/2023, 2:29:05 PM (America/New_York)
+     */
+    private function batchEmployeeUpdate(array $local_employees_array): void
+    {
+        $vars = array(
```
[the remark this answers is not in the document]
jampaul3: I added the check here, I also added it where it's being called. Figure might just as well catch it before calling the method. I did add the json, api returns here. Although I'm not sure that it'll do much good as this is a batch call meant to be done from a script that runs behind the scenes but it's there, if for no other reason than to debug.

LEAF_Nexus/sources/Employee.php
jampaul3: Same here

LEAF_Nexus/sources/Employee.php
jampaul3: added the json, api return data.

LEAF_Nexus/sources/Employee.php
Pelentan: Ok.  Then just a "success" json return

shaneodd (approved):

Pelentan (approved):

jrodcode (approved):

maxrichardjr (approved):

maxrichardjr (approved):

PR #2009, opened 2023-06-20
LEAF-3472 - Handle Large Queries 

shaneodd:
**Summary**
This is to handle large queries from the report section of the website. This is mainly going to be for the people who use the api to pull in data into power bi and excel. The process will be setup to recheck to see if there are any new entries every 5 minutes. This should allow more than enough time for processes to clear run their course before calling more. 

**Opportunities for further improvement**
We will need to maybe adjust the frequency or number of calls at a time.

**Potential Impacts**
This could have impact anywhere we are loading in data from the api that the report builder uses. Custom code may have issues if they are not slotting in the data in 1000 row chunks. They will get a 202 error code and some text to let them know their data will be processed. 

**Testing**
Report builder, the entry page of the portal and mass actions are areas that we are currently using the updated api. Portals with custom code should be tested as well to make sure they load data properly

One note, this is a table that will need to go to the launch pad, since we do not have an official way of updating that at this time here is the database create.
`CREATE TABLE `site_large_query` (
	`id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT,
	`site_id` INT(10) UNSIGNED NOT NULL,
	`created_at` INT(10) UNSIGNED NULL DEFAULT NULL,
	`currently_running` INT(10) UNSIGNED NULL DEFAULT '0',
	PRIMARY KEY (`id`) USING BTREE,
	UNIQUE INDEX `site_id` (`site_id`) USING BTREE,
	INDEX `created_at` (`created_at`) USING BTREE,
	INDEX `currently_running` (`currently_running`) USING BTREE
)
COLLATE='latin1_swedish_ci'
ENGINE=InnoDB
AUTO_INCREMENT=26
;
`

**Prerequisites to Prod Deployment**
- [ ] Need a script to search for "api/form/query/?q={" within custom templates, Identify sites where they occur, reach out to affected admins informing them of the change (TBD: work with comms on message)
- [ ] Need the ability for LEAF dev team members to easily visualize the status of the query queue


maxrichardjr (approved):

LEAF_Request_Portal/db_config.php
raphaelfretter, line 1: Doesn't this need to be DbConfig.php?

raphaelfretter (changes requested):

LEAF_Request_Portal/sources/Form.php
```diff
@@ -2091,21 +2064,18 @@ public function checkReadAccess($records)
      */
     public function isMasked($indicatorID, $recordID = null)
```
Pelentan, line 2065: Same here

LEAF_Request_Portal/sources/Form.php
```diff
@@ -2120,32 +2090,28 @@ public function isMasked($indicatorID, $recordID = null)
      */
     public function isNeedToKnow($recordID = null)
```
Pelentan, line 2091: And here.  Just take a look at these methods to see if they can be cleaned up a bit without adding too much extra work on things.

LEAF_Request_Portal/db_config.php
shaneodd: You are correct, the file keeps popping back in. I think I have a branch that has it saved.

LEAF_Request_Portal/sources/Form.php
```diff
@@ -2660,64 +2594,410 @@ public function setInitiator($recordID, $userID)
...
+
+        return $retData;
+    }
+
+    /**
+     * @param string $inQuery - the query used in the processingF
+     * @param bool $processQuery - allows the processing of the query via script
+     * @return mixed
+     */
+    public function query(string $inQuery, bool $processQuery = false)
```
Pelentan, line 2968: Don't forget return types.  ^^^^&vvv

LEAF_Request_Portal/sources/Form.php
```diff
@@ -2921,179 +3192,152 @@ public function query($inQuery)
...
-                            {
+                            if (
+                                $joinSearchAllData
+                                || $joinSearchOrgchartEmployeeData
+                            ) {
                                 $dataTerm = 'lj_data.data';
                             }
 
                             $dataMatch = ":data{$count}";
                             switch ($tResTypeHint[0]['format']) {
```
Pelentan, line 3311: Can you toss a toLower() on this one.  The above ones would probably take too long to fix and would require a lot of additional testing.  But it looks like this one is already using all lowercase for now.  Might as well lock it in.

LEAF_Request_Portal/sources/Form.php
```diff
@@ -3224,307 +3457,189 @@ public function query($inQuery)
...
-                $res2 = $this->db->prepared_query($recordResolutionBySQL, array());
 
-                foreach ($res2 as $item) {
-                    $user = $dir->lookupLogin($item['resolvedBy'], true);
-                    $nameResolved = isset($user[0]) ? "{$user[0]['Lname']}, {$user[0]['Fname']} " : $item['resolvedBy'];
-                    $data[$item['recordID']]['recordResolutionBy']['resolvedBy'] = $nameResolved;
+                // copy each file off to other entries that match
+                if (!empty($processQueryCheckRes)) {
+                    foreach ($processQueryCheckRes as $processQueryCheckRow) {
+                        copy($currentFileName, $directory . $processQueryCheckRow['id'] . '_' . $processQueryCheckRow['userID'] . '.json');
```
Pelentan, line 3556: Yikes!  Can this be exploited?  None of those are user-entered values are they?

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -797,15 +797,58 @@ function openShareDialog() {
...
 
+function setExportFormat() {
+        if($('#shortenLink').css('display') === 'none') {
+            $('#exportFormat').html('?');
+        }
+        else {
+            $('#exportFormat').html('&');
+        }
+
+        switch($('#format').val()) {
```
Pelentan, line 808: toLower(), please.

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -797,15 +797,58 @@ function openShareDialog() {
...
+
+let maxLimit = 10000;
+
+/**
+ * @param currentQuery
+ * @return currentQuery
+ */
+function setQueryLimit(currentQuery){
+    
+    switch($('#format').val()) {
```
Pelentan, line 829: Same here please.

docker/mysql/db/db_upgrade/portal/Update_RMC_DB_2023052500-2023060500.sql
```diff
@@ -0,0 +1,21 @@
+START TRANSACTION;
+
+CREATE TABLE `process_query` (
+  `id` INT(11) NOT NULL AUTO_INCREMENT,
+  `userID` VARCHAR(50) DEFAULT NULL,
+  `url` TEXT DEFAULT NULL,
+  `lastProcess` INT(10) NOT NULL,
+  PRIMARY KEY (`id`)
```
Pelentan, line 8: Unsure how big this table is.  But it's likely we want an index on more than the id.

Pelentan (changes requested):

LEAF_Request_Portal/sources/Form.php
shaneodd: Unless there is a path I am missing I was not seeing a way to exploit this. This was to reduce the number of times the script had to run. I can remove it if you feel this could have an exploit hidden away.

docker/mysql/db/db_upgrade/portal/Update_RMC_DB_2023052500-2023060500.sql
shaneodd: I could maybe throw something at url since that gets used for look up, I could also throw an unsigned on the last processes now that I am looking at it.

maxrichardjr (approved):

LEAF_Request_Portal/scripts/process_queries.php
```diff
@@ -5,13 +5,28 @@
...
 $form = new Portal\Form($db, $login);
 
+// last process time
+$vars = [
+    //':lastProcess' => strtotime("30 minutes ago")
+    ':lastProcess' => time()
+];
+
+$processQueryTotalSQL = 'SELECT id,userID,`url` FROM process_query WHERE lastProcess <= :lastProcess group by `url`';
+$processQueryTotalRes = $db->prepared_query($processQueryTotalSQL, $vars);
```
jampaul3, line 60: Just thinking of the future, if we want to get to a more MVC style application then we should start moving queries from flat files like this one into classes. I personally like class names that match Db table names but that is for the team to decide how we want to do that.

Not sure if this query would ever get used again somewhere else but regardless having it here ensures that it can never be used again for anything else. In a class it could be used again if needed without having to write it.

Technically, the if and foreach that follows could be placed in a class as well. Placing this inside a method in a class makes it easier to test. It could have a unit test built to do that. That would leave the flat file really to just kick this all off and let the heavy lifting be done by the class.

Just some thoughts for us to be thinking about.

jampaul3 (approved):

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -612,177 +616,204 @@ var LeafFormGrid = function (containerID, options) {
...
-    }
-    var counter = 0;
-    var validateHtml = document.createElement("div");
-    for (var i = startIdx; i < currentData.length; i++) {
-      if (counter >= limit) {
-        currentRenderIndex = i;
-        break;
+        );
       }
+      var counter = 0;
```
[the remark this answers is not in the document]
shaneodd: let us let lets let.

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -612,177 +616,204 @@ var LeafFormGrid = function (containerID, options) {
...
-    currLimit = limit;
-
-    var fullRender = false;
-    if (startIdx == undefined || startIdx == 0) {
-      startIdx = 0;
-      $("#" + prefixID + "tbody").empty();
-      renderHistory = {};
-      fullRender = true;
-    }
+      if (limit == undefined) {
```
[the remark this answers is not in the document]
shaneodd: went through and worked on the typeofs there will need to be extra regression testing that will need to be done to make sure it does not break things.

LEAF_Request_Portal/sources/Form.php
```diff
@@ -427,33 +408,36 @@ public function newForm($userID)
     public function getIndicator($indicatorID, $series, $recordID = null, $parseTemplate = true)
```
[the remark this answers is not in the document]
shaneodd: Worked on returns and hinting, there were some that I needed to remove hinting and remove return typing. I think this should suffice for a first pass since some of these functions will need some rework.

LEAF_Request_Portal/sources/Form.php
shaneodd: As discussed I removed this and added in a run for each user.

docker/mysql/db/db_upgrade/portal/Update_RMC_DB_2023052500-2023060500.sql
shaneodd: I asked Pete to take a look at this, we will need to hold until his notes arrive.

LEAF_Request_Portal/scripts/process_queries.php
shaneodd: I agree, but as discussed not sure if we want to do things as model per tableor do model perfunctionality. They have their pros and cons. I prefer the model per table since it takes away ambiguity.

maxrichardjr (approved):

maxrichardjr (approved):

LEAF_Request_Portal/api/RESTfulResponse.php
```diff
@@ -674,7 +674,9 @@ private function filterDataActionHistory(array $actionHistory, array $filter): a
      */
     private function filterData(array|string|null $data): array|string|null
     {
-        if (isset($_GET['x-filterData'])) {
+        // we need a way to pass status messages if the data is not going to be sent. I cannot really set this in advance
+        if (isset($_GET['x-filterData']) && !in_array('status',array_keys($data))) {
```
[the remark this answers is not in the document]
shaneodd: Should I be doing that on the auth.leaf side of the house? Changing the status code for me just gave me the blank result set with no sign to the user.  I could have been doing something wrong there where it was not tripping part of the logic but I did not see anything that would handle that.

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -797,15 +797,58 @@ function openShareDialog() {
...
+    let pwd = document.URL.substr(0,document.URL.lastIndexOf('?'));
+    leafSearch.getLeafFormQuery().setLimit(0, maxLimit);
+    let currentQuery = leafSearch.getLeafFormQuery().getQuery();
+    let jsonPath = pwd + leafSearch.getLeafFormQuery().getRootURL() + 'api/form/query/?q=' + JSON.stringify(currentQuery);
+    let powerQueryURL = '<!--{$powerQueryURL}-->' + window.location.pathname;
+
 
     dialog_message.setTitle('Data Endpoints');
-    dialog_message.setContent('<p>This provides a live data source for custom dashboards or automated programs.</p><p><b>A configurable limit of 10,000 records has been preset</b>.</p><br />'
+    dialog_message.setContent('<p>This provides a live data source for custom dashboards or automated programs.</p><p><b>A configurable limit of '+maxLimit.toLocaleString('en-US')+' records has been preset</b>.</p><br />'
```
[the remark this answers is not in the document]
shaneodd: Wouldn't we want to keep some limits in place? The bulk download made sense if some limits were in place, if we are removing that then we would want to not have that option either.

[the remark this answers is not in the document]
mgaoVA: I meant we don't need the preset 10,000 limit. Internally we'll still support the limit parameter since it's used as part of a "lazy loading" feature.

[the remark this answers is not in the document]
shaneodd: Okay removing.

maxrichardjr (approved):

LEAF_Request_Portal/api/RESTfulResponse.php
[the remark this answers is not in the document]
shaneodd: I have this updated

docker/mysql/db/db_upgrade/portal/Update_RMC_DB_2023052500-2023060500.sql
shaneodd: Pete and I went over this and cleaned this up, there were a few other things that I did not think were issues like BTREE and ollation which were added by HEIDI when I was exporting the table from that. I cleaned that up as well.

maxrichardjr (approved):

Pelentan (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

Pelentan (approved):

aerinkayne (approved):

raphaelfretter (approved):

maxrichardjr (approved):

maxrichardjr (approved):

pete-nerantzinis (approved):

AmandaSumner (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

jrodcode (approved):

maxrichardjr (approved):

maxrichardjr (approved):

mgaoVA:
I added prerequisites for Prod deployment. Good catch on the potential impact on custom code. We will need to scan custom templates and provide guidance to admins.

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

PR #2011, opened 2023-06-22
Create formSearchMultisite.js

mgaoVA:
This is used in a multi-site search tool and is an abbreviated version of formSearch.js. Search features that involve custom form/workflow content have been removed.

Example on prod: https://leaf.va.gov/Academy/Demo1/report.php?a=Sitemap_Search

This example allows administrators to easily find requests within all sites registered in the Sitemap Editor.

### DIFF
--- a/LEAF_Request_Portal/js/formSearchMultisite.js
+++ b/LEAF_Request_Portal/js/formSearchMultisite.js
@@ -1,0 +1,1111 @@
+/************************
+ Form Search Widget for multiple sites
+ Search features are limited to common denominators across all LEAF sites.
+ */
+
+var LeafFormSearchMultisite = function (containerID) {
+  var containerID = containerID;
+  var prefixID = "LeafFormSearchMultisite" + Math.floor(Math.random() * 1000) + "_";
+  var localStorageNamespace = "LeafFormSearchMultisite" + getLocalStorageHash();
+  var orgchartPath = "";
+  var timer = 0;
+  var q = "";
+  var intervalID = null;
+  var currRequest = null;
+  var numResults = 0;
+  var searchFunc = null;
+  var leafFormQuery = new LeafFormQuery();
+  var widgetCounter = 0;
+  var rootURL = "";
+
+  // constants
+  var ALL_DATA_FIELDS = "0";
+  var ALL_OC_EMPLOYEE_DATA_FIELDS = "0.0";
+
+  function renderUI() {
+    $("#" + containerID).html(
+      '<div style="display:flex; align-items:center; width:fit-content; width: -moz-fit-content;">\
+			    <img id="' +
+        prefixID +
+        'searchIcon" class="searchIcon" alt="search" style="vertical-align: middle; padding-right: 4px; display: inline;" src="' +
+        rootURL +
+        'dynicons/?img=search.svg&w=16">\
+			    <img id="' +
+        prefixID +
+        'searchIconBusy" class="searchIcon" alt="loading" style="vertical-align: middle; padding-right: 4px; display:none" src="' +
+        rootURL +
+        'images/indicator.gif">\
+					<span style="position: absolute; width: 60%; height: 1px; margin: -1px; padding: 0; overflow: hidden; clip: rect(0,0,0,0); border: 0;" aria-atomic="true" aria-live="polite" class="status" role="status"></span>\
+			    <input style="border: 1px solid black; padding: 4px" type="text" id="' +
+        prefixID +
+        'searchtxt" name="searchtxt" size="50" title="Enter your search text" value="" />\
+			    <button class="buttonNorm" id="' +
+        prefixID +
+        'advancedSearchButton">Advanced Options</button>\
+			    <fieldset id="' +
+        prefixID +
+        'advancedOptions" style="position: relative; display: none; margin: 0px; border: 1px solid black; background-color: white">\
+		        <legend>Advanced Search Options</legend>\
+		        <button id="' +
+        prefixID +
+        'advancedOptionsClose" style="float: right; margin-top: -20px; margin-right: -14px; display: none; cursor: pointer; background-image:url(' +
+        rootURL +
+        'dynicons/?img=process-stop.svg&w=16); height: 16px;width: 16px; border: none; background-color: transparent; text-indent: -9999em" alt="Close advanced search">Close advanced search</button>\
+		        <div style="width: 550px">Find items where...</div>\
+		        <table id="' +
+        prefixID +
+        'searchTerms"></table>\
+		        <button class="buttonNorm" id="' +
+        prefixID +
+        'addTerm" style="float: left">And...</button>\
+		        <button class="buttonNorm" id="' +
+        prefixID +
+        'orTerm" style="float: left">Or...</button>\
+		        <br /><br />\
+		        <button id="' +
+        prefixID +
+        'advancedSearchApply" class="buttonNorm" style="text-align: center; width: 100%">Apply Filters</button>\
+		    </fieldset>\
+		    </div>\
+		    <div id="' +
+        prefixID +
+        '_result" style="margin-top: 8px" aria-label="Search Results">\
+		    </div>'
+    );
+
+    var searchOrigWidth = 0;
+    $("#" + prefixID + "advancedOptionsClose").on("click", function () {
+      localStorage.setItem(localStorageNamespace + ".search", "");
+      $("#" + prefixID + "searchtxt").val("");
+      search("");
+      $("#" + prefixID + "advancedOptionsClose").css("display", "none");
+      $("#" + prefixID + "advancedOptions").slideUp(function () {
+        $("#" + prefixID + "advancedSearchButton").fadeIn();
+        $("#" + prefixID + "searchtxt").css("display", "inline");
+        $("#" + prefixID + "searchtxt").animate(
+          { width: searchOrigWidth },
+          400,
+          "swing"
+        );
+        $("#" + prefixID + "searchtxt").focus();
+      });
+    });
+    //added for keyboard navigation and accessibility to close advanced search options
+    var searchOrigWidth = 0;
+    $("#" + prefixID + "advancedOptionsClose").on("keydown", function (e) {
+      if (e.keyCode == 13) {
+        localStorage.setItem(localStorageNamespace + ".search", "");
+        $("#" + prefixID + "searchtxt").val("");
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+        $("#" + prefixID + "searchtxt").animate(
+          { width: searchOrigWidth },
+          400,
+          "swing"
+        );
+        $("#" + prefixID + "searchtxt").focus();
+      });
+    });
+    //added for keyboard navigation and accessibility to close advanced search options
+    var searchOrigWidth = 0;
```
Pelentan, line 94: This is the same as line 76 above.  And does this need to be a var?  Looks like it would be a case for let

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
+/************************
+ Form Search Widget for multiple sites
+ Search features are limited to common denominators across all LEAF sites.
+ */
+
+var LeafFormSearchMultisite = function (containerID) {
+  var containerID = containerID;
```
Pelentan, line 7: Do these all need to be vars?  Or can they be lets?

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+      window.location.search !== ""
+    ) {
+      focus();
+    }
+    if (getLastSearch() == null) {
+      search("*");
+    } else {
+      var lastSearch = getLastSearch();
+
+      var isJSON = true;
```
Pelentan, line 181: These two variables are declared vars in multiple menthods.  It might be better to declare them outside of the methods once if they absolutely need to be global.

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+   * @memberOf LeafFormSearchMultisite
+   */
+  function inputLoop() {
+    if ($("#" + prefixID + "searchtxt") == null) {
+      clearInterval(intervalID);
+      return false;
+    }
+    timer += timer > 5000 ? 0 : 200;
+    if (timer > 400) {
+      var txt = $("#" + prefixID + "searchtxt").val();
```
Pelentan, line 337: There are a lot of vars in here and no lets.  Is this copied from other code?  We've been working on updating code to standards as we come across it.

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+  function showNotBusy() {
+    $("#" + prefixID + "searchIcon").css("display", "inline");
+    $("#" + prefixID + "searchIconBusy").css("display", "none");
+  }
+
+  /**
+   * @memberOf LeafFormSearchMultisite
+   */
+  function createEmployeeSelectorWidget(widgetID, type) {
+    if (type == undefined) {
```
Pelentan, line 393: Do we want this to be typeof like below?

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+        '<link type="text/css" rel="stylesheet" href="' +
+          orgchartPath +
+          '/css/employeeSelector.css" />'
+      );
+      $.ajax({
+        type: "GET",
+        url: orgchartPath + "/js/employeeSelector.js",
+        dataType: "script",
+        success: function () {
+          let empSel = new employeeSelector(prefixID + "widgetEmp_" + widgetID);
```
Pelentan, line 407: Need a fail-catch for the ajax function.

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+      $("head").append(
+        '<link type="text/css" rel="stylesheet" href="' +
+          orgchartPath +
+          '/css/positionSelector.css" />'
+      );
+      $.ajax({
+        type: "GET",
+        url: orgchartPath + "/js/positionSelector.js",
+        dataType: "script",
+        success: function () {
```
Pelentan, line 478: fail-catch

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+      $("head").append(
+        '<link type="text/css" rel="stylesheet" href="' +
+          orgchartPath +
+          '/css/groupSelector.css" />'
+      );
+      $.ajax({
+        type: "GET",
+        url: orgchartPath + "/js/groupSelector.js",
+        dataType: "script",
+        success: function () {
```
Pelentan, line 523: fail-catch

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+  }
+
+  /**
+   * Render the query match condition's input type for dropdown and radio fields
+   * @param widgetID
+   * @param options <select> html section matching the widgetID
+   * @memberOf LeafFormSearchMultisite
+   */
+  function renderSingleSelectInputType(widgetID, options) {
+    switch ($("#" + prefixID + "widgetCod_" + widgetID).val()) {
```
Pelentan, line 561: we've been using toLower on any text fed to a switch and then putting all the putting all the case in lower.

At least for ones that we're creating or small ones that are an easy fix.  I see another below that you brought over from other code.  We've been kinda letting those slide for now.

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+            '" class="chosen"  aria-label="title" style="width: 120px">\
+                        <option value="=">IS</option>\
+                        <option value="!=">IS NOT</option>\
+                    </select>'
+        );
+        $.ajax({
+          type: "GET",
+          url: "./api/system/services",
+          dataType: "json",
+          success: function (res) {
```
Pelentan, line 620: fail-catch

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+          '<select id="' +
+            prefixID +
+            "widgetCod_" +
+            widgetID +
+            '" style="width: 140px" class="chosen" aria-label="categoryID">\
+	            		<option value="=">IS</option>\
+	            		<option value="!=">IS NOT</option>\
+	            	</select>'
+        );
+        $.ajax({
```
Pelentan, line 684: fail-catch

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+            prefixID +
+            "widgetCod_" +
+            widgetID +
+            '" value="=" /> ='
+        );
+        $.ajax({
+          type: "GET",
+          url: "./api/workflow/dependencies",
+          dataType: "json",
+          success: function (res) {
```
Pelentan, line 746: fail-catch

LEAF_Request_Portal/js/formSearchMultisite.js
```diff
@@ -0,0 +1,1111 @@
...
+
+    widgetCounter++;
+  }
+
+  /**
+   * @memberOf LeafFormSearchMultisite
+   */
+  function generateSearchQuery() {
+    leafFormQuery.clearTerms();
+    for (var i = 0; i < widgetCounter; i++) {
```
Pelentan, line 1011: var to let

Pelentan (changes requested):

Pelentan (changes requested):

maxrichardjr (approved):

LEAF_Request_Portal/js/formSearchMultisite.js
mgaoVA: Nice catch. This was a copy/paste job. I'll update these here, but updating the other file should go into a different effort to reduce risk.

LEAF_Request_Portal/js/formSearchMultisite.js
mgaoVA: While both are valid, I like the former better since it describes what we're doing more accurately: checking to see if the variable is undefined rather than checking that the type is undefined.

I went down a rabbit hole for this one... apparently "typeof" was preferred (in jQuery) for Global variables in the past because "undefined" used to be a mutable variable in older browsers.

LEAF_Request_Portal/js/formSearchMultisite.js
mgaoVA: In this situation I'd want to keep it as-is because the variable being fed into this uses an SQL all-caps convention, which should be consistent everywhere. The input is controlled by us, and if there's a case sensitivity issue, it would mean the upstream source should be fixed to ensure consistency.

LEAF_Request_Portal/templates/reports/LEAF_Sitemap_Search.tpl
```diff
@@ -0,0 +1,1417 @@
...
+    function setExportFormat() {
+        if($('#shortenLink').css('display') === 'none') {
+            $('#exportFormat').html('?');
+        }
+        else {
+            $('#exportFormat').html('&');
+        }
+
+
+        switch($('#format').val()) {
```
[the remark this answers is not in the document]
mgaoVA: Same as the other one -- we control these, and upstream should be made consistent if it doesn't match.

mgaoVA:
Updated to clean up var/let scopes and a document-wide fail catch for ajax calls.

I think the issues relating to "typeOf" might be ok in modern browsers, unless I missed something. I'd prefer to have code readability be more aligned with our intent.

The instances here not using toLower should be ok as well, since these are generated by us. If the variable doesn't have an exact match it would mean it's technically wrong.

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

Pelentan (approved):

maxrichardjr (approved):

jampaul3 (approved):

shaneodd (approved):

AmandaSumner (approved):

PR #2014, opened 2023-06-26
Leaf 3818 - connect button to new api

jampaul3:


maxrichardjr (approved):

maxrichardjr (approved):

LEAF_Nexus/sources/Employee.php
```diff
@@ -171,7 +171,7 @@ private function formatUserNames(array $employee_list): array
     {
         $local_employee_usernames = [];
 
-        foreach ($employee_list as $employee) {
+        foreach ($employee_list['data'] as $employee) {
```
shaneodd, line 174: Should 'data' be checked if it exists and is array?

shaneodd (changes requested):

LEAF_Nexus/sources/Employee.php
raphaelfretter: The employee_list appears to be checked if empty before calling this method

shaneodd: employee_list  is checked but the ['data'] would not be checked. We would know employee_list  is an array at this point however what is ['data']

jampaul3: In an attempt to get the json responses in the format that Michael Shaffer is looking for them in data is just that, it's the data being sent back. The json response has 2 elements to it, the status and the data. status just gives the status with a code and message (the message is generally blank when the status is ok,), the  data is usually going to be an array of the data being returned.

PR #2015, opened 2023-06-26
LEAF 3782 Vue UI config update and Site Designer (off)

aerinkayne:
The Webpack config for the Vue_UI container has been updated to build both the Form Editor and Site Designer Vue apps to their respective folders in libs/js/vue-dest and extract CSS files from the build.  This allows CSS to be brought in through the php index files prior to the JS files and also decreases the JS files sizes.
A webpack config for the admin SASS folder has also been added but will remain commented out for now.

This branch also brings in code updates from previously reviewed Jira 3444 in an OFF state.
-- The admin card to the site_designer template is commented out.
-- The site_designer_vue.tpl file is wrapped with  so that it is also not available through a direct URL

Other:
A minor update to Vue resulted in a deprecation warning for a setting used in the Vue app main.js files.  This setting is no longer needed and has been removed.   Fixed unintended removal of the refresh-side IFTHEN 'loading' message.


maxrichardjr (approved):

LEAF_Request_Portal/js/vue_conditions_editor/LEAF_conditions_editor.js
```diff
@@ -707,9 +707,8 @@ const ConditionsEditor = Vue.createApp({
      * @returns {Array} of condition objects
      */
     savedConditions() {
-      return this.childIndicator.conditions
-        ? JSON.parse(this.childIndicator.conditions)
-        : [];
+      return typeof this.childIndicator.conditions === 'string' && this.childIndicator.conditions[0] === '[' ?
```
raphaelfretter, line 710:
```suggestion
      return typeof this.childIndicator.conditions === 'string' && this.childIndicator.startsWith('[') ?

```

raphaelfretter (approved):

LEAF_Request_Portal/js/vue_conditions_editor/LEAF_conditions_editor.js
aerinkayne: I'll note this for when I do the next round of refresh updates.  I did double-check to make sure that [0] wouldn't break it here in the case of an empty string, so both essentially do the same thing.  We don't need to worry about IE any more though, so this choice is a bit neater.

Pelentan (approved):

AmandaSumner (approved):

PR #2019, opened 2023-06-28
LEAF-3561 - Upload employee data to orgchart

shaneodd:
LEAF_Nexus -> DATABASE_DB_DIRECTORY should be setup to have a local "NATIONAL" database.

Excel doc will need to be created with employee email, supervisor email and position title

The employee email should not exist in the local orgchart but has to exist in the national db that you created. The supervisor and position title must exist, the position title must exist and have the supervisor assigned to that position.

### DIFF
--- a/libs/jsapi/nexus/LEAFNexusAPI.js
+++ b/libs/jsapi/nexus/LEAFNexusAPI.js
@@ -76,7 +76,7 @@ var NexusEmployeeAPI = function (baseAPIURL) {
         setCSRFToken = function (token) { csrfToken = token; },
 
         getByEmail = function (emailAddress, onSuccess, onFail) {
-            var fetchURL = apiURL + '/search&q=' + emailAddress + '&noLimit=0';
+            var fetchURL = apiURL + '/search?q=' + emailAddress + '&noLimit=0';
 
             $.ajax({
                 method: 'GET',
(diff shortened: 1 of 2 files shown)


aerinkayne (approved):

libs/jsapi/nexus/LEAFNexusAPI.js
```diff
@@ -76,7 +76,7 @@ var NexusEmployeeAPI = function (baseAPIURL) {
         setCSRFToken = function (token) { csrfToken = token; },
 
         getByEmail = function (emailAddress, onSuccess, onFail) {
-            var fetchURL = apiURL + '/search&q=' + emailAddress + '&noLimit=0';
+            var fetchURL = apiURL + '/search?q=' + emailAddress + '&noLimit=0';
```
Pelentan, line 79: I'm thinking these should be "let"s.  Here and the one(s) below.

Pelentan (changes requested):

libs/jsapi/nexus/LEAFNexusAPI.js
shaneodd: I have these vars updated.

maxrichardjr (approved):

jampaul3 (approved):

AmandaSumner (approved):

PR #2037, opened 2023-07-10
Leaf - 3364 user groups users with backups not right

jampaul3:
This change requires some db modifications. The reason we are having the backup issues within the user groups is because of the index on users in the portal db. The primary key is set on userName and groupID but when a person is a member of the group and they are a backup to another member they aren't going to show up for one or the other because they are only in the db once.

The main change to all this is:
The primary key is now based on userName, groupID, and backupID - this allows for a person to be a member of the group and also be a backup to one or more other members of the group.

There are other changes that made this change possible, like the backupID is no longer NULLABLE, it has  to be a blank string to be included in the primary key. The code reflects these changes by looking for a blank string instead of a NULL value.


LEAF_Request_Portal/scripts/firstRun.php
```diff
@@ -12,8 +12,8 @@
 if (count($res) == 0) {
     if (strlen(DATABASE_DB_ADMIN) > 0) {
         $vars = array(':name' => DATABASE_DB_ADMIN);
-        $res = $db->prepared_query('INSERT INTO users (userID, groupID)
-                                        VALUES (:name, 1)', $vars);
+        $res = $db->prepared_query('INSERT INTO users (userID, groupID, backupID)
```
Pelentan, line 15: This isn't best practices.  We should probably modify the database so it has a default of empty-string from that column.  Otherwise, somewhere down the road, nulls will start to creep in.

LEAF_Request_Portal/sources/Group.php
```diff
@@ -269,54 +269,82 @@ public function getMembers($groupID, bool $searchDeleted = false): array|string
      *
      * @return string
      */
-    public function addMember($member, $groupID): string
+    public function addMember($member, $groupID): array
```
Pelentan, line 272: Do we want to do type-hinting here?  Or might that cause issues?  I ask because normally you're pretty good about those and if you didn't use them here, I'm guessing there's a reason.

LEAF_Request_Portal/sources/Group.php
```diff
@@ -269,54 +269,82 @@ public function getMembers($groupID, bool $searchDeleted = false): array|string
...
     {
         $oc_db = new \Leaf\Db(\DIRECTORY_HOST, \DIRECTORY_USER, \DIRECTORY_PASS, \ORGCHART_DB);
         $employee = new \Orgchart\Employee($oc_db, $this->login);
 
         if (is_numeric($groupID)) {
-            $sql_vars = array(':userID' => $member,
-                ':groupID' => $groupID,);
+            $vars = array(':userID' => $member,
```
Pelentan, line 278: Just a quick chuckle here.  Every time I see this in everyone's code, my knee-jerk reaction is "shouldn't this be a 'let'?"  Takes me a second to switch from js to php.  And then I had the thought, it would be kinda funny to mess with future developers if we started using the variable name "$lets" for things like this.  

To be clear, no change needed.  Just thought I would share where my brain was going.  ;-)

LEAF_Request_Portal/sources/Service.php
```diff
@@ -173,8 +173,8 @@ public function addMember($groupID, $member)
             {
                 $sql_vars = array(':userID' => $member,
                               ':groupID' => $groupID, );
-                $this->db->prepared_query('INSERT INTO users (userID, groupID)
-                                                VALUES (:userID, :groupID)', $sql_vars);
+                $this->db->prepared_query('INSERT INTO users (userID, groupID, backupID)
```
Pelentan, line 176: This goes back to making the modification on the database itself.  Otherwise you have to make sure every single insert has the empty value.

docker/mysql/db/db_upgrade/portal/Update_RMC_DB_2023052500-2023072000.sql
```diff
@@ -0,0 +1,32 @@
+START TRANSACTION;
```
Pelentan, line 1: And since you're already modifying... Let's put it in.  ;-)

Pelentan (changes requested):

maxrichardjr (commented):
I don't see any code?

LEAF_Request_Portal/scripts/firstRun.php
jampaul3: Default of '' added

LEAF_Request_Portal/sources/Group.php
jampaul3: I hadn't thought about it and just missed it in this case. The string hint was already there and I needed to change that obviously but didn't even cross my mind for the others. Not sure if we might run into anything if we add them here. I did look to see where this method is used and decided to add them. Thing is if the come in as null (which I think is the biggest culprit here) and it adds it to the db it would be a dead record with no way to get the data because the userID and groupID are required for retrieval.

LEAF_Request_Portal/sources/Service.php
jampaul3: I did make sure of that but added the default value anyway, that is a better way of doing it.

jampaul3:
@maxrichardjr  if you are looking at the last commit that's because it was a clean merge, nothing to show.

maxrichardjr:
Ok, appreciate it Jamie.


Max Richard Jr | Sr. QA Automation Tester
AdaptiveStack Technologies, Inc.
623 N. Grandview Ave, Suite 101, Daytona Beach FL 32118
(c)  205-902-8540
***@***.******@***.***> | www.adaptivestack.com<http://www.adaptivestack.com/> | ***@***.*** <https://www.facebook.com/adaptivestack/>   ***@***.***
***@***.***
This electronic message contains information from AdaptiveStack Technologies Inc. , which may be company sensitive, proprietary, privileged or otherwise protected from disclosure. The information is intended to be used solely by the recipient(s) named above. If you are not an intended recipient, be aware that any review, disclosure, copying, distribution or use of this transmission or its contents is prohibited.   With the exception of messages sent by authorized AdaptiveStack Technologies Inc. contracts or purchasing personnel, nothing in this message may be interpreted as a digital or electronic signature that can be used to: (a) authenticate either the submission or the acceptance of a proposal or offer to contract, or (b) modify an existing contract. If you have received this transmission in error, please notify us immediately at ***@***.******@***.***>.

maxrichardjr (approved):

LEAF_Request_Portal/sources/Group.php
```diff
@@ -269,20 +269,23 @@ public function getMembers($groupID, bool $searchDeleted = false): array|string
      *
      * @return string
      */
-    public function addMember($member, $groupID): string
+    public function addMember(string $member, int $groupID): array
```
Pelentan, line 272: Is this a function we can easily update to the preferred "json" return?

LEAF_Request_Portal/sources/Group.php
```diff
@@ -325,20 +344,21 @@ public function addMember($member, $groupID): string
...
+        $sql = 'INSERT INTO users (groupID, userID, backupID)
+                VALUES (:groupID, :userID, :backupID)
+                ON DUPLICATE KEY UPDATE userID=:userID';
 
-        $result = $this->db->prepared_query('INSERT INTO users (groupID, userID, backupID)
-                                                    VALUES (:groupID, :userID, :backupID)
-                                                    ON DUPLICATE KEY UPDATE userID=:userID', $sql_vars);
+        $result = $this->db->pdo_insert_query($sql, $vars);
 
         return (array) $result;
```
Pelentan, line 363: Same here?

Pelentan (changes requested):

raphaelfretter:
I have checked the LEAF-Scripts repository and it looks like the Vax Portal will be compatible with these changes.

raphaelfretter (approved):

Pelentan (approved):

LEAF_Request_Portal/sources/Group.php
jampaul3: This one already is, the pdo_insert_query returns a properly formatted json response.

LEAF_Request_Portal/sources/Group.php
jampaul3: This is returning the properly formatted json response but php doesn't have json as a type hint. The only thing that could be done differently here would be to json_encode and send it back as a string. But I think it would be better to send the array here and if it's actually an endpoint have the endpoint json_encode it before sending it back to the caller.

maxrichardjr (approved):

AmandaSumner (approved):

PR #2038, opened 2023-07-10
LEAF-3822: Destruction of LEAF Records - Backend

pete-nerantzinis:


LEAF_Request_Portal/sources/System.php
```diff
@@ -1096,4 +1096,139 @@ private function updateNexusWithPortalGroups(array $portal_groups, \Orgchart\Gro
...
     }
+
+    /**
+     * Create flag (days) for destruction records
+     *
+     * @param int $flag - number of days to mark a record for destruction
+     *
+     * @return bool
+     */
+    public function setDestructionFlag($flag = null): bool
```
shaneodd, line 1107: Dont we want typing on the function parameters as well?

shaneodd (changes requested):

LEAF_Request_Portal/sources/System.php
```diff
@@ -1096,4 +1098,124 @@ private function updateNexusWithPortalGroups(array $portal_groups, \Orgchart\Gro
...
+        if ($flag) {
+            $vars = array(':data' => $flag);
+            $strSQL = 'INSERT INTO `settings` (`setting`, `data`)
+            VALUES ("destructionFlag", :data)
+            ON DUPLICATE KEY UPDATE `data`=:data';
+
+            $this->db->prepared_query($strSQL, $vars);
+
+            return true;
+        }
```
jampaul3, line 1118: I assume most if not all of these methods will be used via an api call. Don't we want to start making sure our responses are in a proper json format? I have been using 
                    $return_value = array (
                        'status' => array (
                            'code' => 2,
                            'message' => ''
                        ),
                      'data' => `array of data to be returned`
                    );

Not saying what I have been using is what we ought to use but that is something we should talk about as a team.

Pelentan: Yes, please.

LEAF_Request_Portal/sources/System.php
```diff
@@ -1096,4 +1096,139 @@ private function updateNexusWithPortalGroups(array $portal_groups, \Orgchart\Gro
...
+    /**
+     * Create flag (days) for destruction records
+     *
+     * @param int $flag - number of days to mark a record for destruction
+     *
+     * @return bool
+     */
+    public function setDestructionFlag($flag = null): bool
+    {
+        if ($flag) {
```
Pelentan, line 1109: $flag isn't a boolean.  You'll want to use empty() here.

LEAF_Request_Portal/sources/System.php
```diff
@@ -1096,4 +1096,139 @@ private function updateNexusWithPortalGroups(array $portal_groups, \Orgchart\Gro
...
+
+        return false;
+    }
+
+    /**
+     * Get flag (days) for destruction records
+     *
+     * @return string|null
+     */
+    public function getDestructionFlag(): string|null
```
Pelentan, line 1128: New functions return json/array as Jamie pointed out above.

LEAF_Request_Portal/sources/System.php
```diff
@@ -1096,4 +1096,139 @@ private function updateNexusWithPortalGroups(array $portal_groups, \Orgchart\Gro
...
+    {
+        $strSQL = 'SELECT `data` FROM `settings` WHERE `settings`.`setting` = destructionFlag';
+        $res = $this->db->query($strSQL);
+
+        if (count($res) > 0 ) {
+            return $res['data'];
+        }
+
+        return null;
```
Pelentan, line 1137: And one return as we talked about before.

LEAF_Request_Portal/sources/System.php
```diff
@@ -1096,4 +1096,139 @@ private function updateNexusWithPortalGroups(array $portal_groups, \Orgchart\Gro
...
+     *
+     * @param int $flag - number of days to mark a record for destruction
+     *
+     * @return array|null
+     */
+    public function markForDestruction(): array|null
+    {
+        $flag = $this->getDestructionFlag();
+
+        if ($flag) {
```
Pelentan, line 1151: $flag isn't a boolean

LEAF_Request_Portal/sources/System.php
```diff
@@ -1096,4 +1096,139 @@ private function updateNexusWithPortalGroups(array $portal_groups, \Orgchart\Gro
...
+                    $destructionArray[] = [
+                        'recordID' => $record['recordID'],
+                        'destructionGraceTime' => date('Y-m-d', strtotime("+30 days"))
+                    ];
+                }
+
+                $this->db->insert_batch('records', $destructionArray, ['destructionGraceTime']);
+            }
+
+            return $destructionArray;
```
Pelentan, line 1182: Same as above.

Pelentan: So it shall be below.  ;-)

Pelentan (changes requested):

PR #2058, opened 2023-07-21
Leaf 3891 - Remove unused empUID argument

mgaoVA:
empUID is passed into the API, but it's not used in the associated PHP function: https://github.com/department-of-veterans-affairs/LEAF/blob/master/LEAF_Nexus/sources/Employee.php#L87

```php
public function refresh(string $user_name): array
```

### DIFF
--- a/LEAF_Nexus/api/controllers/EmployeeController.php
+++ b/LEAF_Nexus/api/controllers/EmployeeController.php
@@ -69,7 +69,7 @@ class EmployeeController extends RESTfulResponse
         });
 
 
-        $this->index['POST']->register('employee/refresh/[text]/[digit]', function ($args) use ($employee, $national_db) {
+        $this->index['POST']->register('employee/refresh/[text]', function ($args) use ($employee, $national_db) {
             if (!$national_db) {
                 $return_value = $employee->refresh($args[0]);
             } else {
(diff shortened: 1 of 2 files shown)


LEAF_Nexus/api/controllers/EmployeeController.php
```diff
@@ -69,7 +69,7 @@ public function post($act)
         });
 
 
-        $this->index['POST']->register('employee/refresh/[text]/[digit]', function ($args) use ($employee, $national_db) {
+        $this->index['POST']->register('employee/refresh/[text]', function ($args) use ($employee, $national_db) {
```
jampaul3, line 72: Is it possible that someone might be using this endpoint in custom code? If so then there call isn't going to find the endpoint they are pointing to.

maxrichardjr (approved):

raphaelfretter (approved):

jrodcode (approved):

LEAF_Nexus/api/controllers/EmployeeController.php
mgaoVA: It's possible, but unlikely as this endpoint was only added last month.

To be sure, I've added it back and opened https://github.com/department-of-veterans-affairs/LEAF/issues/2062

maxrichardjr (approved):

jampaul3 (approved):

AmandaSumner (approved):

aerinkayne (approved):

PR #2060, opened 2023-07-24
LEAF-2568: Create Automated Scripts for Moving Vax Data

raphaelfretter:


scripts/automatedVaxScripts.php
```diff
@@ -0,0 +1,21 @@
...
+echo date('Y-m-d g:i:s a') . "\r\n";
+
+try {
+
+    // in the future stuff like this will be spun off through a different program.
+    // Until then to keep thinks straight forward this could be done in individual files to be called.
+    $test = new TimeBracketCmd('automatedVaxScripts.sh');
+
+    $test->setRunAtExactTime('6 am');
+    $test->run();
```
Pelentan, line 15: Ok.  This might just be because my brain is pretty much dead today.  But it looks like all the scripts in the automatedVaxScripts file would be run at the same time?  This might be a good time to add onto the "cron" class so that it alerts if there is a failure to run a script.  With a flag that can be set so if there are multiple scripts to be run, whether the entire process should fail if one fails to complete.  You would need to get with Shane on this.

Pelentan (changes requested):

maxrichardjr (approved):

shaneodd (approved):

jampaul3 (approved):

Pelentan (approved):

PR #2063, opened 2023-07-26
Leaf 3841 - First round duplicate workflows

jampaul3:
I believe I  got all aspects of the workflow in here. It doesn't matter how complete a workflow is it will duplicate everything in the current workflow.


raphaelfretter (approved):

shaneodd (approved):

maxrichardjr (approved):

maxrichardjr (approved):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -2272,6 +2134,398 @@ console.log(params);
         dialog.show();
     }
 
+    function duplicateWorkflow() {
```
Pelentan, line 2137: All of the below functions need comments on them, including describing the variables needed.

Also, should probably consider adding in some default values to ensure things don't break silently.

LEAF_Request_Portal/sources/Workflow.php
```diff
@@ -383,6 +376,35 @@ public function createAction($stepID, $nextStepID, $action)
...
     }
 
+    /**
+     * @param int $workflowID
+     * @param int $stepID
+     * @param int $nextStepID
+     * @param string $action
+     * @param string $conditional
+     *
+     * @return array
```
Pelentan, line 386: Thinking "out loud" here.  Should it be noted that the return from the db class call is in the new format?  Or do you think it would just be a waste of typing as generally no one reads these anyway?

Pelentan (changes requested):

LEAF_Request_Portal/sources/Workflow.php
jampaul3: That's up to you, I can add them, and adding them to the PHPDoc area would be a good place for them. If anyone does read the doc at least it tells you.

jampaul3: done

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
jampaul3: They have been added.

Pelentan (approved):

maxrichardjr (approved):

maxrichardjr (approved):

AmandaSumner (approved):

aerinkayne (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

PR #2088, opened 2023-08-16
Leaf 3902 - Update groups, service and user

jampaul3:
This started out as a revamp of services but because of found bugs it became a revamp of services and update of user groups.

This will require a db update, I would recommend keeping a copy of the db as it is before testing this so that you can go back to that db if you need to work on something else.

For testing it is important to check both sides, Portal and Nexus. When a change happens on the Nexus side it should be reflected on the portal upon a sync services or closing out of the group modal. Also, multiple backups should be checked along with mutual backups of people.

I think the biggest part of testing is going to be emails. Making sure that the correct people get the emails and such.

I also updated the inactive user section for both user and services. You are only allowed to prune locally managed non Nexus members so in place of the prune button I put a reactivate button for that person (This will prevent the user from having to add them locally or removing the group and re-adding them)

The services modal section is now updated to look like user access groups. The exception to this is a local member shouldn't be allowed to be added to Nexus so that button was removed. It also doesn't show a Nexus check mark.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_groups.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_groups.tpl
@@ -395,16 +395,20 @@ function toTitleCase(str) {
 }
 
 function addAdmin(userID) {
-    $.ajax({
-        type: 'POST',
-        url: "../api/group/" + 1 + "/members",
-        data: {'userID': userID,
-               'CSRFToken': '<!--{$CSRFToken}-->'},
-        success: function(response) {
-        	getMembers(1);
-        },
-        cache: false
-    });
+    if (userID === '') {
+        return;
+    } else {
+        $.ajax({
+            type: 'POST',
+            url: "../api/group/" + 1 + "/members",
+            data: {'userID': userID,
+                'CSRFToken': '<!--{$CSRFToken}-->'},
+            success: function(response) {
+                getMembers(1);
+            },
+            cache: false
+        });
+    }
 }
 
 function removeAdmin(userID) {
--- a/LEAF_Request_Portal/admin/templates/mod_groups.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_groups.tpl
@@ -353,6 +353,18 @@ function pruneMember(groupID, userID) {
     });
 }
 
+function reactivateMember(groupID, userID) {
+    $.ajax({
+        type: 'POST',
+        url: "../api/group/" + groupID + "/members/_" + userID + "/reactivate",
+        data: {'CSRFToken': '<!--{$CSRFToken}-->'},
+        fail: function(err) {
+            console.log(err);
+        },
+        cache: false
+    });
+}
+
 function addNexusMember(groupID, empUID) {
     $.ajax({
         type: 'POST',
@@ -385,9 +397,8 @@ function toTitleCase(str) {
 function addAdmin(userID) {
     $.ajax({
         type: 'POST',
-        url: "ajaxIndex.php?a=add_user",
+        url: "../api/group/" + 1 + "/members",
         data: {'userID': userID,
-               'groupID': 1,
                'CSRFToken': '<!--{$CSRFToken}-->'},
         success: function(response) {
         	getMembers(1);
@@ -515,7 +526,12 @@ function getGroupList() {
                                                 actions += '</td>';
                                                 employee_table += `<tr>${employeeName}${employeeUserName}${backups}${isLocal}${isRegional}${actions}</tr>`;
                                             } else {
-                                                let pruneMemberButton = `<td style="font-size: 0.8em; text-align: center;"><button id="pruneMember_${counter}" class="usa-button usa-button--secondary leaf-btn-small leaf-font0-8rem" style="font-size: 0.8em; display: inline-block; float: left; margin: auto; min-width: 4rem;" title="Prune this user from this group">Prune</button>`;
+                                                let pruneMemberButton = '';
+                                                if (res[i].regionallyManaged === false) {
+                                                    pruneMemberButton = `<td style="font-size: 0.8em; text-align: center;"><button id="pruneMember_${counter}" class="usa-button usa-button--secondary leaf-btn-small leaf-font0-8rem" style="font-size: 0.8em; display: inline-block; float: left; margin: auto; min-width: 4rem;" title="Prune this user from this group">Prune</button>`;
+                                                } else {
+                                                    pruneMemberButton = `<td style="font-size: 0.8em; text-align: center;"><button id="reActivateMember_${counter}" class="usa-button usa-button leaf-btn-small leaf-font0-8rem" style="font-size: 0.8em; display: inline-block; float: left; margin: auto; min-width: 4rem;" title="Reactivate this user for this group">Reactivate</button>`;
+                                                }
                                                 let actions = `${pruneMemberButton}`;
                                                 actions += '</td>';
                                                 inactive_table += `<tr>${employeeName}${employeeUserName}${backups}${isLocal}${isRegional}${actions}</tr>`;
@@ -565,6 +581,16 @@ function getGroupList() {
                                                     dialog_confirm.setContent('Are you sure you want to prune this member?');
                                                     dialog_confirm.setSaveHandler(function () {
                                                         pruneMember(groupID, res[i].userName);
+                                                        dialog_confirm.hide();
+                                                        dialog.hide();
+                                                    });
+                                                    dialog_confirm.show();
+                                                });
+
+                                                $('#reActivateMember_' + counter).on('click', function () {
+                                                    dialog_confirm.setContent('Are you sure you want to Reactivate this member?');
+                                                    dialog_confirm.setSaveHandler(function () {
+                                                        reactivateMember(groupID, res[i].userName);
                                                         dialog_confirm.hide();
                                                         dialog.hide();
                                                     });
--- a/LEAF_Request_Portal/admin/templates/mod_groups.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_groups.tpl
@@ -413,11 +413,9 @@ function addAdmin(userID) {
 
 function removeAdmin(userID) {
     $.ajax({
-    	type: 'POST',
-        url: "ajaxIndex.php?a=remove_user",
-        data: {'userID': userID,
-        	   'groupID': 1,
-        	   'CSRFToken': '<!--{$CSRFToken}-->'},
+    	type: 'DELETE',
+        url: "../api/group/1/members/" + userID,
+        data: {'CSRFToken': '<!--{$CSRFToken}-->'},
         success: function(response) {
         	getMembers(1);
         },
--- a/LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
@@ -124,7 +124,7 @@ function getMembers(groupID = -1) {
     $.ajax({
         type: 'GET',
         url: '../api/system/updateService/' + groupID,
-        success: function() {
+        success: function(res) {
             $.ajax({
                 url: "../api/service/" + groupID + "/members",
                 dataType: "json",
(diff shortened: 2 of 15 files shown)


LEAF_Request_Portal/admin/templates/mod_groups.tpl
```diff
@@ -353,6 +353,18 @@ function pruneMember(groupID, userID) {
     });
 }
 
+function reactivateMember(groupID, userID) {
```
Pelentan, line 356: This sounds like a method that we want a return message for.

shaneodd (approved):

LEAF_Request_Portal/admin/templates/mod_groups.tpl
```diff
@@ -570,6 +586,16 @@ function getGroupList() {
...
                                                 });
+
+                                                $('#reActivateMember_' + counter).on('click', function () {
+                                                    dialog_confirm.setContent('Are you sure you want to Reactivate this member?');
+                                                    dialog_confirm.setSaveHandler(function () {
+                                                        reactivateMember(groupID, res[i].userName);
+                                                        dialog_confirm.hide();
+                                                        dialog.hide();
+                                                    });
+                                                    dialog_confirm.show();
```
Pelentan, line 597: This looks like what is calling the method above.  And it's decision loop is based solely on if the ajax call succeeds?

LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
```diff
@@ -219,15 +222,15 @@ function importUser(serviceID = 0, selectedUserName = '') {
     if (selectedUserName === '') {
         console.log('Invalid username');
     }
+
     $.ajax({
```
Pelentan, line 226: looks like this needs a fail-catch

LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
```diff
@@ -342,6 +418,31 @@ function initiateModal(serviceID = 0, serviceName = '') {
     }
 }
 
+function pruneMember(groupID, userID) {
```
Pelentan, line 421: Probably want to make sure these inputs are valid before we ping the db

LEAF_Request_Portal/sources/Group.php
```diff
@@ -388,37 +362,29 @@ public function importUser(string $userID, int $groupID, string $backupID): arra
      */
     public function deactivateMember($member, $groupID): void
```
Pelentan, line 363: Can we add type hinting here or will it break things?

LEAF_Request_Portal/sources/Service.php
```diff
@@ -144,19 +144,29 @@ public function removeSyncService(int $groupID): bool
...
 
+    /**
+     * @param int $groupID
+     * @param string $member
+     *
+     * @return bool
+     *
+     * Created at: 8/16/2023, 8:45:10 AM (America/New_York)
+     */
     public function addMember($groupID, $member)
```
Pelentan, line 155: type hinting and return type

LEAF_Request_Portal/sources/System.php
```diff
@@ -289,413 +264,153 @@ public function updateGroup(int $groupID): array
     /**
      * @param int $groupID
      *
-     * @return array
-     *
-     * Created at: 6/30/2023, 1:25:07 PM (America/New_York)
+     * @return string
      */
-    private function updateCatPrivs(int $groupID): array
```
Pelentan, line 296: Just back-checking.  Are we sure that nothing is going to be broken by this "old" function going away?

LEAF_Request_Portal/sources/System.php
```diff
@@ -289,413 +264,153 @@ public function updateGroup(int $groupID): array
...
+                    AND `groupID` = :groupID';
+
+            $this->db->prepared_query($sql, $vars);
+
+            foreach ($resEmp as $emp) {
+                if ($emp['userName'] != '') {
+                    $vars = array(':userID' => $emp['userName'],
+                        ':groupID' => $groupID,);
+
+                    $this->db->prepared_query('INSERT INTO users (userID, groupID, backupID)
```
Pelentan, line 325: should the tables and column names here have back-ticks?

LEAF_Request_Portal/sources/System.php
```diff
@@ -289,413 +264,153 @@ public function updateGroup(int $groupID): array
...
+                    $res = $this->db->prepared_query('SELECT * FROM users WHERE userID=:userID AND groupID=:groupID', $vars);
+                    if ($res[0]['active'] == 1) {
+                        $backups = $employee->getBackups($emp['empUID']);
+                        foreach ($backups as $backup) {
+                            $vars = array(':userID' => $backup['userName'],
+                                ':groupID' => $groupID,
+                                ':backupID' => $emp['userName'],);
+
+                            // Add backupID check for updates
+                            $this->db->prepared_query('INSERT INTO users (userID, groupID, backupID)
```
Pelentan, line 339: and here

LEAF_Request_Portal/sources/System.php
```diff
@@ -289,413 +264,153 @@ public function updateGroup(int $groupID): array
...
     {
-        $vars = array(':groupID' => $groupID);
-        $sql = 'DELETE
-                FROM `category_privs`
-                WHERE `groupID` = :groupID';
-
-        $return_value = $this->db->pdo_delete_query($sql, $vars);
-
-        return $return_value;
+        return $this->db->prepared_query('SELECT groupID as parentID,
```
Pelentan, line 354: and here

LEAF_Request_Portal/sources/System.php
```diff
@@ -289,413 +264,153 @@ public function updateGroup(int $groupID): array
...
-     * @param int $groupID
-     *
-     * @return array
+     * Get the current database version
      *
-     * Created at: 6/30/2023, 1:25:38 PM (America/New_York)
+     * @return string the current database version
      */
-    private function getCatPrivs(int $groupID): array
+    public function getDatabaseVersion()
```
Pelentan, line 367: Return value and same question about changing the name of the function from above.

LEAF_Request_Portal/sources/System.php
```diff
@@ -289,413 +264,153 @@ public function updateGroup(int $groupID): array
...
-                WHERE `category_privs`.`groupID` = :groupID
-                AND `groups`.`groupID` IS NULL';
+        $version = $this->db->prepared_query('SELECT data FROM settings WHERE setting = "dbVersion"', array());
+        if (count($version) > 0 && $version[0]['data'] !== null)
+        {
+            return $version[0]['data'];
+        }
 
-        $return_value = $this->db->pdo_select_query($sql, $vars);
+        return 'unknown';
```
Pelentan, line 375: Two different return types.  Is that required because of old code?

LEAF_Request_Portal/sources/System.php
```diff
@@ -289,413 +264,153 @@ public function updateGroup(int $groupID): array
...
+    }
 
-        return $return_value;
+    public function getGroups()
```
Pelentan, line 378: return type.

Pelentan (changes requested):

LEAF_Request_Portal/sources/System.php
jampaul3: I'm not sure, see my comment on your last comment below.

LEAF_Request_Portal/sources/System.php
jampaul3: This just got moved down the page, it's still here, just not in this location.

LEAF_Request_Portal/sources/System.php
jampaul3: I didn't change the name I rearranged. So git doesn't like that and thinks it changed.

LEAF_Request_Portal/sources/Service.php
jampaul3: I didn't add it here because I'm afraid it will break things. I did add the PHPDoc section to indicate what should be passed in.

LEAF_Request_Portal/sources/Group.php
jampaul3: I'm not sure if it will break anything or not. I did add the PHPDoc section so we know what should be passed to this method.

LEAF_Request_Portal/admin/templates/mod_groups.tpl
jampaul3: I'm not sure I understand what you are asking.

LEAF_Request_Portal/admin/templates/mod_groups.tpl
```diff
@@ -385,9 +397,8 @@ function toTitleCase(str) {
 function addAdmin(userID) {
```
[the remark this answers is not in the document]
jampaul3: Done

LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
jampaul3: I changed this to be like others on the page and just return if the data isn't valid.

LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
jampaul3: done

LEAF_Request_Portal/admin/templates/mod_svcChief.tpl
```diff
@@ -342,6 +418,31 @@ function initiateModal(serviceID = 0, serviceName = '') {
...
+        url: "../api/service/" + groupID + "/members/_" + userID + "/prune",
+        data: {'CSRFToken': '<!--{$CSRFToken}-->'},
+        fail: function(err) {
+            console.log(err);
+        },
+        cache: false
+    });
+}
+
+function reactivateMember(groupID, userID) {
```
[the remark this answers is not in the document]
jampaul3: done

LEAF_Request_Portal/sources/Service.php
```diff
@@ -227,7 +243,133 @@ public function importChief(int $serviceID, string $userID, string|null $backupI
...
+
+    /**
+     * @param int $serviceID
+     * @param string $userName
+     *
+     * @return array
+     *
+     * Created at: 8/16/2023, 8:46:50 AM (America/New_York)
+     */
+    public function pruneChief(int $serviceID, string $userName): array
```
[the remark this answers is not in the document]
jampaul3: failure return value is in the Db method

LEAF_Request_Portal/sources/Service.php
```diff
@@ -227,7 +243,133 @@ public function importChief(int $serviceID, string $userID, string|null $backupI
...
+
+    /**
+     * @param string $member
+     * @param int $serviceID
+     *
+     * @return array
+     *
+     * Created at: 8/16/2023, 8:47:05 AM (America/New_York)
+     */
+    public function reactivateChief(string $member, int $serviceID): array
```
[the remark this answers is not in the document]
jampaul3: same here

LEAF_Request_Portal/sources/Service.php
```diff
@@ -317,62 +459,59 @@ public function removeMember($groupID, $member)
      *
      * Created at: 9/14/2022, 11:33:53 AM (America/New_York)
      */
-    public function removeChief(int $serviceID, string $userID, string|null $backupID): array
+    public function removeChief(int $serviceID, string $userID, string $backupID = ""): array
```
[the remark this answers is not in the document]
jampaul3: Doesn't look like this method is used anymore.

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

raphaelfretter (approved):

Pelentan (approved):

maxrichardjr (approved):

AmandaSumner (approved):

maxrichardjr (approved):

maxrichardjr (approved):

PR #2109, opened 2023-08-29
LEAF 3457 - site designer homepage (review/comment only)

aerinkayne:
Multiple designs can be stored, and they can be deleted or copied.  Publication prompts for confirmation.

Adds Header design section with prosemirror WYSIWYG / markdown text options.

Adds nav with menu options, and various UI improvements.

Adds dedicated class, controller and table for design information.
Logs content change, draft/publish, create and delete events.

other
App and dialog organization improvements.
Removes (jQuery) ajax calls and replaces with fetch API calls.
User facing page is one page instead of homepage + search combo.


maxrichardjr (approved):

libs/js/vue-dest/site_designer/LEAF_Designer.js.LICENSE.txt
raphaelfretter, line 1: what is this license for?

raphaelfretter (changes requested):

libs/js/vue-dest/site_designer/LEAF_Designer.js.LICENSE.txt
shaneodd: This package is getting pulled in, I am guessing  probably a wysiwyg editor of some sort probably pulled it in. npm is a wild ride sometimes. https://github.com/facebook/regenerator

shaneodd (approved):

libs/js/vue-dest/site_designer/LEAF_Designer.js.LICENSE.txt
aerinkayne: I am fairly certain all of these are due to using async await in the code and how Babel compiles it.  I was trying to proactively avoid ajax calls and used the fetch API instead.  Using async await there also simplified a couple things but (I realized later) also resulted in the regenerator runtime licenses.  If there were only one, I wouldn't care but I think I was getting 1 per view plus some mystery chunk files.  When I pick this up again, I am just going to nest the JS and hopefully not get too many callback doom pyramids.

jampaul3 (approved):

aerinkayne:
Adding a comment for visibility / future sanity

Current master should be used for any src/common conflicts.  Site Designer dialogs should be updated to use dialogData the way it is used in the Form Editor App.
Use of async-await resulted in the 'generator-runtime' LICENSE files - replacing it with chaining should prevent these.

PR #2112, opened 2023-08-29
LEAF-3976 - Email sending on behalf of the requester not person taking action

shaneodd:


maxrichardjr (approved):

aerinkayne (approved):

LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -1053,26 +1053,36 @@ public function handleEvents(int $workflowID, int $stepID, string $actionType, s
...
+                    $requestRecords = $this->db->prepared_query($strSQL, $vars);
 
 
-                    $title = strlen($approvers[0]['title']) > 45 ? substr($approvers[0]['title'], 0, 42) . '...' : $approvers[0]['title'];
+                    $title = strlen($requestRecords[0]['title']) > 45 ? substr($requestRecords[0]['title'], 0, 42) . '...' : $requestRecords[0]['title'];
 
                     $email->addSmartyVariables(array(
                         "truncatedTitle" => $title,
-                        "fullTitle" => $approvers[0]['title'],
+                        "fullTitle" => $requestRecords[0]['title'],
```
Pelentan, line 1063: What sort of testing are you doing to ensure something usable is returned by these queries?  If the query returns no results or fails, this would generate all sorts of errors and not return what was expected to the user.

Pelentan (changes requested):

LEAF_Request_Portal/sources/FormWorkflow.php
shaneodd: I was going off of existing logic. Added in a check for the requestRecords and my pull for the person of last action. Would we want to raise an error if there was some information not set for email? That would be the only thing I could consider adding. Thoughts are "recordID in __file__ was not found" or something like that.

maxrichardjr (approved):

Pelentan (approved):

jrodcode (approved):

PR #2119, opened 2023-09-01
Test Automation Local

maxrichardjr:
This is where the local version of Test Automation is located (Test-Automation). Also, I created a new branch, Test-Automation-Local, which is the primary branch for local development


raphaelfretter (approved):

shaneodd:
Is the local test and the gfe tests going to conflict?

shaneodd (changes requested):
I want to make sure that these two folders for local and gfe are not going to cause problems with each other since they appear to be slight variations of each other from first glance.

jampaul3 (approved):

PR #2120, opened 2023-09-05
Leaf 3985 - move globals, autoloader and db

jampaul3:
As the title suggests this moves the globals.php, the Leaf_autoloader and the db.php to a new file structure.

This branch will require a docker compose down and up. This will open up the new app folder for use by the application and add some globals that will be needed.

The best way to test that this is actually working as expected is to delete or rename the files in the original location. We will want these files to remain in prod so that it won't break anything for backward compatability.


maxrichardjr (approved):

LEAF_Nexus/admin/index.php
```diff
@@ -15,8 +15,7 @@
 */
 error_reporting(E_ERROR);
 
-require_once '../globals.php';
-require_once LIB_PATH . '/loaders/Leaf_autoloader.php';
+require_once '/var/www/html/app/libs/loaders/Leaf_autoloader.php';
```
mgaoVA, line 18: Are absolute paths necessary? This can cause issues if the environment changes in the future.

mgaoVA (changes requested):

LEAF_Nexus/admin/index.php
jampaul3: It is for the meantime. Until we can move these files to  the app folder and make them all single source of truth there is no way to know the total number of folders deep we are. When they move we can then change these back to relative paths.

We're trying to do this move in small incremental steps to make it less overwhelming for the review process. Plus it makes testing easier. All these changes will be full regression testing but if we run into something it will be easier to debug if we move things over slowly. We are also moving things in such a way as to make sure everything is backward compatible.

Pelentan (changes requested):
Looks like way too many files got touched.  Including the docker files and a few others with only spacing changes?

LEAF_Nexus/admin/index.php
mgaoVA: The part I'm concerned with is:
```
/var/www/html
```

Would it be feasible to do something like:
```
require_once HTML_ROOT . '/app/libs/loaders/Leaf_autoloader.php';
```

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

scripts/scheduled-task-commands/refreshOrgchartEmployees.php
```diff
@@ -1,62 +1,23 @@
 <?php
```
shaneodd, line 1: This has a lot going on, we just pushed a change for this it looks like everything is there but just want to call that out.  The LEAFNexusAPI is another one that had some changes and look to be correct.

shaneodd (approved):

Pelentan (approved):

aerinkayne (approved):

pete-nerantzinis (approved):

maxrichardjr (approved):

jampaul3:
This was included in Leaf 4002 and was merged onto RC before this one

PR #2116, opened 2023-08-30
Test automation GFE

maxrichardjr:
Re-PRing all Test Automation Files

PR #2121, opened 2023-09-05
Exclude Java from CodeQL

mgaoVA:
Excludes Java from CodeQL scans since the code is only stored for archival purposes.

### DIFF
--- a/.github/codeql.yml
+++ b/.github/codeql.yml
@@ -1,0 +1,3 @@
+ excluded_languages:
+   - name: java
+     reason: "Java code is inactive and only used for archival purposes."


shaneodd:
Would this still hold  true now that we are bringing in the testing stuff? I know it is not running on the site however it would be there for people to test it. I remember having issues with fortify and code that was only there for the testing suite. Testing code I am referencing is here : https://github.com/department-of-veterans-affairs/LEAF/pull/2116/files

shaneodd (changes requested):
See my comment in the main conversation.

mgaoVA:
> Would this still hold true now that we are bringing in the testing stuff? I know it is not running on the site however it would be there for people to test it. I remember having issues with fortify and code that was only there for the testing suite. Testing code I am referencing is here : https://github.com/department-of-veterans-affairs/LEAF/pull/2116/files

Yes, test files are excluded because they're not part of the runtime environment in production.

Apparently there's a way to exclude paths, so I'll update this change to exclude paths instead of languages.

raphaelfretter:
Are we allowed to change the codeQL analysis yaml? If so then that should be modified to exclude java from the languages being scanned.

shaneodd (approved):

shaneodd:
> > Would this still hold true now that we are bringing in the testing stuff? I know it is not running on the site however it would be there for people to test it. I remember having issues with fortify and code that was only there for the testing suite. Testing code I am referencing is here : https://github.com/department-of-veterans-affairs/LEAF/pull/2116/files
> 
> Yes, test files are excluded because they're not part of the runtime environment in production.
> 
> Apparently there's a way to exclude paths, so I'll update this change to exclude paths instead of languages.

Aww okay either way I think would work then though dirs would allow us to show that its not code that is in production.

jampaul3 (approved):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

mgaoVA:
> Are we allowed to change the codeQL analysis yaml? If so then that should be modified to exclude java from the languages being scanned.

Yes we can, however in this case it's safer to exclude specific directories to prevent issues in the future.

Pelentan (approved):

raphaelfretter (approved):

shaneodd (approved):

AmandaSumner (approved):

PR #2124, opened 2023-09-06
LEAF-3995 - Optimize queries for actionable records 

mgaoVA:
This improves ./api/form/query performance when searching for actionable records and many parallel workflow requirements exist. For employee clearance processes with 20+ parallel requirements, this can provide a 20x improvement (Martinsburg) in load time.

Additionally this enables the masquerade=nonAdmin parameter in the Report Builder, allowing admins to view reports in a non-admin context.

- The original getActionable() was copied and renamed to a private function retrieveRecordDependencies()
- Dependencies on the old getActionable() were updated to the new function to minimize impact on functionality that depends on the original behavior.
- getActionable() was updated to align with its intended behavior, as described in the function comments: "add an isActionable parameter"
- getActionable() was optimized to focus on actionable records by updating the query to include a "filled=0" term, which represent requirements that have not already been acted on. The second optimization uses an internal index based on the recordID instead of dependencyID, since the intent of identifying actionable records is based on the recordID.
- Updates getEmpUIDByUserName() to include null as a return type hint

### Potential Impacts:
- Dependencies on ./api/form/query , such as the Report Builder

### Testing:

Report Builder -> Search for "Status IS Actionable"
- [x] Results as an admin are the same as before
- [x] Results as a non-admin are the same as before
- [x] As an admin, add "&masquerade=nonAdmin" to the URL. Results should appear as if you're not an admin.

PR #2126, opened 2023-09-08
LEAF-4004 - Inbox v2

mgaoVA:
Prerequisite: https://github.com/department-of-veterans-affairs/LEAF/pull/2124

This PR merge base has been set to PR#2124 to simplify review
- [ ] Update the merge base of this PR to the latest working RC branch after review

This implements a new standard Inbox, replaces the Combined Inbox (which it was mostly copied from), and server-side changes to support efficient loading of new functionality. The old Inbox will remain available from its original link. The key improvement is that the new inbox loads large datasets in smaller chunks, which provides better time-sharing on the server.

By default, the new Inbox organizes actionable results by type of form. An "Organize by Roles" option is available to provide a similar layout to the old Inbox. Admins have a "View as Admin" option to see all requests pending action.

### Summary
- ./api/form/query has a new "join" term: unfilledDependencies. This provides data needed for the Inbox.
- A new option in formQuery.js and intervalQueue.js enables the use of a DOM AbortController to cancel processing
- Form.php -> getIndicator() has a new parameter, forceReadOnly. It defaults to disabled and should only be enabled in safe contexts. This is necessary to keep overall performance equivalent between the new/old inboxes.
- Hot paths have been optimized in Form.php and FormWorkflow.php, reducing database queries.
- Workflow UI headers are more descriptive. For "requestor followup" and "designated group" requirements, the name of the requestor/group is included.

### Potential Impacts
- Dependencies on ./api/form/query
- Dependencies on ./api/formWorkflow/[recordID]/currentStep
- Issues with cache control could cause cached *.js files to prevent the inbox from working

### Testing
- [x] The new Inbox (must click on "View as Admin" and also "Organize by Roles") should have the same number of results as the old Inbox
- [x] Report Builder results as an admin are the same as before
- [x] Report Builder results as a non-admin are the same as before, including data fields within records
- [x] The Workflow section on the "a=printview" page works the same as before


LEAF_Request_Portal/sources/Form.php
```diff
@@ -1678,13 +1704,21 @@ public function hasWriteAccess($recordID, $categoryID = 0, $indicatorID = 0)
...
-            												AND writable=1', $vars);
+            $resCategoryPrivs = null;
+            $cacheHash = 'hasWriteAccess_catPrivs_'.$categoryID.$this->login->getUserID();
+            if(isset($this->cache[$cacheHash])) {
+                $resCategoryPrivs = $this->cache[$cacheHash];
+            }
+            else {
+                $vars = array(':categoryID' => $categoryID,
+                              ':userID' => $this->login->getUserID());
+                $resCategoryPrivs = $this->db->prepared_query('SELECT * FROM category_privs
```
Pelentan, line 1715: Does this need to be a "select *"?

LEAF_Request_Portal/sources/Form.php
```diff
@@ -495,9 +520,8 @@ public function getIndicator($indicatorID, $series, $recordID = null, $parseTemp
                 $form[$idx]['value'] = $this->fileToArray($data[0]['data']);
                 $form[$idx]['raw'] = $data[0]['data'];
             }
-
             // special handling for org chart data types
-            if ($data[0]['format'] == 'orgchart_employee'
+            else if ($data[0]['format'] == 'orgchart_employee'
```
Pelentan, line 524: Would like these to be fed through toLower().  We're working on future-proofing things.

LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -218,12 +242,13 @@ public function getActionable(object $form, array $records): array {
...
      * @param object $form instance of Form
      * @param array $records result set from a query on db:records. Requires 'recordID'.
+     * @param bool optional Only select unfilled dependencies
      * @return array $records list of dependencies * records
      */
-    private function retrieveRecordDependencies(object $form, array $records): array {
+    public function getRecordsDependencyData(object $form, array $records, bool $selectUnfilled = false): array {
         $numRecords = count($records);
         if ($numRecords == 0) {
             return $records;
```
Pelentan, line 254: Would like this method to get modernized a bit more.  There should be only one return.  And we're moving to a json/array format where there are at least two parts.  1 telling the success or fail with data and 1 with the returned data.  How much effort would it be to update this method and its callers?

maxrichardjr (approved):

maxrichardjr (approved):

LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
```diff
@@ -0,0 +1,926 @@
...
+
+    function buildWorkflowCategoryCache(site) {
+        return $.ajax({
+            type: 'GET',
+            url: site.url + 'api/workflow/categories?x-filterData=categoryID',
+            success: function(res) {
+                res.forEach(w => {
+                    dataWorkflowCategories[w.categoryID] = 1;
+                });
+            }
```
Pelentan, line 677: Need a fail/catch

Pelentan (changes requested):

LEAF_Request_Portal/sources/Form.php
mgaoVA: Looks like it'd be safe to change this to a COUNT(*)

LEAF_Request_Portal/sources/Form.php
mgaoVA: These should be left as-is. If the format is coming in with a different case, it should be considered wrong.

LEAF_Request_Portal/sources/FormWorkflow.php
mgaoVA: I think we should formalize a standard response class first, as part of a separate modernization task. Right now the response format is being implemented ad-hoc and can potentially be inconsistent.

LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
mgaoVA: Updated

raphaelfretter:
Unrelated to actual code review: This would be a good example to use for constructing a PR template in documentation

maxrichardjr (approved):

maxrichardjr (approved):

jrodcode (approved):

Pelentan (approved):

AmandaSumner (approved):

PR #2150, opened 2023-09-21
LEAF - 4035 - New Account Updater Search Bug

jrodcode:
This issue was brought up by a user saying that he wasn't able to pull disabled users by email or alias. The issue has been resolved and the way to test this by:

1. Going to the admin panel and click on New Account Updater
2. Search for the user that is disabled using the email or alias.

This can also be tested in the Nexus org chart, by searching for email or alias.

### DIFF
--- a/LEAF_Nexus/sources/Employee.php
+++ b/LEAF_Nexus/sources/Employee.php
@@ -1005,6 +1005,7 @@ class Employee extends Data
         $sql = "SELECT *
                 FROM {$this->tableName}
                 WHERE firstName LIKE :firstName
+                AND deleted = 0
                 ORDER BY {$this->sortBy} {$this->sortDir}
                 {$this->limit}";
 
@@ -1015,6 +1016,7 @@ class Employee extends Data
             $sql = "SELECT *
                     FROM {$this->tableName}
                     WHERE phoneticFirstName LIKE :firstName
+                    AND deleted = 0
                     ORDER BY {$this->sortBy} {$this->sortDir}
                     {$this->limit}";
 
@@ -1091,13 +1093,15 @@ class Employee extends Data
         return $result;
     }
 
-    public function lookupEmail($email)
+    public function lookupEmail(string $email, bool $searchDisabledEmail = false): array
     {
+        $accountStatus = $searchDisabledEmail ? '' : ' AND deleted = 0';
         $sql = "SELECT *
                 FROM {$this->dataTable}
                 LEFT JOIN {$this->tableName} USING (empUID)
                 WHERE indicatorID = 6
-                    AND data = :email
+                AND data = :email
+                {$accountStatus}
                 {$this->limit}";
 
         $vars = array(':email' => $email);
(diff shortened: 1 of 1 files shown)


LEAF_Nexus/sources/Employee.php
```diff
@@ -887,7 +887,7 @@ public function getPositions(int|string $empUID): int|array|bool
      * @param string $login
      * @param bool $searchDeleted
      */
-    public function lookupLogin($login, bool $searchDeleted = false): array
+    public function lookupLogin($login, bool $searchDeleted = true): array
```
Pelentan, line 890: This is going to need some good regression testing.  Changing a default to the opposite can have a slew of unintended consequences.

raphaelfretter: This should not be changed. The endpoint call on the account updater page should be updated to pull deleted users. That will ensure the rest of the site works as it did before.

raphaelfretter (changes requested):

maxrichardjr (approved):

shaneodd (approved):

mgaoVA (changes requested):
Normal employee searches should never include data flagged as disabled or deleted. This can cause confusion if someone inadvertently selects an old account.

We need to retain the explicit search syntax in the short term: 
```
username_disabled:
```
People have been instructed to use this, and we need time to migrate them off.


A new prefix should be added: 
```
disabled:
```
This prefix should have similar search UX as a normal search.

Another prefix should be added:
```
disabled.username:
```
This is the same as username_disabled.

maxrichardjr (approved):

aerinkayne (approved):

jampaul3 (approved):

LEAF_Nexus/sources/Employee.php
```diff
@@ -1098,7 +1098,6 @@ public function lookupEmail($email)
                 LEFT JOIN {$this->tableName} USING (empUID)
                 WHERE indicatorID = 6
                     AND data = :email
-                    AND deleted = 0
```
mgaoVA, line 1101: This should retain the deleted = 0 check. The default behavior must exclude inactive "deleted" employees.

If it's necessary to search for inactive employees by email, it needs to be made as an option.

mgaoVA (changes requested):

maxrichardjr (approved):

LEAF_Nexus/sources/Employee.php
```diff
@@ -1337,15 +1340,28 @@ public function search($input, $indicatorID = '')
...
+                } else if (substr(strtolower($input), 0, 15) === 'disabled.email:'){
+                    $input = str_replace('disabled.email:', '', strtolower($input));
+                    $searchResult = $this->lookupEmail($input, true);
+                } else {
+                    $searchResult = $this->lookupEmail($input);
+                }
 
                 break;
-            // Format: Loginname
+            // Format: Loginname && Will search all active and disabled accounts
```
mgaoVA, line 1358: Update comment, it doesn't search disabled accounts

mgaoVA (changes requested):

maxrichardjr (approved):

Pelentan (approved):

AmandaSumner (approved):

maxrichardjr (approved):

maxrichardjr (approved):

PR #2154, opened 2023-09-22
LEAF 4038 timeline explorer infinite loop

aerinkayne:
Removes custom batching from timeline explorer query to avoid potential infinite loop.

Adds a check for the version property.

Adjusts some spacing.

### DIFF
--- a/LEAF_Request_Portal/templates/reports/LEAF_Timeline_Explorer.tpl
+++ b/LEAF_Request_Portal/templates/reports/LEAF_Timeline_Explorer.tpl
@@ -356,7 +356,6 @@ function loadData(site, categoryID) {
     let siteURL = getSiteURL(site);
     let promise = new Promise((resolve, reject) => {
         let query = new LeafFormQuery();
-        let batchLimit = 5000;
         let index = 0;
 
         query.addTerm('dateSubmitted', '>=', queryFirstDateSubmitted);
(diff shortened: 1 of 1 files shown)


jampaul3 (approved):

shaneodd (approved):

maxrichardjr (approved):

mgaoVA (changes requested):
The original issue may need a closer look. It should not be necessary to check batch sizes or set limits when using formQuery.

Note that formQuery may appear as if it's in an infinite loop, however it's probably not actually stuck.

In order to provide accurate information, the database must scan all records that match the initial query. Since formQuery splits the query via SQL LIMIT, it's necessary to iterate through all rows. But how does the front-end know when to stop?

The signaling is handled by a custom HTTP header: [leaf-query: continue](https://github.com/department-of-veterans-affairs/LEAF/blob/master/LEAF_Request_Portal/js/formQuery.js#L254). When set, this tells the frontend there are still more rows to be scanned.

mgaoVA (approved):

maxrichardjr (approved):

jampaul3 (approved):

Pelentan (approved):

AmandaSumner (approved):

raphaelfretter (approved):

PR #2159, opened 2023-09-28
Leaf 4042 - remove double unlink in mailer

jampaul3:
This just places an if statement around the last two unlinks checking to make sure the file exists before trying to unlink it.

### DIFF
--- a/LEAF_Request_Portal/mailer/mailer.php
+++ b/LEAF_Request_Portal/mailer/mailer.php
@@ -56,15 +56,13 @@ foreach ($queue as $item)
     if (strlen($item) == 40)
     {
         // attempt to resend email if its 5 minutes old
-        if (time() - filemtime($folder . $item) >= 300)
+        if (file_exists($folder . $item) && time() - filemtime($folder . $item) >= 300)
         {
             $email = unserialize(file_get_contents($folder . $item));
             if (strlen(trim($email['recipient'])) == 0)
             {
                 // delete invalid cache
-                if (file_exists($folder . $item)) {
-                    unlink($folder . $item);
-                }
+                unlink($folder . $item);
 
                 trigger_error('Mail no recipient: ' . $email['subject']);
             }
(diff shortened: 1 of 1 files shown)


aerinkayne (approved):

LEAF_Request_Portal/mailer/mailer.php
```diff
@@ -62,15 +62,20 @@
...
             }
             else
             {
                 touch($folder . $item);    // reset timer
                 if (mail($email['recipient'], $email['subject'], $email['body'], $email['headers']))
                 {
-                    unlink($folder . $item);
+                    if (file_exists($folder . $item)) {
+                        unlink($folder . $item);
+                    }
```
shaneodd, line 78: I was seeing a couple different things going on here. The file_esixts might need to go on the outside of the if(mail and touch 70/73 above. Similar to line 37, 

I was trying to figure out why the touch was failing to find the file which might be a condition where two people are triggering the email sending close to the same time. The error was happening a lot in the morning which could lend to people running through reminding folks or moving something through the workflow. This second part is the thing that I am uncertain of.

shaneodd (changes requested):

LEAF_Request_Portal/mailer/mailer.php
jampaul3: This is more for Pete maybe, if we move it outside the if statement then the trigger error won't get hit. I know Pete you said this could use some looking over, should we back this out and take a deeper dive into it or do we want to get this out to remove those errors? I can move the if if that seems better.

maxrichardjr (approved):

AmandaSumner (approved):

raphaelfretter (approved):

jrodcode (approved):

LEAF_Request_Portal/mailer/mailer.php
pete-nerantzinis: Yeah the error wont get triggered but that just means its already sent or gone. So there shouldn't be an error.

pete-nerantzinis (approved):

Pelentan (approved):

shaneodd (approved):

maxrichardjr (approved):

PR #2163, opened 2023-10-03
Leaf 4010 - move img jsapi smarty

jampaul3:
**Summary:**
This moves the rest of the files and folders currently in html/libs to the one source of truth at html/app/libs and html/app/Leaf.
In this release html/libs/img, html/libs/jsapi and html/libs/smarty are all moved to html/app/libs. jsapi was moved to html/app/libs/js to keep things consistent. jsapi had two folders with subfolders and/or files, these folders were just copide to the js folder.

**Potential impact:**
If a reference to the original was completely missed and it is still referring to the old location nothing should happen, the old files are still in their original location and accessible. However, if a reference were changed and it isn't quite right we could end up with broken references and 404 errors.

**Testing:**
A full regression testing is needed for this release. We are looking for broken references where files could not be found and therefore not loaded. 

Special note, on local and staging the html/libs file can be removed to ensure that all references have been changed and are working. DO NOT do this on prod as doing so will break every site until the full push is complete.

Pay special attention to the following pages as they have references to the old img and jsapi (all of which should be changed to the new location):
Portal/report.php
Portal/index.php - import_data, reports, printview
Portal/admin/index.php - workflow, form, import_data
coach_roster.tpl

loading spinner on the Access User Groups


maxrichardjr (approved):

maxrichardjr (approved):

jampaul3:
The coping of files and the changing of files have been separated into separate PR's

PR #2164, opened 2023-10-03
LEAF 4051 - Remove cache workaround, update form*.js files in /app/libs/js/LEAF

mgaoVA:
This saves some bandwidth by removing an anti-cache workaround in formSearch.js. This is safe because cache control is configured correctly for API endpoints in Prod.

Additionally I noticed the new /app/libs/js/LEAF folder contains outdated JS assets and has started to diverge. I've overwritten them with updated versions from LEAF_Request_Portal/js. When the file consolidation effort is complete, these files should only be present in one folder.

### Potential Impact
No impact expected if the /app/libs/js/LEAF files are currently unused.

### Testing
- [x] Search box on the homepage works normally

### DIFF
--- a/app/libs/js/LEAF/formGrid.js
+++ b/app/libs/js/LEAF/formGrid.js
@@ -14,14 +14,15 @@ var LeafFormGrid = function (containerID, options) {
   var isDataLoaded = false;
   var defaultLimit = 50;
   var currLimit = 50;
-  var headerColor = "#d1dfff";
   var dataBlob = {}; // if data needs to be passed in
   var postProcessDataFunc = null;
   var preRenderFunc = null;
   var postRenderFunc = null;
+  let postSortRequestFunc = null;
   var rootURL = "";
   var isRenderingVirtualHeader = true;
   var isRenderingBody = false;
+  let renderHistory = {}; // index of rendered recordIDs
 
   $("#" + containerID).html(
     '<div id="' +
@@ -47,8 +48,6 @@ var LeafFormGrid = function (containerID, options) {
       'tfoot"></tfoot></table>'
   );
 
-  $("#" + prefixID + "thead").css({ "background-color": headerColor });
-
   if (options == undefined) {
     form = new LeafForm(prefixID + "form");
   }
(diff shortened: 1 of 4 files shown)


aerinkayne (approved):

raphaelfretter (approved):

jampaul3 (approved):

maxrichardjr (approved):

shaneodd (changes requested):

app/libs/js/LEAF/formGrid.js
```diff
@@ -61,6 +60,107 @@ var LeafFormGrid = function (containerID, options) {
...
 
+  /**
+   * @param values (required) object of cells and names to generate grid
+   * @param showScriptTags (default false) whether to display script tags
+   * @memberOf LeafFormGrid
+   * Returns copy of values with cells property html entities decoded
+   */
+  function decodeCellHTMLEntities(values, showScriptTags = false) {
+    let gridInfo = { ...values };
+    if (gridInfo?.cells) {
```
Pelentan, line 71: gridInfo.cells is not a boolean.  We are trying to make all our "if"s on actual boolean values to keep things future proof.

app/libs/js/LEAF/formGrid.js
```diff
@@ -61,6 +60,107 @@ var LeafFormGrid = function (containerID, options) {
...
+          v = v.replaceAll(">", "&gt;");
+          let elDiv = document.createElement("div");
+          elDiv.innerHTML = v;
+          let text = elDiv.innerText;
+          if (showScriptTags !== true)
+            text = text.replaceAll(
+              /(<script[\s\S]*?>)|(<\/script[\s\S]*?>)/gi,
+              ""
+            );
+          return text;
```
Pelentan, line 85: This almost caught me.  Looked like you had two different returns for the same function.  Then I remembered you had called the map function above.  ;-)

app/libs/js/LEAF/formGrid.js
```diff
@@ -61,6 +60,107 @@ var LeafFormGrid = function (containerID, options) {
...
+    //finds and displays column names
+    for (let i = 0; i < columns; i++) {
+      tDelim = i === columns - 1 ? "" : delim;
+      gridHeadBuffer +=
+        '<td style="width: 100px;">' + values.format[i].name + tDelim + "</td>";
+      columnOrder.push(values.format[i].id);
+    }
+
+    //populates table
+    for (var i = 0; i < rows; i++) {
```
Pelentan, line 125: This should be a "let i".  Code cleanup since you're in here already.

app/libs/js/LEAF/formGrid.js
```diff
@@ -61,6 +60,107 @@ var LeafFormGrid = function (containerID, options) {
...
+      var gridRow = "<tr>";
+      var rowBuffer = [];
+
+      //makes array of cells
+      for (var j = 0; j < columns; j++) {
```
Pelentan, line 130: Here too.

app/libs/js/LEAF/formGrid.js
```diff
@@ -61,6 +60,107 @@ var LeafFormGrid = function (containerID, options) {
...
+        rowBuffer.push('<td style="width:100px"></td>');
+      }
+
+      //for all values with matching column id, replaces cell with value
+      for (var j = 0; j < values.columns.length; j++) {
```
Pelentan, line 135: And here please.

app/libs/js/LEAF/formGrid.js
```diff
@@ -61,6 +60,107 @@ var LeafFormGrid = function (containerID, options) {
...
+  function printTableReportBuilder(values, columnValues) {
+    // remove unused columns
+    values = decodeCellHTMLEntities(values);
+    if (columnValues !== null && columnValues !== undefined) {
+      values.format = values.format.filter(function (value) {
+        return columnValues.includes(value.id);
+      });
+    }
+
+    var gridBodyBuffer = "";
```
Pelentan, line 107: I'm thinking all these should be "let" as well.  Looks like many of them are used inside the function only.

app/libs/js/LEAF/formGrid.js
```diff
@@ -119,44 +230,26 @@ var LeafFormGrid = function (containerID, options) {
...
+        '<th tabindex="0" id="' +
         prefixID +
         'header_UID" style="text-align: center">UID</th>';
-      virtualHeader += '<th style="text-align: center">UID</th>';
+      virtualHeader +=
+        '<th id="Vheader_UID" style="text-align: center">UID</th>';
     }
     $("#" + prefixID + "thead").html(temp);
 
     if (showIndex) {
```
Pelentan, line 241: Why are there two if-checks for showIndex?  One right after the other.  Couldn't they be combined?

app/libs/js/LEAF/formGrid.js
```diff
@@ -813,52 +998,88 @@ var LeafFormGrid = function (containerID, options) {
...
+      $("#" + prefixID + "thead>tr>th>span").each(function (idx, val) {
+        $(val).html("");
+      });
       $("#" + prefixID + "thead>tr>th").each(function (idx, val) {
         headers.push($(val).text().trim());
       });
+      output.push(headers); //first row will be headers
 
-      var line = {};
+      var line = [];
```
Pelentan, line 1010: Can this be a let?

app/libs/js/LEAF/formGrid.js
```diff
@@ -813,52 +998,88 @@ var LeafFormGrid = function (containerID, options) {
...
-        type: "hidden",
-        name: "input",
-        value: JSON.stringify(output),
+      document
+        .querySelectorAll("#" + prefixID + "tbody>tr>td")
+        .forEach(function (val) {
+          var foundScripts = val.querySelectorAll("script");
+
+          for (var tIdx = 0; tIdx < foundScripts.length; tIdx++) {
+            foundScripts[tIdx].parentNode.removeChild(foundScripts[tIdx]);
```
Pelentan, line 1020: This might be either.

app/libs/js/LEAF/formSearch.js
```diff
@@ -440,6 +447,10 @@ var LeafFormSearch = function (containerID) {
                             $("#" + prefixID + "widgetMat_" + widgetID).val(
                                 selection
                             );
+                            //uses id.  report builder/search will not take userName:<username>
+                            $("#" + empSel.prefixID + "input").val(
+                                "#" + empSel.selection
+                            );
                         }
                     });
```
Pelentan, line 455: I'm not seeing a fail/catch on this ajax call.

Pelentan (changes requested):

app/libs/js/LEAF/formGrid.js
mgaoVA: @Pelentan For all of these I'd rather scope code cleanup in a different story. This PR is a literal copy paste from LEAF_Request_Portal/js to /app/libs/js/LEAF, and meant to be a preventative measure to realign files and avoid someone adding new things on the wrong file. I noticed that some work was being done on the unused files, which would definitely cause problems later.


Ideally this would be a symbolic link (until the consolidation "single source of truth" is complete) instead of copy/pasting files, but I didn't look into how this would impact Windows/Linux envs.

Pelentan (approved):

AmandaSumner (approved):

PR #2165, opened 2023-10-04
LEAF 4050 - Remove unused SQL JOIN

mgaoVA:
This removes an unused JOIN, which caused poor performance when the following conditions are met: 1) record submission or workflow actions 2) there are many data fields on the form 3) an email notification is sent to the next approver. Removing the JOIN avoids around [# data fields] * 3 queries in the subsequent loop.

## Potential Impact
None because the query doesn't select any data from the joined table.

## Testing
- [x] Email "notify next approver" works normally


jampaul3 (approved):

maxrichardjr (approved):

aerinkayne (approved):

shaneodd (approved):
This reduces the amount of redundant data that comes back. My test had one row vs twenty one rows.

raphaelfretter (approved):

AmandaSumner (approved):

PR #2167, opened 2023-10-04
Leaf 4010-2 - update references to copied files

jampaul3:
**Summary:**
This is a follow up to Leaf 4010.1. It updates any references to the old location to the new location. It also refactors the globals file to construct the PORTAL_PATH. 

**Potential impact:**
If a reference to the original was completely missed and it is still referring to the old location nothing should happen, the old files are still in their original location and accessible. However, if a reference were changed and it isn't quite right we could end up with broken references and 404 errors.

Where the globals are concerned, if a portal has the exact name as any of the folders we have it will break the site. A different way to construct the portal path would need to be figured out.

**Testing:**
A full regression testing is needed for this release. We are looking for broken references where files could not be found and therefore not loaded.

Special note, on local and staging the html/libs file can be removed/renamed to ensure that all references have been changed and are working. DO NOT do this on prod as doing so will break every site until the full push is complete.

Pay special attention to the following pages as they have references to the old img and jsapi (all of which should be changed to the new location):
Portal/report.php
Portal/index.php - import_data, reports, printview
Portal/admin/index.php - workflow, form, import_data
coach_roster.tpl

loading spinner on the Access User Groups

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_groups.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_groups.tpl
@@ -78,7 +78,7 @@
     </main>
 </div>
 <!--Loading Modal-->
-<!--{include file="../../../app/libs/smarty/loading_spinner.tpl" title='User Groups'}-->
+<!--{include file="<!--{$app_libs}-->/smarty/loading_spinner.tpl" title='User Groups'}-->
 
 <!--{include file="site_elements/generic_xhrDialog.tpl"}-->
 <!--{include file="site_elements/import_dialog.tpl"}-->
--- a/LEAF_Request_Portal/admin/templates/mod_groups.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_groups.tpl
@@ -78,7 +78,7 @@
     </main>
 </div>
 <!--Loading Modal-->
-<!--{include file="../../../libs/smarty/loading_spinner.tpl" title='User Groups'}-->
+<!--{include file="../../../app/libs/smarty/loading_spinner.tpl" title='User Groups'}-->
 
 <!--{include file="site_elements/generic_xhrDialog.tpl"}-->
 <!--{include file="site_elements/import_dialog.tpl"}-->
--- a/app/Leaf/Site.php
+++ b/app/Leaf/Site.php
@@ -1,0 +1,92 @@
+<?php
+/*
+ * As a work of the United States government, this project is in the public domain within the United States.
+ */
+
+namespace App\Leaf;
+
+use App\Leaf\Db;
+
+class Site
+{
+    private $db;
+
+    private $match;
+
+    private $portal_path;
+
+    private $site_paths;
+
+    public $error = false;
+
+    /**
+     * @param Db $db
+     *
+     * Created at: 9/5/2023, 10:56:00 AM (America/New_York)
+     */
+    public function __construct(Db $db, string $path)
+    {
+        $this->db = $db;
+
+        preg_match('(\/.+\/)', $path, $match);
+        $this->match = rtrim(str_replace('/var/www/html', '', $match[0]), '/');
+        // the only time that more than one folder gets removed is here so going to strip it here rather than wait.
+        $this->match = str_replace('/sources/../mailer', '', $this->match);
+
+        $portal_path = $this->checkPath();
+
+        if ($portal_path['status']['code'] == 2 && !empty($portal_path['data'])) {
+            $this->portal_path = $this->match;
+            $this->site_paths = $portal_path['data'][0];
+        } else {
+            // the original url does not produce a site, need to extract the end of the url and try again.
+            $this->stripLast();
+            $portal_path = $this->checkPath();
+
+            if ($portal_path['status']['code'] == 2 && !empty($portal_path['data'])) {
+                $this->portal_path = $this->match;
+                $this->site_paths = $portal_path['data'][0];
+            } else {
+                $this->error = true;
+            }
+        }
+    }
+
+    public function getPortalPath(): string
+    {
+        return $this->portal_path;
+    }
+
+    public function getSitePath(): array
+    {
+        return $this->site_paths;
+    }
+
+    private function stripLast(): void
+    {
+        $path_array = explode('/', $this->match);
+
+        $path = '';
+
+        // starting with 1 because the first element in the array will be blank
+        // need to subtract 1 from the count because of the blank first element
+        for ($i = 1; $i < count($path_array) - 1; $i++) {
+            $path .= '/' . $path_array[$i];
+        }
+
+        $this->match = $path;
+    }
+
+    private function checkPath(): array
+    {
+        $vars = array(':site_path' => $this->match);
+        $sql = 'SELECT `site_path`, `site_uploads`, `portal_database`, `orgchart_path`,
+                    `orgchart_database`
+                FROM `sites`
+                WHERE `site_path` = BINARY :site_path';
+
+        $return_value = $this->db->pdo_select_query($sql, $vars);
+
+        return $return_value;
+    }
+}
--- a/app/Leaf/Site.php
+++ b/app/Leaf/Site.php
@@ -65,12 +65,12 @@ class Site
     private function stripLast(): void
     {
         $path_array = explode('/', $this->match);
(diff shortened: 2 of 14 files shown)


app/libs/smarty/plugins/modifier.sanitizeRichtext.php
```diff
@@ -24,7 +24,9 @@
  * @return string
  */
 
- require_once getenv('APP_LIBS_PATH') . '/loaders/Leaf_autoloader.php';
+ if(!class_exists('App\Leaf\XSSHelpers')){
+    include_once __DIR__ . '/../../../Leaf/XSSHelpers.php';
```
shaneodd, line 28: Looks like we have a few different iterations of this, qrcode in nexus vs portal are different to eachother and what is here for loading xsshelpers.

app/libs/smarty/plugins/modifier.sanitize.php
```diff
@@ -23,8 +23,8 @@
  *
  * @return string
  */
-if(!class_exists('Leaf\XSSHelpers')){
-    include_once __DIR__ . '/../../php-commons/XSSHelpers.php';
+if(!class_exists('App\Leaf\XSSHelpers')){
+    include_once __DIR__ . '/../../../Leaf/XSSHelpers.php';
```
shaneodd, line 27: same as below

shaneodd (changes requested):

app/libs/globals.php
```diff
@@ -22,20 +22,19 @@
...
-$path = str_replace('/auth_cookie/', '/', $path);
-$path = str_replace('/auth_domain/', '/', $path);
-$path = str_replace('/auth_token/', '/', $path);
-$path = str_replace('/dynicons/', '/', $path);
-$path = str_replace('/login/', '/', $path);
-$path = str_replace('/qrcode/', '/', $path);
-$path = str_replace('/scripts/', '/', $path);
-$path = str_replace('sources/../mailer', '', $path);
-$path = str_replace('/utils/', '/', $path);
+$folders = array('admin', 'api', 'auth.cookie', 'auth_domain', 'auth_token', 'dynicons', 'login', 'qrcode', 'scripts', 'utils', 'sources', '..', 'mailer', 'var', 'www', 'html', '');
```
Pelentan, line 25: Please add some comments here to explain to future devs what exactly this array is for.  And I know it's more for show, but put each on a separate line.  Others may want to add comments per path.  Let this be one of the new methods we can point to as "this is how to do it."

To include a "why" we're doing it this way.

app/libs/loaders/Leaf_autoloader.php
```diff
@@ -35,30 +35,29 @@
 //error_log(print_r($site_paths, true));
 $site_paths = $site_paths['data'][0];
 
-if (is_dir($curr_dir . '/libs/php-commons')) {
-    //Commenting this out for testing purposes, this will be uncommented before going to prod.
+/** Here down is old loader stuff, will be deprecated as we go along getting everything into a single source of code. */
```
Pelentan, line 38: Good.  But can you put some comment brackets around the "old" code that will be deprecated?

Pelentan (approved):

app/libs/smarty/plugins/modifier.sanitizeRichtext.php
raphaelfretter: Do we want a separate PR for the Nexus?

maxrichardjr (approved):

raphaelfretter (approved):

app/libs/smarty/plugins/modifier.sanitizeRichtext.php
jampaul3: I fixed the qrcode differences. Here the only thing needed is the XSSHelper so that is all we are loading. In the other cases we are loading globals and others because they are all needed.

@raphaelfretter I was only told to separate moved/copied files from modified files.

LEAF_Request_Portal/admin/templates/mod_groups.tpl
```diff
@@ -78,7 +78,7 @@
     </main>
 </div>
 <!--Loading Modal-->
-<!--{include file="../../../libs/smarty/loading_spinner.tpl" title='User Groups'}-->
+<!--{include file="../../../app/libs/smarty/loading_spinner.tpl" title='User Groups'}-->
```
pete-nerantzinis, line 81: Dont use ../../../app when trying to reference somthing on the base just use the base /app because this wont work for all portals also app folder isnt the relativity for all portals/orgcharts so make sure you are using correct relativiness.

Maybe passing an APP_LIBS_PATH here would work better just be sure

app/libs/globals.php
```diff
@@ -20,22 +20,27 @@
...
+    needs to be striped from the url
+
+    i.e. /Academy/Demo1/admin
+    I decided it best to put this into a class and have the class deal with it to keep this file clean
+*/
+
+require_once getenv('APP_PATH') . '/Leaf/Db.php';
+require_once getenv('APP_PATH') . '/Leaf/Site.php';
+
+$file_paths_db = new App\Leaf\Db(DIRECTORY_HOST, DIRECTORY_USER, DIRECTORY_PASS, 'national_leaf_launchpad');
```
pete-nerantzinis, line 35: I'm curious if putting a db call in the globals will cause another increase in connections like we faced with the last time we put a db call into one of our config/global. Depends how often this will be called I guess or maybe we can cache it so its not constantly calling it, more looking at the cache.

pete-nerantzinis (changes requested):

maxrichardjr (approved):

LEAF_Request_Portal/admin/templates/mod_groups.tpl
jampaul3: Yeah, it had the .. reference there so I didn't change it because it appears to be working the way it is. But I agree it would be better to have it referenced absolutely. I made the necessary change, appears to be working still locally.

maxrichardjr (approved):

app/Leaf/Site.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function getSitePath(): array
+    {
+        return $this->site_paths;
+    }
+
+    private function stripLast(): void
+    {
+        $path_array = explode('/', $this->match);
```
Pelentan, line 67: The below complicates things a bit much.  Might be better to do this:

$path_array = array_shift(explode('/', $this->match));

Most skip comments unless something forces them to read.

app/libs/globals_plus.php
```diff
@@ -0,0 +1,30 @@
...
+
+require_once getenv('APP_PATH') . '/Leaf/Db.php';
+require_once getenv('APP_PATH') . '/Leaf/Site.php';
+
+$file_paths_db = new App\Leaf\Db(DIRECTORY_HOST, DIRECTORY_USER, DIRECTORY_PASS, 'national_leaf_launchpad');
+
+$site = new App\Leaf\Site($file_paths_db, $_SERVER['SCRIPT_FILENAME']);
+
+if ($site->error) {
+    // redirect to 404 not found
```
Pelentan, line 24: Did you plan to this now, or later?

app/libs/loaders/Leaf_autoloader.php
```diff
@@ -20,47 +20,48 @@
 
 $file_paths_db = new Db(DIRECTORY_HOST, DIRECTORY_USER, DIRECTORY_PASS, 'national_leaf_launchpad');
 
-if (substr(PORTAL_PATH, 0, 1) !== '/') {
-    $my_path = '/' . PORTAL_PATH;
+$site = new App\Leaf\Site($file_paths_db, $_SERVER['SCRIPT_FILENAME']);
+
+if ($site->error) {
+    // redirect to 404 not found
```
Pelentan, line 26: Same here?  At the very least we need an error written to a log.

Pelentan (changes requested):

maxrichardjr (approved):

Pelentan (approved):

aerinkayne (approved):

maxrichardjr (approved):

AmandaSumner (approved):

shaneodd (approved):

PR #2176, opened 2023-10-06
LEAF 4060 - Optimize requestor followup (1 of 3)

mgaoVA:
Similar to PR#2155, this simplifies and removes redundant queries to reduce DB roundtrips. The same logic in PR#2155 used to check employee backups was also implemented here.

### Potential Impact
Features that depend on requestor follow-up and employee backups

### Testing
Preconditions:
- Form is configured with "need to know" mode enabled
- Form's workflow has a requestor followup step
- A record has been created with the above preconditions, and is on the requestor followup step
- [x] As a non-admin, the requestor can see the action buttons when it's pending their action
- [x] As a non-admin, the current user cannot see the action buttons when it's not pending their action, and they are not the requestor
- [x] As a non-admin, a backup of the requestor can see the action buttons when it's pending the requestor's action

### DIFF
--- a/LEAF_Request_Portal/sources/Form.php
+++ b/LEAF_Request_Portal/sources/Form.php
@@ -1825,6 +1825,7 @@ class Form
      * Checks if the current user has access to a particular dependency
      * @param int $dependencyID
      * @param array $details - Associative Array containing dependency-specific details, eg: $details['groupID']
+     *                         Must contain DB reference to records.userID
      * @return boolean
      */
     public function hasDependencyAccess($dependencyID, $details)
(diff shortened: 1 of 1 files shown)


jampaul3 (approved):

shaneodd (approved):

Pelentan (approved):

aerinkayne (approved):

raphaelfretter (approved):

AmandaSumner (approved):

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1968,26 +1967,35 @@ public function getEmpUID($userName){
...
 
-            return false;
+        $this->cache['checkIfBackup'] = [];
+        foreach ($backupIds as $row)
+        {
+            $this->cache['checkIfBackup'][$row['empUID']] = true;
         }
 
-        return true;
+        return isset($this->cache['checkEmployeeAccess'][$empUID]);
```
mgaoVA, line 1998: Note issue with wrong variable here fixed in https://github.com/department-of-veterans-affairs/LEAF/pull/2178

PR #2177, opened 2023-10-06
LEAF 4061 - Optimize form/query and getCustomData's read access check (2 of 3)

mgaoVA:
Note the PR merge base. This series of optimizations have been split into multiple PRs to simplify review.

This modifies getCustomData() to avoid duplicate checkReadAccess() queries. If checkReadAccess() was already run, then it's not necessary for getCustomData() to call checkReadAccess() again because it could only check a subset of the already checked records.

### Potential Impact
Read access to records

### Testing
Preconditions:
- A record that the current user doesn't have access to read
- The record doesn't show up in the Report Builder
- [x] After this patch, the current user still cannot open the record, and cannot see it in the Report Builder

### DIFF
--- a/LEAF_Request_Portal/sources/Form.php
+++ b/LEAF_Request_Portal/sources/Form.php
@@ -2798,7 +2798,14 @@ class Form
         return 1;
     }
 
-    public function query($inQuery)
+    /**
+     * query parses a JSON formatted user query defined in formQuery.js.
+     * 
+     * Returns an array on success, and string/int for malformed queries
+     * 
+     * @return mixed
+     */
+    public function query(string $inQuery): mixed
     {
         $query = json_decode(html_entity_decode(html_entity_decode($inQuery)), true);
         if ($query == null)
@@ -3636,7 +3643,7 @@ class Form
                 $indicatorIDs .= $indicatorID . ',';
             }
 
-            return $this->getCustomData($data, $indicatorIDs, $alreadyCheckedReadAccess);
+            $data = $this->getCustomData($data, $indicatorIDs, $alreadyCheckedReadAccess);
         }
 
         return $data;
--- a/LEAF_Request_Portal/sources/Form.php
+++ b/LEAF_Request_Portal/sources/Form.php
@@ -2309,9 +2309,9 @@ class Form
      * @param array $recordID_list
      * @param array $indicatorID_list
      * @param bool (optional) $alreadyCheckedReadAccess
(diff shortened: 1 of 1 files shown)


maxrichardjr (approved):

jampaul3 (approved):

LEAF_Request_Portal/sources/Form.php
```diff
@@ -2302,10 +2302,16 @@ public function getChildForms($recordID)
...
+     * WARNING: $alreadyCheckedReadAccess can only be set to true if $recordID_list has been
+     *          processed by checkReadAccess().
      *
+     * @param array $recordID_list
+     * @param array $indicatorID_list
+     * @param bool (optional) $alreadyCheckedReadAccess
      * @return array on success | false on malformed input
      */
-    public function getCustomData(array $recordID_list, string|null $indicatorID_list)
+    public function getCustomData(array $recordID_list, string|null $indicatorID_list, bool $alreadyCheckedReadAccess = false)
```
Pelentan, line 2314: Needs return type

LEAF_Request_Portal/sources/Form.php
```diff
@@ -3628,7 +3636,7 @@ public function query($inQuery)
                 $indicatorIDs .= $indicatorID . ',';
             }
 
-            return $this->getCustomData($data, $indicatorIDs);
+            return $this->getCustomData($data, $indicatorIDs, $alreadyCheckedReadAccess);
```
Pelentan, line 3639: Functions should have only one return.  Can this be turned to a variable and then passed back with the final return?

Pelentan (changes requested):

LEAF_Request_Portal/sources/Form.php
mgaoVA: Updated.

maxrichardjr (approved):

maxrichardjr (approved):

LEAF_Request_Portal/sources/Form.php
mgaoVA: Updated.

maxrichardjr (approved):

maxrichardjr (approved):

Pelentan (approved):

aerinkayne (approved):

raphaelfretter (approved):

AmandaSumner (approved):

PR #2184, opened 2023-10-12
Hotfix remove invalid comparison

mgaoVA:
This resolves an issue where a comparison is made on a non-existent property "dependencyID". It's not necessary to check the dependencyID in this function because the information is already provided in the second function argument (pdRecords).

### Potential Impact
Access rules that involve Person Designated records

### Testing
1. Create a record that routes to a designated person
2. As the designated person AND non-admin, the record should be visible in the Inbox

### DIFF
--- a/LEAF_Request_Portal/sources/FormWorkflow.php
+++ b/LEAF_Request_Portal/sources/FormWorkflow.php
@@ -175,7 +175,7 @@ class FormWorkflow
                         AND `disabled`=0";
         $res = $this->db->prepared_query($query, []);
 
-        // create map
+        // create map of recordIDs with "person designated"
         $dRecords = [];
         foreach($res as $record) {
             $dRecords[$record['recordID']]['data'] = $record['data'];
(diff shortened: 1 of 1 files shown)


Pelentan:
Good morning my lovely snuff-muse.

More proof that we don't have to worry about AI's.  This morning the
predictive text saw I typed "Good" and suggested "evening".

Keeping my eye on the Israeli war.  I don't know.  But this feels like it
could kick things off.  It will depend on how stupid the ragheads want to
be.  If they keep it focused just on Israel and some of the countries in
Europe, then probably not.  If they hit the US again, then yeah.  911 was
coordinated by a group of people inside the US government.  Likely all
Democrats, of course.  But it's not even a question anymore.  It was
_supposed_ to happen on Al Gore's watch so he could get a lot of the
changes they wanted done, including getting rid a lot of the Constitution.
However, people turned out en-masse to prevent the stealing of that
election.  But just barely.  They really didn't think Trump had a chance in
hell the first time around.  So the second time they overreacted and proved
how they steal elections.  This time they may try a two-fer.  Get us in a
war so they can cover up stealing the next election.  So far it's the only
plausible reason for this kicking off when it did.  It's the only trigger I
or any of my contacts can see.  So...

Any plans for this weekend?

Rolling into the day.

jampaul3 (approved):

Pelentan (approved):

aerinkayne (approved):

maxrichardjr (approved):

AmandaSumner (approved):

raphaelfretter (approved):

shaneodd (approved):

PR #2191, opened 2023-10-18
LEAF-4030: Multi Select Options on Email Templates not formatted correctly

raphaelfretter:
Implemented update/fix to have multiselect fields show in emails formatted properly and in basic bullets.

### Potential Impact
Emails sent with custom fields now show properly formatted multiselect field values.

### Testing
- Navigate to the Email Template Editor
- Update custom email template to include a multiselect field
- Navigate to Workflow Editor
- Alter action in designated workflow to include the custom email template being used to send emails
- Navigate to Form Editor
- Verify form is associated with above workflow
- Create new request using above form
- Submit the form and activate the above action
- Verify the field is formatted correctly in the sent email


LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -1497,10 +1497,10 @@ private function getFields(): array
                     }
                     break;
                 case (str_starts_with($format, "checkboxes") != false):
+                    $data = $this->buildCheckboxes(unserialize($data));
+                    break;
                 case (str_starts_with($format, "multiselect") != false):
-                    if ($this->isJsonString($data) && is_array(json_decode($data))) {
-                        $data = $this->buildMultiselect(json_decode($data));
-                    }
+                    $data = $this->buildMultiselect(unserialize($data));
```
Pelentan, line 1503: Removing checks makes me nervous.  What happens if it gets something that isn't json?  How badly will it fail?

Pelentan (changes requested):

LEAF_Request_Portal/sources/FormWorkflow.php
raphaelfretter: This check was causing the fail in the first place. I don't remember why this is checking for JSON. This is only ever serialized PHP arrays, which is a string in this context.

Pelentan (approved):

jampaul3 (approved):

maxrichardjr (approved):

pete-nerantzinis (approved):

aerinkayne (approved):

AmandaSumner (approved):

shaneodd (approved):

shaneodd (approved):

PR #2196, opened 2023-10-20
LEAF 4055 Combined Inbox Editor and Sitemap Search updates

aerinkayne:
Description

-Adds URL checks to the Inbox Editor app and the Sitemap Search report to prevent breaking errors caused by potential sitemap reports.
-Adds a check for unique sites for the Sitemap Search report and adds preprod and local domains to facilitate testing of this report.

Potential Impact
Admin area Combined Inbox Editor and Sitemap Search features

### DIFF
--- a/LEAF_Request_Portal/js/combined_inbox_editor/src/combined_inbox_editor.vue.js
+++ b/LEAF_Request_Portal/js/combined_inbox_editor/src/combined_inbox_editor.vue.js
@@ -93,7 +93,7 @@ const CombinedInboxEditor = Vue.createApp({
                             resolve();
                         });
                     },
-                    fail: (err) => {
+                    error: (err) => {
                         reject(err);
                     }
                 });
@@ -118,7 +118,7 @@ const CombinedInboxEditor = Vue.createApp({
                     success: (res) => {
                         resolve();
                     },
-                    fail: (err) => {
+                    error: (err) => {
                         reject(err);
                     }
                 });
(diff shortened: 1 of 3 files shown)


maxrichardjr (approved):

jampaul3 (approved):

LEAF_Request_Portal/js/combined_inbox_editor/src/combined_inbox_editor.vue.js
```diff
@@ -153,24 +153,37 @@ const CombinedInboxEditor = Vue.createApp({
...
-            return indicators.json();
+            return await indicators.json();
         }
     },
     created() {
         this.getMapSites().then(() => {
             this.sites.forEach((site) => {
-                this.getIndicators(site).then((indicators) => {
+                let portalURL = site.target;
+                if(portalURL.indexOf('/admin/') != -1) {
```
Pelentan, line 173: Is this going to run into the same issues Jamie has been for this "admin" name?

Pelentan (approved):

LEAF_Request_Portal/js/combined_inbox_editor/src/combined_inbox_editor.vue.js
raphaelfretter: It shouldn't because of the "/" on either side.

raphaelfretter (approved):

LEAF_Request_Portal/templates/reports/LEAF_Sitemap_Search.tpl
```diff
@@ -981,6 +981,24 @@ function buildURLComponents(baseURL, update){
     window.history.pushState('', '', url);
 }
 
+function getPortalURL(url = '') {
```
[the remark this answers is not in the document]
aerinkayne: I think it's better not to for now, since the Vue app is its own file and importing another JS file to use the method there seems more complicated.  If this turns out to be needed in some additional places, then we can figure out the best place to put it.  It might also be better to circle back to this after the JS file moves are finished anyway.

LEAF_Request_Portal/js/combined_inbox_editor/src/combined_inbox_editor.vue.js
shaneodd: https://localhost/LEAF_Request_Portal/admin/?a=formLibrary not 100% sure what is going on with the admin stuff right now, however that is a url I just grabbed off of local with the slashes both sides.

AmandaSumner (approved):

LEAF_Request_Portal/js/combined_inbox_editor/src/combined_inbox_editor.vue.js
aerinkayne: The above example would be fine as a card for what this ticket is addressing, because at this point it's trying to get the site's base URL to query for requests to display in the inbox.  The local formLibrary view has its own issues not related to this one (we don't have a local LEAF/library portal).

However, there is one portal named 'admin'.  They currently have no sitemap cards, but the inbox and other pages that make queries for requests based on card URLs would error if they were to add one for their portal or, -- more problematically -- if they were linked to with cards by other sites.  Since these URLs are from sitemap cards, it's not certain that the nexus of the site parsing the card and the nexus part of the card are the same (otherwise we could look for /admin/ not immediately preceded by the nexus).

aerinkayne: closing per AM discussion.

PR #2202, opened 2023-10-24
LEAF-4099 - Service chief users were not showing up 

shaneodd:
*Summary*

On the service page and within the modal users were not showing up. Data was showing behind the scenes. This was due to a value change within backup id. This went from an empty string to a null. 

*Testing*

This will be shown on the service chief page, without this change some or all user names will not show, with it will show all user.


maxrichardjr (approved):

shaneodd:
It turns out this was due to db update not firing properly on portal creation.

PR #2206, opened 2023-10-25
Leaf 4063.1.75 - constructing portal path

jampaul3:
Notice the base branch this is built off of.

Summary:
With this branch we begin to use the copied files and the new files created in the previous 2 branches. This branch was created as it's own to limit the amount of changes in the next (Leaf 4063.2) and final branch. Adding this to that branch made that branch extremely large. This branch deals with the new global sessions. It builds the portal path and stores it in a table on the national leaf launchpad db. This record will then be used in subsequent calls to be used as the portal, unless a new valid url is given. (It checks the current url, compares it to the saved url and if different will check the db to see if the new one is a valid portal or orgchart, if not it will continue to use the saved url, if the new url is valid it will update the db with the new url)

Potential Impact:
If the url obtained from the $_SERVER['SCRIPT_FILENAME'] isn't valid on the first visit a 404 will be displayed.  

Improvements to be made:
the global_sessions table has a datetime field that is updated on each update. This table is going to grow very large very quickly and could use a system to clean it out if a session has been in there and not used after a specified amount of time (maybe an hour).

Testing:
Regression testing would be ideal here. Things to keep an eye out for are hiccups, especially when going from one portal to another or portal to orgchart.


maxrichardjr (approved):

raphaelfretter (approved):

app/Leaf/Model/GlobalSession.php
```diff
@@ -17,7 +17,7 @@ public function getSession(string $csrf): array
     {
         $vars = array(':session' => $csrf);
         $sql = 'SELECT `csrf`, `session`, `established`
-                FROM `sessions`
+                FROM `global_sessions`
```
shaneodd, line 20: I may have missed it but where is this table getting created?

app/libs/loaders/Leaf_autoloader.php
```diff
@@ -93,7 +108,7 @@
 $oc_config = new Orgchart\Config($site_paths, $oc_settings);
 if (!defined('ORGCHART_CONFIG')) define('ORGCHART_CONFIG', $oc_config);
 
-ini_set('session.gc_maxlifetime', 2592000);
+//ini_set('session.gc_maxlifetime', 2592000);
```
shaneodd, line 111: Is this something we wanted to do or was this from testing changes? Is this setup in the docker configs?

shaneodd (changes requested):

app/Leaf/Model/GlobalSession.php
jampaul3: It's going to be created in the national leaf launchpad

app/libs/loaders/Leaf_autoloader.php
jampaul3: This has been in there. When I added the session_start to the top of this page this errored so I commented it out. I would assume we probably want to add it to the docker config or somewhere if this is really what we want.

shaneodd (approved):

aerinkayne (approved):

PR #2208, opened 2023-10-25
Leaf 4063.2 - references to national api endpoints

jampaul3:
Notice the base branch

Summary:
This pulls together everything that took place in the previous 3 branches. The references to the national api's have all changed to include v1 before the national. The nginx was updated in the previous branch to handle the new api endpoint url (example of an endpoint would be api/v1/national/employee/search...). The v1 tells nginx to route it to the new app/api/v1 folder.

Potential Impact:
With changing the api endpoint, we could run into situations where the endpoint isn't structured properly. If they are structured properly there should be minimal impact as the endpoints should return the same data as before.

Improvements to be made:
make it so the return is in a json format that we choose. And that the calling method handles this new format properly.

Testing:
Full regression testing needs to be done here. Special attention needs to be paid to areas where there is a employee lookup.

### DIFF
--- a/LEAF_Nexus/admin/templates/orgChart_import.tpl
+++ b/LEAF_Nexus/admin/templates/orgChart_import.tpl
@@ -252,7 +252,7 @@
                                                 getEmployeePositionData(fullSupervisorInfo.empUID).then(function(result){
 
                                                     supervisorPositions = result;
-                                                    var positionSupervisors = getPositionSupervisors(positionResponse.supervisor);
+                                                    var positionSupervisors = getPositionSupervisors(positionResponse.employeeList);
 
                                                     if(supervisorPositions){
                                                         var positions = supervisorPositions;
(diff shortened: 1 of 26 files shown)


maxrichardjr (approved):

shaneodd (approved):

jrodcode (approved):

LEAF_Nexus/admin/templates/orgChart_import.tpl
```diff
@@ -445,7 +445,7 @@
             return new Promise(function(resolve, reject){
                 $.ajax({
                     type: 'GET',
-                    url: '../api/national/employee/search?q=' + email + '&noLimit=0',
+                    url: '../api/v1/national/employee/search?q=' + email + '&noLimit=0',
```
mgaoVA, line 448: @Pelentan the current API should not be organized in a v1 like this. It creates unnecessary complexity and has downstream impact.

API related organization should be centered on v2 effort, and would require additional planning.

mgaoVA (changes requested):

PR #2209, opened 2023-10-25
LEAF-4019 - refresh orgchart employees disable

shaneodd:
The auto updater isn’t working correctly, as the national directory might not be updating. When I looked, their account was only marked as “disabled” after manually refreshing. This is a script that does not get sent out the the portals but gets setup in the scripts folder and run.

Issue:

The employee was not found in the ad. The script that runs nightly does not check to see when the last time that user was updated. With the refresh employee button on the orgchart side looks at the last updated time, if it has not been updated within a week it will disable the account.

Testing:
Users who have not been updated in over a set time will be deleted. Right now We have it set for 2 weeks but that time will be reduced to 1 before the final run.

Query to use in testing is below, the < unixtimestamp can be setup usingthe unix timestamp website.
`select * from employee where lastUpdated > 0 and lastUpdated < 1697057318 and deleted = 0`

### DIFF
--- a/scripts/scheduled-task-commands/disableOldUsers.php
+++ b/scripts/scheduled-task-commands/disableOldUsers.php
@@ -5,25 +5,9 @@ require_once LIB_PATH . '/php-commons/Db.php';
 // national_orgchart for stage/live leaf_user for dev
 $db = new Leaf\Db(DIRECTORY_HOST, DIRECTORY_USER, DIRECTORY_PASS, 'national_orgchart');
 
-$vars = [
-    ':lastUpdated' => strtotime('-2 weeks')
-    //':lastUpdated' => 1623606681
+$employeeVars = [
+    ':lastUpdated' => strtotime('-1 week'),
+    ':deleted' => time()
 ];
-
-$sql = "SELECT empUID, userName, lastUpdated FROM `employee` Where `lastUpdated` > 0 and `lastUpdated` < :lastUpdated and `deleted` = 0";
-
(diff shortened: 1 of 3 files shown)


maxrichardjr (approved):

maxrichardjr (approved):

raphaelfretter (approved):

maxrichardjr (approved):

scripts/scheduled-task-commands/disableOldUsers.php
```diff
@@ -0,0 +1,29 @@
+<?php
+require_once 'globals.php';
+require_once LIB_PATH . '/php-commons/Db.php';
```
jampaul3, line 3: require_once getenv('APP_LIBS_PATH') . 'globals.php';
require_once getenv('APP_LIBS_PATH') . 'globals_plus.php';

These should be used now rather than the above. The db is already required in globals_plus.php

jampaul3 (changes requested):

scripts/scheduled-task-commands/disableOldUsers.php
jampaul3: If you can't use the globals_plus.php (this is to get the portal_path) then you can include the db with this

require_once getenv('APP_PATH') . '/Leaf/Db.php';

aerinkayne (approved):

shaneodd:
There is a bit of code on the auth server that the call was commented out on staging and the call was missing on production. I have sorted this out and it appears to be working as it did previously.

PR #2222, opened 2023-11-06
LEAF-4123: Optimize "group designated" hotspot

mgaoVA:
This expands on earlier optimization efforts and improves performance when "group designated" records are involved. The optimization logic was largely copied from the "person designated" implementation to minimize complexity at the cost of some code duplication.

Performance improvement compared to Sprint-64-c1 when testing with ~500 "group designated" records:
```
goos: windows
goarch: amd64
pkg: LEAF/API-tester
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
                           │ sprint64c1.txt │     optimize-group-selected.txt     │
                           │     sec/op     │   sec/op     vs base                │
Inbox_nonAdminActionable-8     156.62m ± 4%   96.32m ± 2%  -38.50% (p=0.000 n=10)
```

Total improvement since Sprint-60-c2:
```
goos: windows
goarch: amd64
pkg: LEAF/API-tester
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
                           │ sprint60c2.txt │     optimize-group-selected.txt     │
                           │     sec/op     │   sec/op     vs base                │
Inbox_nonAdminActionable-8     426.97m ± 4%   96.32m ± 2%  -77.44% (p=0.000 n=10)
```

### Potential Impact
- Areas depending on form/query actionable data (Inbox, Report Builder)
- form/query access checks relating to "group designated" records

### Testing
- [x] You can access records that have been assigned to a designated group, and you are a member of that group, and you can take action on the record if it's on the requisite step - [experimental API test](https://github.com/mgaoVA/LEAF_test_experiments/commit/2166733fb04268569ded7ca03baf956894510496#diff-b7b1922b928b6c4ca9f0342a10188972bc7aa07897032f05faff1cb2b96f5234R59)
- [x] You cannot access records (as non-admin) that have been assigned to a designated group - [experimental API test](https://github.com/mgaoVA/LEAF_test_experiments/commit/2166733fb04268569ded7ca03baf956894510496#diff-b7b1922b928b6c4ca9f0342a10188972bc7aa07897032f05faff1cb2b96f5234R63)

### DIFF
--- a/LEAF_Request_Portal/sources/FormWorkflow.php
+++ b/LEAF_Request_Portal/sources/FormWorkflow.php
@@ -140,7 +140,7 @@ class FormWorkflow
     }
 
     /**
-     * includePersonDesignatedApproverData efficiently merges approver data to $srcRecords, for a
+     * includePersonDesignatedData efficiently merges approver data to $srcRecords, for a
      * given list of $pdRecordList and $pdIndicators.
      * 
      * When $skipNames is false, $srcRecords is expected to contain dependencyIDs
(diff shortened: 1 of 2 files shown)


jampaul3 (approved):

maxrichardjr (approved):

LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -222,6 +222,97 @@ private function includePersonDesignatedData(array $srcRecords, array $pdRecords
...
+        $gdIndicators = implode(',', $gdIndicators);
+
+        $gdRecordIDs = array_keys($gdRecordsMap);
+        $gdRecordIDs = array_map(function($x) {
+                        return (int)$x;
+                    }, $gdRecordIDs);
+
+        $gdRecordIDs = implode(',', $gdRecordIDs);
+
+        $query = "SELECT recordID, `data`, `name`, indicatorID FROM `data`
```
Pelentan, line 257: Is there a reason you're not using the "standard" PDO practice of feeding the variables for the sql statement in via the variables array here and above?

Pelentan (changes requested):

LEAF_Request_Portal/sources/FormWorkflow.php
mgaoVA: Yes, this was primarily to balance complexity and benefit, as parameterizing this would involves more logic. Even if the "WHERE IN" query was parameterized, the query would still have to be directly modified, necessitating validation of the modification, due to the unknown length of the IN(?, ?, ?,...) component. Since these queries only deal with integers, it's more straightforward to simply check the integers and make the validation easier to understand.

shaneodd (approved):

Pelentan (approved):

aerinkayne (approved):

AmandaSumner (approved):

raphaelfretter (approved):

PR #2223, opened 2023-11-06
Leaf 4084 - custom emails added to subordinate

jampaul3:
Summary:
This adds the custom email templates created on the Primary site to the list of what gets pushed to a subordinate.

Potential Impact:
The files are getting copied over, I'm not sure if there are db entries that need to be copied over as well or not. If so that will need to be added.

Testing:
1. On the primary site. create a custom email, assign it to a workflow
2. push changes out to the subordinates
3. create a request and make sure the custom email is received.


LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -47,6 +48,8 @@
 $db->query("delete FROM `actions`;");
 $db->query("delete FROM `records_dependencies` WHERE dependencyID=0;"); // delete invalid items -- TODO: figure out how these invalid items got written
 $db->query("TRUNCATE TABLE `step_modules`;");
+$db->query("TRUNCATE TABLE `email_templates`;");
+$db->query("TRUNCATE TABLE `events`;");
```
mgaoVA, line 52: If new tables are being wiped, they need to be paired with a matching import. Otherwise this would break events for the subordinate site.

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -5,16 +5,17 @@
...
 
-$res = $db->query_kv('SELECT * FROM settings', 'setting', 'data');
-
 // For Jira Ticket:LEAF-2471/remove-all-http-redirects-from-code
 //$protocol = isset($_SERVER['HTTPS']) ? 'https://' : 'http://';
 $protocol = 'https://';
 $siteRootURL = $protocol . HTTP_HOST;
-$relativePath = trim(str_replace($siteRootURL, '', $res['national_linkedPrimary']));
+
+$relativePath = trim(str_replace($siteRootURL, '', $settings['national_linkedPrimary']));
```
mgaoVA, line 13: Side note: If $settings is being initialized by the autoloader, we need to change the syntax to make it more obvious. For example rename to LEAF_SETTINGS. This would change it to a constant, and also uses a common all-caps convention for global declarations.

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -100,6 +102,11 @@ function importTable($db, $tempFolder, $table) {
 importTable($db, $tempFolder, 'workflow_routes');
 importTable($db, $tempFolder, 'step_modules');
 
+// remove files from templates/email/custom_override on subordinate
+passthru('rm -rf ' . $paste_custom_templates . '*');
```
mgaoVA, line 106: 1. There should be a check to ensure the folder exists.
2. PHP Filesystem Functions should be used instead of using OS-specific commands.

mgaoVA (changes requested):

maxrichardjr (approved):

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
jampaul3: I think I just forgot to remove this. It's now gone.

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
jampaul3: This global was created and added to this file. I created another Jira ticket to make it universal. $settings is used a lot throughout the code base.

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
jampaul3: Done.

maxrichardjr (approved):

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -100,7 +100,41 @@ function importTable($db, $tempFolder, $table) {
...
+    foreach ($files as $file) {
+        if ($file !== '.' && $file !== '..') {
+            // not sure if the / is needed here
+            $sourceFile = $source . '/' . $file;
+            $destinationFile = $destination . '/' . $file;
+
+            if (is_dir($sourceFile)) {
+                copyDirectory($sourceFile, $destinationFile);
+            } else {
+                copy($sourceFile, $destinationFile);
```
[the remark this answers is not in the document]
jampaul3: It shouldn't be needed as it's a portal and all portals have the same directory structure and so it should be there. But I did add the check JUST IN CASE.

shaneodd (approved):

aerinkayne (approved):

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -5,16 +5,17 @@
...
 $siteRootURL = $protocol . HTTP_HOST;
-$relativePath = trim(str_replace($siteRootURL, '', $res['national_linkedPrimary']));
+
+$relativePath = trim(str_replace($siteRootURL, '', LEAF_SETTINGS['national_linkedPrimary']));
 $tempFolder = $_SERVER['DOCUMENT_ROOT'] . $relativePath . 'files/temp/';
+$copy_custom_templates = $_SERVER['DOCUMENT_ROOT'] . $relativePath . 'templates/email/custom_override/';
+$paste_custom_templates = getenv('APP_LIB_PATH') . '../' . PORTAL_PATH . '/templates/email/custom_override/';
 
-if($res['siteType'] != 'national_subordinate') {
+if($settings['siteType'] != 'national_subordinate') {
```
mgaoVA, line 18:
```suggestion
if(LEAF_SETTINGS['siteType'] != 'national_subordinate') {
```

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -100,7 +100,49 @@ function importTable($db, $tempFolder, $table) {
...
+
 $db->query("ALTER TABLE `records_dependencies` ADD CONSTRAINT `fk_records_dependencyID` FOREIGN KEY (`dependencyID`) REFERENCES `dependencies`(`dependencyID`) ON DELETE RESTRICT ON UPDATE RESTRICT;");
 $db->query("ALTER TABLE `dependency_privs` ADD CONSTRAINT `fk_privs_dependencyID` FOREIGN KEY (`dependencyID`) REFERENCES `dependencies`(`dependencyID`) ON DELETE RESTRICT ON UPDATE RESTRICT;");
 $db->query("ALTER TABLE `category_count` ADD CONSTRAINT `category_count_ibfk_1` FOREIGN KEY (`categoryID`) REFERENCES `categories`(`categoryID`) ON DELETE RESTRICT ON UPDATE RESTRICT;");
 $db->query("ALTER TABLE `category_privs` ADD CONSTRAINT `category_privs_ibfk_2` FOREIGN KEY (`categoryID`) REFERENCES `categories`(`categoryID`) ON DELETE RESTRICT ON UPDATE RESTRICT;");
+
+function copyDirectory($source, $destination) {
+    $files = scandir($source);
+
+    foreach ($files as $file) {
```
mgaoVA, line 127: https://www.php.net/manual/en/function.scandir.php

Probably need to add an is_array() check on $files.

Since scandir() can return false, $files would not be iterable and the foreach would fail.

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -100,7 +100,49 @@ function importTable($db, $tempFolder, $table) {
...
+        if ($file !== '.' && $file !== '..') {
+            // not sure if the / is needed here
```
mgaoVA, line 129: Since the code is intended to be used in prod, this should be checked so we're sure.

The first iteration starts with a path ending with a slash, so the slash is not necessary. But this means subsequent iterations will need a "/" appended to the end.

It might be more clear if we make the naming convention more explicit.

For example:
- function copyDirectory($sourceDir, $destinationDir)  -- and document the function to explain it needs a trailing slash
- $newPath = $sourceDir . $name;
- if (is_dir($newPath)) { copyDirectory($newPath...

The function would still work even with extra //'s thanks to how the OS handles it, however making these changes would eliminate any ambiguity and avoid the "not sure".

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -100,7 +100,49 @@ function importTable($db, $tempFolder, $table) {
...
+                if (is_dir($destinationFile)) {
+                    copy($sourceFile, $destinationFile);
+                } else {
+                    // theoretically this should never be hit. All portals should have the same dir structure
+                    // thereby making this mute, but jsut in case.
+                    mkdir($destinationFile);
+                    copy($sourceFile, $destinationFile);
+                }
+
+            }
```
mgaoVA, line 145: I think your original implementation was fine.

scandir() already ensures you're either getting a directory or a file, and it's not necessary to check again.

```suggestion
            if (is_dir($sourceFile)) {
                copyDirectory($sourceFile, $destinationFile);
            } else {
                copy($sourceFile, $destinationFile);
            }
```
(double check that merging the suggestion worked, this has caused issues before)

mgaoVA (changes requested):
There are DB changes that need to occur for emails to work on subordinate sites.

For clarity, if the DB related changes were intended to be put in a separate PR, this should be noted in the title as a part 1 of 2.

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
jampaul3: It was not my intent to leave this for another PR. I have added them to this PR.

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
jampaul3: This was put here for testing purposes, our local environment is not setup to easily test this right now so I put the comment here to make debugging easier once we got to staging.

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
jampaul3: This has been changed back.

raphaelfretter (approved):

maxrichardjr (approved):

LEAF_Request_Portal/utils/LEAF_importStandardConfig.php
```diff
@@ -93,14 +97,50 @@ function importTable($db, $tempFolder, $table) {
...
+    // copy templates/email/custom_override from primary to subordinate
+    copyDirectory($copy_custom_templates, $paste_custom_templates);
+}
 
 $db->query("ALTER TABLE `records_dependencies` ADD CONSTRAINT `fk_records_dependencyID` FOREIGN KEY (`dependencyID`) REFERENCES `dependencies`(`dependencyID`) ON DELETE RESTRICT ON UPDATE RESTRICT;");
 $db->query("ALTER TABLE `dependency_privs` ADD CONSTRAINT `fk_privs_dependencyID` FOREIGN KEY (`dependencyID`) REFERENCES `dependencies`(`dependencyID`) ON DELETE RESTRICT ON UPDATE RESTRICT;");
 $db->query("ALTER TABLE `category_count` ADD CONSTRAINT `category_count_ibfk_1` FOREIGN KEY (`categoryID`) REFERENCES `categories`(`categoryID`) ON DELETE RESTRICT ON UPDATE RESTRICT;");
 $db->query("ALTER TABLE `category_privs` ADD CONSTRAINT `category_privs_ibfk_2` FOREIGN KEY (`categoryID`) REFERENCES `categories`(`categoryID`) ON DELETE RESTRICT ON UPDATE RESTRICT;");
+
+function copyDirectory($source, $destination) {
```
Pelentan, line 130: type hinting and return type.

Pelentan (changes requested):

Pelentan (approved):

maxrichardjr (approved):

PR #2230, opened 2023-11-13
Leaf 4130 - update string found

jampaul3:
Summary:
A script to replace the text from Leaf 4116. 4116 saved a file, this script will take that file and go through the array and replace the string searched for and replace it with a new string, or blank string to delete something.

Potential Impact:
This could have a huge impact as it is changing the content of files being used within the app. Special care needs to be taken to ensure that strings aren't being changed that shouldn't be. A careful observation of the json text file created in 4116 should be done to ensure that something isn't going to be changed that shouldn't be.

Testing:
This will require a full regression test. The amount of files and their content isn't readily available at this time. Not until a string has been determined to be replace.

### DIFF
--- a/scripts/evolving.php
+++ b/scripts/evolving.php
@@ -1,0 +1,14 @@
+<?php
+
+$data = file_get_contents('/var/www/log/time_to_evolve.txt');
+$replace = file_get_contents('/var/www/log/replacement.txt');
+$list = json_decode($data, true);
+
+foreach ($list as $value) {
+    // need to open the file to be modified
+    $file_data = file_get_contents($value['file_name']);
+    // modify the file
+    $file_data = str_replace($value['string'], $replace, $file_data);
+    // save the file
+    file_put_contents($value['file_name'], $file_data);
+}
--- a/scripts/time_to_evolve.php
+++ b/scripts/time_to_evolve.php
@@ -33,9 +33,13 @@ function iterator($filename)
     $dir = new DirectoryIterator($filename);
 
     foreach ($dir as $file) {
-        if ($file->isDot()) continue;
-        if ($file->isDir()) iterator($file->getRealPath());
-        getLineWithString($file->getRealPath());
+        if ($file->isDot()) {
+            continue;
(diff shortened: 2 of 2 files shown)


maxrichardjr (approved):

shaneodd (approved):

scripts/evolving.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+$data = file_get_contents('/var/www/log/time_to_evolve.txt');
```
raphaelfretter, line 3: Should be there a check for if the file exists?

scripts/evolving.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+$data = file_get_contents('/var/www/log/time_to_evolve.txt');
+$replace = file_get_contents('/var/www/log/replacement.txt');
```
raphaelfretter, line 4: Same here.

scripts/evolving.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+$data = file_get_contents('/var/www/log/time_to_evolve.txt');
+$replace = file_get_contents('/var/www/log/replacement.txt');
+$list = json_decode($data, true);
+
+foreach ($list as $value) {
+    // need to open the file to be modified
+    $file_data = file_get_contents($value['file_name']);
```
raphaelfretter, line 9: And here.

jampaul3: This is coming from a file that was created in a search so this file had better exist.

scripts/evolving.php
jampaul3: We shouldn't be running this if the files aren't there. This needs to be created in order to run this script.

scripts/evolving.php
jampaul3: This file is created in the other script that does the search. All of these are pre-requisites to run this script.

pete-nerantzinis (approved):

aerinkayne (approved):

PR #2235, opened 2023-11-14
LEAF 4139 update keypress listeners in orgchartForm

aerinkayne:
This updates the .keypress listeners in the Nexus orgchartForm js file to .on("keydown" listeners.

This should restore enter-save functionality of Nexus employee, position and group modals.  These were not changed in the recent update but seem to have been impacted by the addition of a dialogController modal keydown listener. 

The deprecated/removed jQuery size() method was also replaced by the .length property to fix a separate uncovered error.

Impact/Testing
Nexus side, editable (check write access) position, group and employee fields should save when enter is pressed in orgForm modals.  Confirm that these fields consistently save correctly (particularly the Position 'Pay Plan' field, which required a short timeout for the chosen dropdown). 
Confirm that updating a position Total Headcount field does not result in an error.

### DIFF
--- a/LEAF_Nexus/js/orgchartForm.js
+++ b/LEAF_Nexus/js/orgchartForm.js
@@ -63,7 +63,7 @@ orgchartForm.prototype.getForm = function (UID, categoryID, indicatorID) {
         success: function (response) {
           dialog.setContent(response);
           $("input:visible:first, select:visible:first").focus();
-          $("input:visible:first, select:visible:first").keypress(function (
+          $("input:visible:first, select:visible:first").on("keydown", function (
             event
           ) {
             if (event.which == 13) {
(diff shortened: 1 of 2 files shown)


Pelentan (approved):

shaneodd (approved):

maxrichardjr (approved):

raphaelfretter (approved):

jampaul3 (approved):

AmandaSumner (approved):

LEAF_Nexus/js/orgchartForm.js
```diff
@@ -100,11 +100,13 @@ orgchartForm.prototype.getForm = function (UID, categoryID, indicatorID) {
         success: function (response) {
           dialog.setContent(response);
           $("input:visible:first, select:visible:first").focus();
-          $("input:visible:first, select:visible:first").keypress(function (
+          $("input:visible:first, select:visible:first").on("keydown", function (
             event
           ) {
             if (event.which == 13) {
-              $("#" + dialog.btnSaveID).trigger("click");
+              setTimeout(() => { //timeout needed for ind 2 chosen dropdown
```
mgaoVA, line 107: Not sure what else might need to be reworked, but timeouts should not be used as they are less deterministic than typical event handlers.

mgaoVA (changes requested):

LEAF_Nexus/js/orgchartForm.js
aerinkayne: I found another (and I agree better) solution to this which should result in the behavior being what it had been previously.  If the chosen dropbox is open, the first 'enter' will just close it.

I had looked at the orgchart indicator formats and the position case looked like the only one that had a field with a dropdown - if this is not true I can also add this elsewhere.

maxrichardjr (approved):

PR #2236, opened 2023-11-15
LEAF-2478 - Histogram for sizes of uploaded files

shaneodd:
This is a script that will export a csv with the max, min, average filesizes and count of files. It outputs in MB to make things a bit more useful. 

 I also have a copy of the db vers, this  is setup for looking at portal db versions. This had not been copied over and this has been useful for helping find and verify things are working.

These files should have no impact since they are scripts run by developers only.

### DIFF
--- a/scripts/leaf-scripts/filehistorgram.php
+++ b/scripts/leaf-scripts/filehistorgram.php
@@ -11,6 +11,8 @@ $portals = $db->query("SELECT * FROM `sites` WHERE `site_type` = 'portal'");
 $directory = '/var/www/html';
 $totalfilecount = 0;
 $totalfilesize = 0;
+$totalmaxfilesize = 0;
+$totalminfilesize = 0;
 $fp = fopen('filehistogram.csv', 'w');
 fputcsv($fp, ['path', 'max (MB)', 'min (MB)', 'average (MB)', 'count']);
 foreach ($portals as $portal) {
@@ -20,7 +22,7 @@ foreach ($portals as $portal) {
 
     $glob_array = glob($directory . $portal['site_path'] . '/files/*.*');
     $filecount = count($glob_array);
-    $filzesize['count'] = $filecount;
+    $filzesizes['count'] = $filecount;
 
     foreach ($glob_array as $glob) {
         $filesize = filesize($glob);
@@ -31,8 +33,16 @@ foreach ($portals as $portal) {
             $filesizes['max'] = $filesize;
         }
 
-        if ($filesize < $filesizes['min']) {
+        if ($filesize < $filesizes['min'] && $filesize > 0) {
             $filesizes['min'] = $filesize;
+        }
+
+        if ($filesize > $totalmaxfilesize) {
+            $totalmaxfilesize = $filesize;
+        }
+
+        if ($filesize < $totalminfilesize && $filesize > 0) {
+            $totalminfilesize = $filesize;
         }
 
         $filesizes['average'] += $filesize;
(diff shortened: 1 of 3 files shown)


maxrichardjr (approved):

scripts/leaf-scripts/filehistorgram.php
```diff
@@ -0,0 +1,54 @@
+<?php
+error_reporting(E_ALL);
+ini_set('display_errors', 1);
+require_once 'globals.php';
+require_once LIB_PATH . '/php-commons/Db.php';
```
raphaelfretter, line 5: Does this pathing still work with current master @jampaul3 ?

jampaul3: I would do 
require_once getenv('APP_LIBS_PATH') . '/globals.php';
require_once getenv('APP_LIBS_PATH') . '../Leaf/Db.php';

scripts/leaf-scripts/filehistorgram.php
```diff
@@ -0,0 +1,54 @@
+<?php
+error_reporting(E_ALL);
+ini_set('display_errors', 1);
+require_once 'globals.php';
+require_once LIB_PATH . '/php-commons/Db.php';
+
+$startTime = microtime(true);
+
+$db = new Leaf\Db(DIRECTORY_HOST, DIRECTORY_USER, DIRECTORY_PASS, 'national_leaf_launchpad');
+
```
jampaul3, line 10: This will then have to change to $db = new App\Leaf\Db(...

scripts/leaf-scripts/showdbvers.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+require_once 'globals.php';
+require_once LIB_PATH . '/php-commons/Db.php';
+
+$startTime = microtime(true);
+
+$db = new Leaf\Db(DIRECTORY_HOST, DIRECTORY_USER, DIRECTORY_PASS, 'national_leaf_launchpad');
+
```
jampaul3, line 9: I would do the same here for the requires and the Db instantiation.

jampaul3 (changes requested):

maxrichardjr (approved):

maxrichardjr (approved):

maxrichardjr (approved):

aerinkayne (approved):

scripts/leaf-scripts/filehistorgram.php
```diff
@@ -0,0 +1,75 @@
...
+if ($totalminfilesize > $totalmaxfilesize) {
+    $totalminfilesize = 0;
+}
+
+fputcsv($fp, ['total', $totalmaxfilesize, $totalminfilesize, $totalfilesize / $totalfilecount, $totalfilecount]);
+fclose($fp);
+$endTime = microtime(true);
+$timeInMinutes = round(($endTime - $startTime) / 60, 2);
+echo "Destruction took {$timeInMinutes} minutes\r\n";
```
mgaoVA, line 74: Destruction?

scripts/leaf-scripts/showdbvers.php
```diff
@@ -0,0 +1,29 @@
+<?php
```
mgaoVA, line 1: What's this file for?

mgaoVA (changes requested):

scripts/leaf-scripts/filehistorgram.php
shaneodd: I have this text updated now.

scripts/leaf-scripts/showdbvers.php
shaneodd: Added usage text to this and the dbvers file that I had added

PR #2240, opened 2023-11-16
Leaf 2973 - Subordinate sites limited

jampaul3:
Summary:
When a site is configured as a Nationally Standardized Subordinate[0] site, the admin should not be able to:

Add new workflows[1]
Add new steps to existing workflows[2]
Add new forms[3]
Add new fields to existing forms[4]
For functions that can return a human readable error message, the error should state: "To make changes, please contact the administrator for the Nationally Standardized Primary site."

Potential Impact:
Really only for subordinate sites. The loophole that would allow an admin to modify workflows and forms has been closed.

Testing:
Being an admin and from a subordinate site access the workflow
    go to admin
    paste ?a=workflow to the end of the url
    from the console type newWorkflow(); and press enter
Do the same with form
    from admin add ?a=form to the url
    from the console type createForm(); press enter
In both instances you should get an alert with the message from above.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_form.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_form.tpl
@@ -2009,10 +2009,10 @@ function selectForm(categoryID) {
  */
 function showFormBrowser() {
     window.location = '#';
-    $('#menu').html('<div tabindex="0" role="button" class="buttonNorm" onkeydown="onKeyPressClick(event)" id="createFormButton" onclick="createForm();"><img src="../dynicons/?img=document-new.svg&w=32" alt="Create Form" /> Create Form</div><br />');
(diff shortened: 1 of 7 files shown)


maxrichardjr (approved):

raphaelfretter (approved):

shaneodd (approved):

pete-nerantzinis (approved):

mgaoVA:
Sorry, we should've done better backlog grooming.

This is no-longer needed because the existing implementation is sufficient.

The original issue is that people were making modifications to the workflow and forms on subordinate sites. We hid the editing buttons, and more importantly, added a large warning banner. This is sufficient because it communicates why they should not edit forms/workflows.

It's unnecessary to add a server-side check because the admin could easily change to a standard site, which bypasses the check.

PR #2253, opened 2023-11-29
LEAF-4107-Remove-Console-Logs-From-Editors

jrodcode:
### Summary:

Cleaned code for template editors to prevent console logs from displaying in dev tools during certain processes.

### Potential Impact:

Users will no longer encounter console logs during specified processes, enhancing the overall user interface.

### Testing:
1. Navigate to the template editor.
2. Open the browser's developer tools (F12).
3. Select the console tab.
4. Perform actions like saving and deleting. Confirm that no console logs are visible during these processes.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_templates_reports.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_templates_reports.tpl
@@ -218,7 +218,6 @@
             },
             url: '../api/applet/fileHistory/_' + currentFile,
             success: function(res) {
-                console.log("File history has been saved.");
                 getFileHistory(currentFile);
             }
         });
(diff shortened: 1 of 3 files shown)


LEAF_Request_Portal/admin/templates/mod_templates_reports.tpl
```diff
@@ -449,7 +446,6 @@
                 });
 
                 if (fileNames.indexOf(template) === -1) {
-                    console.log('Template file not found in directory');
```
shaneodd, line 452: Since this just returns and does nothing would we still want some sort of notification?  The change above makes sense since there is a message that gets displayed to the user. The change below would be another example where we would want to notify the user. In this case it is just console messages, but I feel like we need to still let the user know something somehow.

shaneodd (changes requested):

aerinkayne (approved):

maxrichardjr (approved):

jampaul3 (approved):

raphaelfretter (approved):

Pelentan (approved):

maxrichardjr (approved):

AmandaSumner (approved):

PR #2255, opened 2023-12-01
LEAF 4184 reset limit offset to 0 on filtered search

aerinkayne:
Sets the query limit offset back to 0 when applying a search filter to LEAF Library results.

Testing/Impact

LEAF Library page.
Business Lines filter should return filtered results.

Follow-up testing will need to occur on staging/preprod
This update cannot easily be tested locally because local dev would need a library portal with specific form/indicators and a sufficient number of posted forms.


raphaelfretter (approved):

shaneodd (approved):

maxrichardjr (approved):

aerinkayne:
Closing as discussed, since 4186 should cover this.

PR #2258, opened 2023-12-08
LEAF-4197 - Cleanup

mgaoVA:
This PR is a minor cleanup.
- Removes unused files
- Updates DB boilerplate files
- Updates ./app/libs/js files with their in-prod versions
- Adds a warning about the duplicate files in ./app/libs/js/LEAF/
- Modifies formQuery.js encodeReadableURI() to use a more easily repeatable pattern.

### Potential Impact
- Dependencies on FormQuery.js will be impacted, but there should be none.
- Launchpad site creation might be impacted due to DB boilerplate change (offline discussion TBD)
- Dependencies on formQuery.js (note case sensitive file), specifically where a plus sign is in a query.

### Testing
- [ ] There are no files that reference FormQuery.js
- [ ] No files reference instances of FormQuery, such as files containing "= FormQuery" or "= new FormQuery"
- [ ] Launchpad site creation works normally
- [ ] Report builder works normally:
       1. A record should contain the words "apple" AND "orange" (This is available in the new API-test branch's database)
       2. In the report builder, search for "Any standard data field CONTAINS EITHER +orange +apple"
       3. The results should only include records that contain both of the words.


aerinkayne (commented):
Wanted to mention that I noticed the new app folder version of formQuery is missing encodeReadableURL and the firstRun changes from rc65c3.  There are a couple nexus pages pulling that one.

Old libs folder version is most out of date, but nothing should be pulling that one anymore and in theory won't in the future.

madmaxrich (approved):

jampaul3 (approved):

raphaelfretter (approved):

app/libs/js/LEAF/formQuery.js
```diff
@@ -222,6 +231,13 @@ var LeafFormQuery = function () { //NOTE: keeping this a var in case custom code
     abortSignal = signal;
   }
 
+  /**
+   * encodeReadableURI provides minimal character URI encoding, prioritizing readible URLs
+   */
+  function encodeReadableURI(url) {
```
Pelentan, line 237: Should this be more future-proofed?  The name gives the impression that it does more than one thing.  So it probably should be set up initially to do so in order to ensure mistakes aren't made down the road.  Specifically, assigning the replacement to a local variable and then returning that.  That way when someone comes in to add to the functionality, they follow a pattern already established.

Pelentan (changes requested):

aerinkayne (approved):

app/libs/js/LEAF/formQuery.js
mgaoVA: I went ahead and updated it. Generally this should be avoided in this specific type of PR where we're simply re-syncing files instead of making any real changes. I've expanded the potential impact and testing to cover this new change.

Pelentan (approved):

shaneodd (approved):

shaneodd:
Things look good from testing locally, I do agree we will need to keep a good eye on the site creation. Also if we have some reports that we should look at that would be another avenue that we may want to go deeper on.

PR #2259, opened 2023-12-08
hotfix sub email template

jampaul3:
Summary:
When a primary site pushes changes to the subordinate sites tables on the subordinate sites are truncated or deleted and then repopulated with data from the primary, With an update that went out a week or two ago the export did not get updated to include the events and email_templates tables. But they were included on the import side. So these tables are getting deleted and nothing is being added back in from the primary. This PR makes sure that the sql files are getting created so they can be pushed to the subordinate.

Potential Impact:
Subordinates will have the events and email_templates now.

Testing:
Push contents from a primary out to it's subordinates and be sure that the email templates are on the subordinate sites.

### DIFF
--- a/LEAF_Request_Portal/utils/LEAF_exportStandardConfig.php
+++ b/LEAF_Request_Portal/utils/LEAF_exportStandardConfig.php
@@ -28,12 +28,14 @@ function exportTable($db, $tempFolder, $table) {
         case 'category_staples':
         case 'dependencies':
         case 'indicators':
+        case 'events':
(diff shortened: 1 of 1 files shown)


aerinkayne (approved):

Pelentan (approved):

jrodcode (approved):

pete-nerantzinis (approved):

madmaxrich (commented):
Reviewed and Approved

PR #2263, opened 2023-12-11
LEAF-4199 - Text updates

mgaoVA:
3 changes:
1. When there's a disabled account associated with a "person designated" request, the workflow shows a visible "\*NEEDS REASSIGNMENT\*" message.
2. Text related to "restore records" has been re-worded to clarify its status.
3. In the Inbox, approverName can contain the label of a field in the event there's a disabled account in a "person designated" field, which can contain HTML formatting. This change scrubs HTML from the text to ensure formatting is consistent.

## Potential Impact
- Admins may require notification to explain the "NEEDS REASSIGNMENT" message

## Testing
1. Set up a record that's pending a "person desginated" field
2. Disable the account associated with the "person desginated" field
- [ ] The workflow for the record should show "NEEDS REASSIGNMENT"
- [ ] The Inbox (Organize by Forms + View as Admin) should show "NEEDS REASSIGNMENT" as one of the sections

### DIFF
--- a/LEAF_Request_Portal/sources/FormWorkflow.php
+++ b/LEAF_Request_Portal/sources/FormWorkflow.php
@@ -206,8 +206,8 @@ class FormWorkflow
                     $approver = $dir->lookupEmpUID($dRecords[$v['recordID']]['data']);
                     
                     if (empty($approver[0]['Fname']) && empty($approver[0]['Lname'])) {
-                        $srcRecords[$i]['description'] = $srcRecords[$i]['stepTitle'] . ' (' . $dRecords[$v['recordID']]['name'] . ')';
-                        $srcRecords[$i]['approverName'] = $dRecords[$v['recordID']]['name'];
+                        $srcRecords[$i]['description'] = $srcRecords[$i]['stepTitle'] . ' ( *NEEDS REASSIGNMENT* ' . $dRecords[$v['recordID']]['name'] . ')';
+                        $srcRecords[$i]['approverName'] = '*NEEDS REASSIGNMENT* ' . $dRecords[$v['recordID']]['name'];
                         $srcRecords[$i]['approverUID'] = 'indicatorID:' . $res[$i]['indicatorID_for_assigned_empUID'];
                     }
                     else {
--- a/LEAF_Request_Portal/templates/print_form.tpl
+++ b/LEAF_Request_Portal/templates/print_form.tpl
@@ -1,5 +1,5 @@
 <!--{if $deleted > 0}-->
-    <div style="font-size: 36px"><img src="dynicons/?img=emblem-unreadable.svg&amp;w=96"
+    <div style="font-size: 36px"><img src="dynicons/?img=emblem-unreadable.svg&amp;w=96" alt=""
             style="float: left" /> Notice: This request has been marked as cancelled and will be permanently deleted.<br />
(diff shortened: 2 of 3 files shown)


LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -1,8 +1,8 @@
 <!--{if $deleted > 0}-->
-    <div style="font-size: 36px"><img src="dynicons/?img=emblem-unreadable.svg&amp;w=96" alt="Unreadable"
-            style="float: left" /> Notice: This request has been marked as deleted.<br />
+    <div style="font-size: 36px"><img src="dynicons/?img=emblem-unreadable.svg&amp;w=96"
```
aerinkayne, line 2: These alt attributes should just be empty strings (rather than absent)

raphaelfretter: Why does it need an alt tag? Screen readers will just ignore the element if there isn't one.

mgaoVA: Thanks, updated.

LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
```diff
@@ -189,6 +189,12 @@
         },
     };
 
+    function scrubHTML(input) {
```
Pelentan, line 192: A concern I have with this.  It doesn't do what it says it's doing.  It's just wrapping it in a div.  I would recommend either a different name or fully explain what it's _meant_ to do in comments inside the function itself.

LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -206,8 +206,8 @@ private function includePersonDesignatedData(array $srcRecords, array $pdRecords
                     $approver = $dir->lookupEmpUID($dRecords[$v['recordID']]['data']);
                     
                     if (empty($approver[0]['Fname']) && empty($approver[0]['Lname'])) {
-                        $srcRecords[$i]['description'] = $srcRecords[$i]['stepTitle'] . ' (' . $dRecords[$v['recordID']]['name'] . ')';
-                        $srcRecords[$i]['approverName'] = $dRecords[$v['recordID']]['name'];
+                        $srcRecords[$i]['description'] = $srcRecords[$i]['stepTitle'] . ' ( *NEEDS REASSIGNMENT* ' . $dRecords[$v['recordID']]['name'] . ')';
```
Pelentan, line 209: Thinking 508 compliance.  This _looks_ like it's trying to flag something for a viewer's attention.  Only what if they can't "view"?  Do we want to create a tag system for a "reader" to catch and give audible alert?

Pelentan (changes requested):

LEAF_Request_Portal/sources/FormWorkflow.php
mgaoVA: A screen reader would simply read out the text.

This prepends the text to the step's title when there's no first/last name associated with the approver, so it'll be seen/heard anywhere the stepTitle shows up.

jampaul3 (approved):

raphaelfretter (approved):

Pelentan (approved):

shaneodd (approved):

shaneodd:
I did not see any issues arising from these changes on local testing.

PR #2266, opened 2024-01-05
LEAF-3190 prevent built in workflow modification

aerinkayne:
Adds an update to prevent users from creating a Requestor --> End connection on build in workflows.  This correction is necessary because it creates an unavoidable help desk ticket as the user cannot undo this change.

Also prevents duplication of built in workflows and provides more obvious feedback about modifications to built-in workflow being restricted.

Impact / Testing

Workflow Editor
-It should no longer be possible to modify the initial requestor --> step connection on built in workflows.
-Confirm that custom workflows still behave as expected.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_workflow.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_workflow.tpl
@@ -983,7 +983,7 @@
         dialog.setSaveHandler(function() {
             addStep(currentWorkflow, $('#stepTitle').val(), function(stepID) {
                 if (isNaN(stepID)) {
-                    console.log(stepID);
+                    alert(stepID);
                 } else {
                     loadWorkflow(currentWorkflow);
                 }
@@ -1374,8 +1374,12 @@
                 type: 'DELETE',
                 url: `../api/workflow/${workflowID}/step/${stepID}/_${action}/${nextStepID}?`
                     + $.param({ 'CSRFToken': CSRFToken }),
-                success: function() {
-                    loadWorkflow(workflowID);
+                success: function(res) {
+                    if (+res === 1) {
+                        loadWorkflow(workflowID);
+                    } else {
+                        alert(res)
+                    }
                 },
                 error: (err) => console.log(err),
             });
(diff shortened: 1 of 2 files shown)


LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -983,7 +983,7 @@
         dialog.setSaveHandler(function() {
             addStep(currentWorkflow, $('#stepTitle').val(), function(stepID) {
                 if (isNaN(stepID)) {
-                    console.log(stepID);
+                    alert(stepID);
```
Pelentan, line 986: Why are we doing an alert with the id?  Can the user do anything  with it?  Will the average user know what it means?

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -1374,8 +1374,12 @@
                 type: 'DELETE',
                 url: `../api/workflow/${workflowID}/step/${stepID}/_${action}/${nextStepID}?`
                     + $.param({ 'CSRFToken': CSRFToken }),
-                success: function() {
-                    loadWorkflow(workflowID);
+                success: function(res) {
+                    if (+res === 1) {
+                        loadWorkflow(workflowID);
+                    } else {
+                        alert(res)
```
Pelentan, line 1381: Same here.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -2177,85 +2181,89 @@
...
+            $('.workflowStepInfo').css('display', 'none');
 
-        dialog.setTitle('Duplicate current workflow');
-        dialog.setContent('<br /><label for="description">New Workflow Title:</label> <input type="text" id="description"/><br /><br />The following will NOT be copied over:<br /><br />&nbsp;&nbsp;&nbsp;&nbsp;Data fields that show up next to the workflow action buttons');
-        dialog.setSaveHandler(function() {
-            let old_steps = {};
-            let workflowID;
-            let title = $('#description').val();
+            dialog.setTitle('Duplicate current workflow');
+            dialog.setContent('<br /><label for="description">New Workflow Title:</label> <input type="text" id="description"/><br /><br />The following will NOT be copied over:<br /><br />&nbsp;&nbsp;&nbsp;&nbsp;Data fields that show up next to the workflow action buttons');
```
Pelentan, line 2188: This doesn't make a lot of sense.  This is a very specific error message and is static.  When you say "the following", you expect to be more than two.   Thinking this should be run by Michael Gao and Su as I'm not sure of better wording myself.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -2177,85 +2181,89 @@
...
-                                AutomatedEmailReminders: {
-                                    'Automate Email Group': auto.AutomatedEmailReminders.AutomateEmailGroup,
-                                    'Days Selected': auto.AutomatedEmailReminders.DaysSelected,
-                                    'Date Selected': auto.AutomatedEmailReminders?.DateSelected || '',
-                                    'Additional Days Selected': auto.AutomatedEmailReminders.AdditionalDaysSelected,
-                                }
-                            };
+                                let seriesData = {
+                                    AutomatedEmailReminders: {
+                                        'Automate Email Group': auto.AutomatedEmailReminders.AutomateEmailGroup,
```
Pelentan, line 2223: Do we need to think about having "default" values for these?  Or are we sure there will always be values and these assignments will never fail to a null?

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -2177,85 +2181,89 @@
...
+                                        'Days Selected': auto.AutomatedEmailReminders.DaysSelected,
+                                        'Date Selected': auto.AutomatedEmailReminders?.DateSelected || '',
+                                        'Additional Days Selected': auto.AutomatedEmailReminders.AdditionalDaysSelected,
+                                    }
+                                };
+
+                                updateStepData(seriesData, stepID, function (res) {
```
Pelentan, line 2230: Why is an empty function being passed in?

Pelentan (changes requested):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
aerinkayne: My only change here was adding the if/else wrapper to prevent duplication of a built-in workflow.  Everything else in this function is from the duplicate workflow feature last summer.  It's easier to see if the master and branch files are side by side - not sure why just wrapping it made the diff extra crazy.

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
aerinkayne: 'Restricted Command' is sent back and already alerted if some other built-in workflow modifications are attempted.  This one was only logged, which does not explain to the user why the step was not added.

jampaul3 (approved):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
aerinkayne: This was deployed half a year ago

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
aerinkayne: If there is something fundamentally wrong with the workflow duplication feature, then I think it should be handled in a ticket specific to that.  This ticket was created to address an issue that makes it possible to unintentionally modify a built in workflow.  The only change for this function was a wrapper to prevent duplication of a built-in.  I guess since it added a tab to everything inside, it make it appear that there were a number of changes here (I had to look at the files side by side as a sanity check myself).

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
aerinkayne: This part was also unchanged.  I think it should be ok though - if stepData isn't null, the field should have these.  The only one that was in question (because emails by date was added later) was Date Selected, which already has a default value.  If additional kinds of stepData are added in the future I agree we should re-visit this, but I also don't want to mess with email reminder default values without looking at the related functionality in more detail.

PR #2269, opened 2024-01-08
LEAF 4194 - workflow requirement related logging

aerinkayne:
Adds requirement(dependency) related information to the workflow history.

Addition or removal of a requirement will be reflected in the workflow history modal.
If a requirement is associated with a step in the workflow, additional information about requirement group privileges and requirement description changes will also be listed.

Impact/Testing
Workflow Editor - confirm that logging occurs as expected and that other behavior remains as expected.

### DIFF
--- a/LEAF_Request_Portal/sources/Workflow.php
+++ b/LEAF_Request_Portal/sources/Workflow.php
@@ -800,9 +800,14 @@ class Workflow
         $res = $this->db->prepared_query('INSERT INTO step_dependencies (stepID, dependencyID)
                                             VALUES (:stepID, :dependencyID)', $vars);
 
+        $depVars = array(':dependencyID' => $dependencyID);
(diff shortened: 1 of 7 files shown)


LEAF_Request_Portal/sources/Workflow.php
```diff
@@ -1549,7 +1567,22 @@ public function getDescription($workflowID){
 
     public function getHistory($filterById)
```
Pelentan, line 1568: Let's get this function set up with hints and type returns.

Pelentan (changes requested):

raphaelfretter (approved):

shaneodd (approved):

LEAF_Request_Portal/sources/Workflow.php
aerinkayne: updated

Pelentan (approved):

shaneodd:
This appears to work as intended, not seeing any issues introduced by this change.

PR #2273, opened 2024-01-16
LEAF 4169 IFTHEN - new controllers and operators

aerinkayne:
Adds number and currency to available controller questions.

Adds comparison options:
-  greater than
-  greater than or equal
-  less than
-  less than or equal

These options are available if the controller format is number or currency, or if other controllers include number in their options.  A number type input field will appear if one of these operators is chosen.

Adjusts save/remove behavior:
Modal now has an explicit close button and does not automatically close on save/remove.

Other updates:
Changed data structures for some longer arrays and simplified setup.

### Testing / Impact

**Condition Entry**

Confirm app loads as expected and conditions can be added, removed and edited as expected.
Changes made to the form should be immediately reflected in the ifthen app without needing to reload the page.  If there are multiple conditions on a question, toggling between list items should not cause any unexpected behaviors.

**Form Entry**

Ensure that new and prior controller comparisons behave as expected.


docker/vue-app/src/form_editor/components/dialog_content/ConditionsEditorDialog.js
```diff
@@ -224,13 +242,23 @@ export default {
         getOperatorText(condition = {}) {
             const parFormat = condition.parentFormat.toLowerCase();
             let text = condition.selectedOp;
-            switch(text) {
+
+            const op = condition.selectedOp;
```
Pelentan, line 246: This bit gets repeated a lot.  We might want to look at pulling it out into into a class/function that can be used by everything that needs it.

Pelentan (approved):
My comment about pulling that bit of code out is for thought, not action at this time.

jampaul3 (approved):

shaneodd (approved):

PR #2275, opened 2024-01-18
LEAF 3428 remediate 508 issues (pass 1)

aerinkayne:
Addresses a number of **site wide** 508 related issues, with some specific focus on the **workflow editor**.

**General updates to template(tpl), report, and js files** include modifying or adding alt attributes, labels, titles, replacing spans/divs with buttons, improving contrast, adding lang attributes.  Following confirmation, some unused methods were also removed.  A tabbing issue for vue side modals was also addressed.  

### Testing / Impact

**Site Wide**
Changes outside of mod_workflow should not be significantly obvious to non screen reader users.  Changes here are generally not code-heavy, but regression testing of general site/nav behavior should still be undertaken in case of typos.

**Workflow Editor**
**mod_workflow.tpl** has a number of behavior related changes mostly related to modal behaviors and workflow step editing.
In addition to keyboard nav ability, the expected behaviors of the workflow editor functions should be confirmed.


LEAF_Request_Portal/admin/templates/mod_workflow.tpl
```diff
@@ -2958,27 +3099,18 @@
     this.portalAPI.setCSRFToken(CSRFToken);
 
 
-
     // Fix dialog boxes not going away when clicking outside of box
-    $(document).mouseup(function(e) {
+    $(document).on('mousedown', function(e) {
```
Pelentan, line 3103: This caught my attention as this is for 508 compliance.  Someone who is blind isn't going to be using a mouse.  So... do we need to come up with a composite stimulus check?

Pelentan (changes requested):

LEAF_Request_Portal/admin/templates/mod_workflow.tpl
aerinkayne: These dialogs are also set up to close on escape, and also with a new more obvious close button, so this should still be ok.

Pelentan (approved):

shaneodd (approved):

jampaul3 (approved):

PR #2279, opened 2024-01-26
LEAF-4071 - Windows Auth Server

shaneodd:
Issues:
This is to get the site to point to the correct end points. On mobile it was not correctly going to the endpoints needed to login. PIV login page on /login were not going to proper final url. 

Potential Impact:
If this is pushed before changes to the production auth server this will cause the login for mobile and piv to fail since it points to a new url. We can push changes to the auth server in advance of these changes to allow for the urls to be present.

Testing:

- [ ] Regression testing to make sure we are not introducing errors.
- [ ] GFE mobile phone will be required to do testing on staging and final testing on production. currently Michael Gao is the only one I know of that has this at this time.

### DIFF
--- a/LEAF_Request_Portal/sources/Login.php
+++ b/LEAF_Request_Portal/sources/Login.php
@@ -101,19 +101,14 @@ class Login
 
         if(defined('AUTH_TYPE') && AUTH_TYPE == 'cookie') {
             $authType = '/../auth_cookie/?r=';
-            $nonBrowserAuth = '/../auth_cookie/?r=';
-          }
+            $nonBrowserAuth = '/../auth_cert/?r=';
+        }
 
         if (!isset($_SESSION['userID']) || $_SESSION['userID'] == '')
         {
             if (php_sapi_name() != 'cli')
             {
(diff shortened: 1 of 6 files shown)


jampaul3 (approved):

Pelentan (approved):

aerinkayne (approved):

LEAF_Request_Portal/sources/Login.php
```diff
@@ -101,19 +101,14 @@ public function loginUser($userID='SYSTEM')
 
         if(defined('AUTH_TYPE') && AUTH_TYPE == 'cookie') {
             $authType = '/../auth_cookie/?r=';
-            $nonBrowserAuth = '/../auth_cookie/?r=';
-          }
+            $nonBrowserAuth = '/../auth_cert/?r=';
```
mgaoVA, line 104: Where are the files in auth_cert/* ?

mgaoVA (changes requested):

LEAF_Request_Portal/sources/Login.php
shaneodd: This has been added.

PR #2280, opened 2024-01-26
LEAF-4201 - Stop search button on main page

shaneodd:
Issue:
Stop button was not properly stopping on sites with larger amounts of data. This was due to the button being added and removed. You could sometimes get the script to stop if you spam clicked the stop button. This brings in some of the logic from the inbox as well as move the button slightly to an area where it is not rewritten to pages

Potential Issues:
Custom code sites would not get this change.

Testing:
This will require a site with 50-100k entries to successfully test, lower numbers could work however this is at the range to really see what is going on.

### DIFF
--- a/LEAF_Request_Portal/templates/view_search.tpl
+++ b/LEAF_Request_Portal/templates/view_search.tpl
@@ -1,5 +1,6 @@
 <section style="display: flex; flex-direction: column; width: fit-content;">
     <div id="searchContainer"></div>
+    <button id="btn_abortSearch" class="buttonNorm">Stop searching for more</button>
     <button id="searchContainer_getMoreResults" class="buttonNorm" style="display: none; margin-left:auto;">Show more records</button>
 </section>
 <script>
@@ -117,6 +118,7 @@ function renderResult(leafSearch, res) {
 
 function main() {
     let query = new LeafFormQuery();
+    let abortController = new AbortController();
     let leafSearch = new LeafFormSearch('searchContainer');
     leafSearch.setJsPath('<!--{$app_js_path}-->');
     leafSearch.setOrgchartPath('<!--{$orgchartPath}-->');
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/templates/view_search.tpl
```diff
@@ -130,6 +132,16 @@ function main() {
     let abortSearch = false;
     let scrollY = 0; // track scroll position for more seamless UX when loading more records
 
+    document.addEventListener("click", abortSearchListener );
+    function abortSearchListener(event){
+        var element = event.target;
```
Pelentan, line 137: Shouldn't this be a "let"?

LEAF_Request_Portal/templates/view_search.tpl
```diff
@@ -1,5 +1,6 @@
 <section style="display: flex; flex-direction: column; width: fit-content;">
     <div id="searchContainer"></div>
+    <button id="btn_abortSearch" class="buttonNorm">Stop searching for more</button>
```
Pelentan, line 3: Please check with Carrie to see if this needs a label to be 508 compliant.

Pelentan (changes requested):

LEAF_Request_Portal/templates/view_search.tpl
shaneodd: Yes, switched to let.

jampaul3 (approved):

LEAF_Request_Portal/templates/view_search.tpl
shaneodd: Added type button, it is a thing that may or may not be needed but helps inform tools that it is in fact a button even through it is a button element.

Pelentan (approved):

aerinkayne (approved):

PR #2285, opened 2024-02-02
Leaf 4204 - Setting the 404 on decommissioned sites

jampaul3:
Summary:
Most of the work for this ticket happened on the Leaf-site_specific repo but we did have to account for decommissioned sites in the base code here. There will be some setup in the workflow once this goes out but all the coding is here for it to function.

Potential Impact:
This will allow a leaf coach to decommission a site from the launchpad. The only potential negative impact I can see is the fields in the db don't get added, causing a fatal for a field not being found in the query.

Testing:
Testing of this can be found in the Leaf-site_specific repo PR.


app/Leaf/Site.php
```diff
@@ -82,7 +82,7 @@ private function checkPath(): array
     {
         $vars = array(':site_path' => $this->match);
         $sql = 'SELECT `site_path`, `site_uploads`, `portal_database`, `orgchart_path`,
-                    `orgchart_database`
+                    `orgchart_database`, `launchpadID`, `decommission`
                 FROM `sites`
```
mgaoVA, line 86: Couple things:
- This highlights another prerequisite that didn't come up before. The "sites" table is apparently missing from the [resources database boilerplate](https://github.com/department-of-veterans-affairs/LEAF/blob/master/docker/mysql/db/boilerplate/resource_database_boilerplate.sql), and should be added. All Request Portals should have identical **core** codebases and database schemas.
- LEAF-4204 outlines a specific naming convention for the new column: "decommissionTimestamp". This should not be changed unless there's a compelling reason.

app/libs/loaders/Leaf_autoloader.php
```diff
@@ -23,12 +23,17 @@
 $site = new App\Leaf\Site($file_paths_db, $_SERVER['SCRIPT_FILENAME']);
 
 if ($site->error) {
-    throw new Exception("Sorry the page you are looking for could not be found, please check the url and try again.");
```
mgaoVA, line 26: Note that http_response_code() does not abort processing, and removing the throw substantially changes program flow. Not sure if this was intended.

mgaoVA (changes requested):

app/libs/loaders/Leaf_autoloader.php
jampaul3: When I tested the throw on staging it threw a 500 with no real message, that is why I changed it to this and added the die() right after it. This way the nginx will capture the 400 and send them to a 4xx page.

app/Leaf/Site.php
jampaul3: Ok, these have been addressed and will be pushed soon.

shaneodd (approved):

aerinkayne (approved):

Pelentan (approved):

Pelentan (approved):

PR #2287, opened 2024-02-12
LEAF 4223 idea inbox customizations

aerinkayne:
**Combined Inbox Editor (beta) and LEAF_Inbox options updates**

-Fix bug that causes non portal sitemap cards to be deleted.

-Fix error that occurred if the sitemap card target does not end in a slash.

-Add js XSSHelpers to strip out all potential HTML in custom header names and update custom headers to use the short label if it's available instead of the full indicator name for the multiselect plugin options.

-Add Priority and Date Submitted header options.  Inbox view will automatically sort by priority if this header is present.

-Add ability to add form-specific settings for indicator questions.

-Inbox View 508 Adjustments:
Category Headers are now buttons instead of divs for keyboard accessibility.
Top buttons for view and organization options now tab left to right.

**Testing / Impact**
This update is specific to the admin side Combined Inbox Editor and the combined inbox view (LEAF_Inbox).
Errors noted above should no longer occur on the Inbox Editor, and both the Editor and Inbox should otherwise continue to behave as expected.

### DIFF
--- a/LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
+++ b/LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
@@ -446,17 +446,19 @@
             });
         }
 
-        let customCols = [];
+        let headerColumns = "";
         if (customColumns === false) {
             const baseColumns = site.columns == null || site.columns == 'UID' ? 'UID,service,title,status' : site.columns;
             const formColumns = site.formColumns[categoryID] || null;
             if (formColumns !== null) {
-                site.columns = formColumns;
+                headerColumns = formColumns;
             } else {
-                site.columns = site.columns == null || site.columns == 'UID' ? 'UID,service,title,status' : site.columns;
+                headerColumns = site.columns == null || site.columns == 'UID' ? 'UID,service,title,status' : site.columns;
             }
+            headerColumns = headerColumns.split(",")
         }
-        site.columns.split(',').forEach(col => {
+        let customCols = [];
+        headerColumns.forEach(col => {
             if (isNaN(col)) {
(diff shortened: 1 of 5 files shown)


shaneodd (approved):

jampaul3 (approved):

LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
```diff
@@ -561,20 +610,30 @@
...
-                            break;
-                        default:
-                            break;
-                    }
+
+        const siteColumns = (site?.columns || "").split(',').filter(c => c !== "");
+        const formColumns = site?.formColumns || {};
+        let arrFormColumns = [];
+        for (let key in formColumns) {
+            let cols = formColumns[key];
```
Pelentan, line 618: Variables shouldn't be defined inside of loops.  The definition  should be set outside, and then use a conditional assignment so it either puts in the new value or null.

LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
```diff
@@ -561,20 +610,30 @@
...
+            cols = cols.split(',');
+            arrFormColumns = arrFormColumns.concat(cols)
+        }
+        let allColumns = siteColumns.concat(arrFormColumns);
+        allColumns = Array.from(new Set(allColumns));
+
+        for (let i in allColumns) {
+            const col = allColumns[i];
```
Pelentan, line 626: Same here.

Pelentan (changes requested):

Pelentan (approved):

PR #2289, opened 2024-02-13
LEAF-4242 -Fix typo

mgaoVA:
This fixes a typo in docker/vue-app/src/form_editor/components/FormBrowser.js

### Potential Impact
A lot of files were changed after running "npm run build". Is this normal?

### Testing
- Form Editor works normally
- Site Designer works normally (not sure why that file was modified in the build process)


aerinkayne:
I suspect the build differences are due to a version update of dependencies but will take a closer look. 
The older package.json associated with the gulp build did not use exact versions so this one does not either (nor is it necessarily good to completely prevent it) but that can also cause docker build differences and could be worth reconsidering.  

There are a couple components and some styling common to both Form Editor and Site Designer so that would explain changes to both.

jampaul3 (approved):

aerinkayne (approved):

Pelentan (approved):

PR #2301, opened 2024-02-23
Feature/leaf 3472/handle large queries

shaneodd:
Summary
This is to handle large queries from the report section of the website. This is mainly going to be for the people who use the api to pull in data into power bi and excel. The process will be setup to recheck to see if there are any new entries every 5 minutes. This should allow more than enough time for processes to clear run their course before calling more.

Opportunities for further improvement
We will need to maybe adjust the frequency or number of calls at a time.

Potential Impacts
This could have impact anywhere we are loading in data from the api that the report builder uses. Custom code may have issues if they are not slotting in the data in 1000 row chunks. They will get a 202 error code and some text to let them know their data will be processed.

Testing
Report builder, the entry page of the portal and mass actions are areas that we are currently using the updated api. Portals with custom code should be tested as well to make sure they load data properly

One note, this is a table that will need to go to the launch pad, since we do not have an official way of updating that at this time here is the database create.
CREATE TABLE site_large_query(idINT(10) UNSIGNED NOT NULL AUTO_INCREMENT,site_idINT(10) UNSIGNED NOT NULL,created_atINT(10) UNSIGNED NULL DEFAULT NULL,currently_running INT(10) UNSIGNED NULL DEFAULT '0', PRIMARY KEY (id) USING BTREE, UNIQUE INDEX site_id (site_id) USING BTREE, INDEX created_at (created_at) USING BTREE, INDEX currently_running (currently_running) USING BTREE ) COLLATE='latin1_swedish_ci' ENGINE=InnoDB AUTO_INCREMENT=26 ; 

Prerequisites to Prod Deployment

 Need a script to search for "api/form/query/?q={" within custom templates, Identify sites where they occur, reach out to affected admins informing them of the change (TBD: work with comms on message)
 Need the ability for LEAF dev team members to easily visualize the status of the query queue


mgaoVA (changes requested):
Need to catch up to master: "339 commits behind"

LEAF_Request_Portal/sources/Form.php
```diff
@@ -2842,36 +2842,433 @@ public function changeFormType($recordID, $categories)
         return 1;
     }
 
+ /**
+     * @param array $query - this is the decoded string used for query
+     * @return bool
+     */
+    private function isProcessQuery(array $query): bool
```
mgaoVA, line 2849:
```suggestion
 /**
     * isLargeQuery determines whether a query could potentially return a large result
     * @param array $query - this is the decoded string used for query
     * @return bool
     */
    private function isLargeQuery(array $query): bool
```

Rename to align with intended behavior.

LEAF_Request_Portal/sources/Form.php
```diff
@@ -2842,36 +2842,433 @@ public function changeFormType($recordID, $categories)
...
+        
+        $externalProcessQuery = false;
+        if (isset($query['limit']) && is_numeric($query['limit'])) {
+
+            // limit is > 10,000 records
+            if ($query['limit'] > 10000) {
+               
+                $externalProcessQuery = true;
+                // limit is <= 10,000 records AND more than 10 indicators are requested
+            } else if ($query['limit'] > 10000 || (!isset($query['limit']) && (!empty($query['getData']) && count($query['getData']) >= 10))) {
```
mgaoVA, line 2860: Redundant: $query['limit'] > 10000

LEAF_Request_Portal/sources/Form.php
```diff
@@ -2842,36 +2842,433 @@ public function changeFormType($recordID, $categories)
...
+ /**
+     * @param array $query - this is the decoded string used for query
+     * @return bool
+     */
+    private function isProcessQuery(array $query): bool
+    {
+        
+        $externalProcessQuery = false;
+        if (isset($query['limit']) && is_numeric($query['limit'])) {
+
```
mgaoVA, line 2854: The change in print_form.tpl identifies another condition that we can account for, and avoid having to add the limit over there:  if there's a recordID term, it wouldn't be considered a large query since there'd only be at most one record.

mgaoVA (changes requested):

LEAF_Request_Portal/sources/Form.php
```diff
@@ -3141,127 +3524,106 @@ public function query(string $inQuery): mixed
...
+                            ) {
                                 $dataTerm = 'lj_data.data';
                             }
 
                             $dataMatch = ":data{$count}";
-                            switch ($tResTypeHint[0]['format']) {
-                            case 'number':
-                            case 'currency':
-                                $dataTerm = "CAST({$dataTerm} as DECIMAL(21,5))";
+                            switch (strtolower($tResTypeHint[0]['format'])) {
```
mgaoVA, line 3597: We're considering the format case sensitive
```suggestion
                            switch ($tResTypeHint[0]['format']) {
```

mgaoVA:
Notes for discussion next week:
- Might need to take a different approach in order to streamline the process for users (run all queries, abort and queue large ones)
- Minimize large query handling logic in the standard query function (only interface with the input and output)
- The result from processQueryData can potentially be empty and still have more data to retrieve (see "LEAF-Query: continue")

shaneodd:
This will be kept around as a reference for future work

PR #2305, opened 2024-02-29
LEAF-4256 - PHP Warning Undefined array key

shaneodd:
Issue:

Error log is overrun by this error, need to add in checks to make sure this goes away. 

[23-Feb-2024 00:00:01 UTC] PHP Warning:  Undefined array key "xxxxxxx" in /var/www/html/VISNx/xxx/orgchart/scripts/refreshOrgchartEmployees.php on line 250

 

Testing:

To do testing you will need an active user in the local orgchart that does not exist in the national orgchart. (deleted user at a national level)

You will also need to have a user that is not locally added at time of script (new user) and verify data for that user is pulled in.

### DIFF
--- a/LEAF_Nexus/scripts/refreshOrgchartEmployees.php
+++ b/LEAF_Nexus/scripts/refreshOrgchartEmployees.php
@@ -156,7 +156,24 @@ function updateLocalOrgchartBatch()
         updateEmployeeDataBatch($localEmployeeUsernames);
     }
 }
+/**
+ * @param getEmployeeUIDs array $localEmployeeUsernames
+ * @return [$localEmpUIDs,$localEmpArray]
+ */
+function getEmployeeUIDs(array $localEmployeeUsernames): array {
+    global $oc_db;
 
+    // get local empuids, need to gather employees that have been added. 
+    $localEmployeeSql = "SELECT empUID, userName FROM employee WHERE userName IN (".implode(",",array_fill(1, count($localEmployeeUsernames), '?')).")";
+    $localEmpUIDs = $oc_db->prepared_query($localEmployeeSql,$localEmployeeUsernames);
+
+    $localEmpArray = [];
+    foreach ($localEmpUIDs as $localUsername) {
+        $localEmpArray[$localUsername['userName']] = $localUsername['empUID'];
+    }
+
+    return [$localEmpUIDs,$localEmpArray];
+}
 
 function updateEmployeeDataBatch(array $localEmployeeUsernames = [])
 {
@@ -184,13 +201,7 @@ function updateEmployeeDataBatch(array $localEmployeeUsernames = [])
     $orgEmployeeRes = $globalDB->prepared_query($orgEmployeeSql, $localEmployeeUsernames);
 
     // get local empuids
-    $localEmployeeSql = "SELECT empUID, userName FROM employee WHERE userName IN (".implode(",",array_fill(1, count($localEmployeeUsernames), '?')).")";
-    $localEmpUIDs = $oc_db->prepared_query($localEmployeeSql,$localEmployeeUsernames);
-
-    $localEmpArray = [];
-    foreach ($localEmpUIDs as $localUsername) {
-        $localEmpArray[$localUsername['userName']] = $localUsername['empUID'];
-    }
+    list($localEmpUIDs,$localEmpArray) = getEmployeeUIDs($localEmployeeUsernames);
 
     //if for some reason there is no data, we need to stop right there.
     if (empty($orgEmployeeRes)) {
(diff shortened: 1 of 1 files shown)


LEAF_Nexus/scripts/refreshOrgchartEmployees.php
```diff
@@ -243,20 +254,33 @@ function updateEmployeeDataBatch(array $localEmployeeUsernames = [])
...
-            // this will need to be checked into further I cannot silence this warning without possibly breaking things.
-            'empUID' => $localEmpArray[$orgEmployeeData['userName']],
-            'indicatorID' => $orgEmployeeData['indicatorID'],
-            'data' => $orgEmployeeData['data'],
-            'author' => $orgEmployeeData['author'],
-            'timestamp' => $orgEmployeeData['timestamp'],
-        ];
-    }
+        // if this user is not found, we will skip adding data for them.
+        if(empty($localEmpArray[$orgEmployeeData['userName']]) == true ){
```
jampaul3, line 264: Do we need the == true in this case? the empty method will return true or false.

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
```diff
@@ -243,20 +254,33 @@ function updateEmployeeDataBatch(array $localEmployeeUsernames = [])
...
+            continue;
+        }
+        else{
```
jampaul3, line 267: PSR 12 states that the else needs to be on the same line as the closing { and there should be a space before and after the else.

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
```diff
@@ -243,20 +254,33 @@ function updateEmployeeDataBatch(array $localEmployeeUsernames = [])
...
+                'timestamp' => $orgEmployeeData['timestamp'],
+            ];
+        }
 
-    $oc_db->insert_batch('employee_data',$localEmployeeDataArray,['indicatorID','data','author','timestamp']);
+        
+    }
 
+    // make sure data array has data before attempting to insert data
+    if(empty($localEmployeeDataArray) == false){
```
jampaul3, line 281: Same here, do we need the == false?

jampaul3 (changes requested):

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
```diff
@@ -156,7 +156,24 @@ function updateLocalOrgchartBatch()
...
 }
+/**
+ * @param getEmployeeUIDs array $localEmployeeUsernames
+ * @return [$localEmpUIDs,$localEmpArray]
+ */
+function getEmployeeUIDs(array $localEmployeeUsernames): array {
+    global $oc_db;
 
+    // get local empuids, need to gather employees that have been added. 
+    $localEmployeeSql = "SELECT empUID, userName FROM employee WHERE userName IN (".implode(",",array_fill(1, count($localEmployeeUsernames), '?')).")";
```
Pelentan, line 167: This needs to be split into the normal way of setting up sql queries.  We definitely don't want complex actions inside an assignment like this.

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
Pelentan: I know I harp on comparing things correctly.  In this case Jamie is correct.  empty() returns true or false so is perfectly fine just being directly referenced.

Pelentan (changes requested):

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
shaneodd: removed true, had it since it made it very clear we are looking at true/false.

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
shaneodd: removed false, had it since it made it very clear we are looking at true/false.

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
shaneodd: adjusted, I have to review the plugin for this since it used to auto format this for me and did not catch that it did not format.

LEAF_Nexus/scripts/refreshOrgchartEmployees.php
shaneodd: Broken this out, there were a few other like this.

jampaul3 (approved):

Pelentan (approved):

aerinkayne (approved):

PR #2315, opened 2024-03-06
Leaf 4260 - service headings now show correctly

jampaul3:
Summary:
The headings on the services page were not always correct. The issue was pulling the service rather than the group name which was brought about by the ELT/Quadrad not being added to the groups table unless they were tagged in the ELT on the Nexus side. This change will add the ELT to the groups table no matter what. The query was changed back to using the group name.

Potential Impact:
This should just show the correct ELT name now. 

Testing:
Ensure there is an ELT in the Nexus that is not tagged for the portal
Complete a sync services
check that the ELT is the heading of a service

### DIFF
--- a/LEAF_Request_Portal/sources/System.php
+++ b/LEAF_Request_Portal/sources/System.php
@@ -122,9 +122,7 @@ class System
                             if ($backups['status']['code'] == 2) {
                                 // check if this service is also an ELT
                                 // if so, update groups table
-                                $tagged = $tag->groupIsTagged($serviceID, Config::$orgchartImportTags[0]);
-
-                                if ($serviceID == $quadID && $tagged['status']['code'] == 2 && !empty($tagged['data'])) {
+                                if ($serviceID == $quadID) {
                                     $this->updateGroup($serviceID, $oc_db);
                                 } else {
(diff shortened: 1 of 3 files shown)


Pelentan (approved):

aerinkayne (approved):

LEAF_Request_Portal/sources/System.php
```diff
@@ -122,9 +122,7 @@ public function updateService(int $serviceID): array
                             if ($backups['status']['code'] == 2) {
                                 // check if this service is also an ELT
                                 // if so, update groups table
-                                $tagged = $tag->groupIsTagged($serviceID, Config::$orgchartImportTags[0]);
-
-                                if ($serviceID == $quadID && $tagged['status']['code'] == 2 && !empty($tagged['data'])) {
+                                if ($serviceID == $quadID) {
```
shaneodd, line 125: What is this status code stuff? I see we are checking a bunch of different variables with this status code and we are not checking if this tagged is empty. I may be missing something obvious why this was needed or not needed.

shaneodd (changes requested):
I have some questions, maybe I am missing a point but want to make sure we want to remove that check.

PR #2317, opened 2024-03-06
Issue/leaf 4241/mail error no recipients

Pelentan:


shaneodd:
Closing, query that was originally here was not the culprit. Using the branch to help troubleshoot at this time. The issue may lie with how the login info is set before sending emails.

PR #2318, opened 2024-03-06
Remove the *_format property from api/form/query

mgaoVA:
This reduces overhead in user-created queries by removing an underutilized property, which improves load time and makes more efficient use of network resources.

Before:
```
[1]: "Hello"
[2]: "World"
[1_format]: "text"
[2_format]: "text"
```

After:
```
[1]: "Hello"
[2]: "World"
```

In _api/form/query_, the _*_format_ property is apparently only used to resolve a date handling issue in Excel exports. Since the majority of queries will never need the format, I took a different approach to resolve the issue by using _api/form/indicator/list_ only when the user requests an Excel export.

### Potential Impact
Is the _*_format_ property used anywhere else? If no, then there are no other dependencies and minimal impact

### Testing
- [ ] Existing date handling logic works in Excel exports
   - [ ] Text field containing 1/23/45 is not seen in Excel as the date Jan 23, 1945
   - [ ] Date field containing 1/23/45 is seen in Excel as the date Jan 23, 1945

### DIFF
--- a/LEAF_Request_Portal/js/formGrid.js
+++ b/LEAF_Request_Portal/js/formGrid.js
@@ -762,13 +762,6 @@ var LeafFormGrid = function (containerID, options) {
                                            }"
                                            data-indicator-id="${
                                              headers[j].indicatorID
-                                           }"
-                                           data-format="${
-                                             currentData[i].s1[
-                                               "id" +
-                                                 headers[j].indicatorID +
-                                                 "_format"
-                                             ]
                                            }">
                                             ${data.data}</td>`;
             }
@@ -886,6 +879,7 @@ var LeafFormGrid = function (containerID, options) {
   }
 
   /**
+   * @deprecated See example.tpl for more efficient formGrid usage.
    * @memberOf LeafFormGrid
    */
   function loadData(recordIDs, callback) {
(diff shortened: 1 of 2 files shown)


aerinkayne (approved):

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -992,7 +986,14 @@ var LeafFormGrid = function (containerID, options) {
         'dynicons/?img=x-office-spreadsheet.svg&w=16" alt="" /> Export</button>'
     );
 
-    $("#" + prefixID + "getExcel").on("click", function () {
+    $("#" + prefixID + "getExcel").on("click", async function () {
+      // get indicator formats in case they need special handling (e.g. dates)
```
aerinkayne, line 990: Are all the formIDs present in the report available at this point by any chance?

mgaoVA: The formIDs aren't guaranteed to be here since it'd require the user to use a Type filter.

All the indicatorIDs are present at this point, but I'll leave expanding the API for a separate story.

Pelentan (approved):

Pelentan (approved):

shaneodd (approved):

jampaul3 (approved):

PR #2331, opened 2024-03-20
LEAF 4269 - 508 audit remediations

aerinkayne:
Please see Jira ticket 4269 for a list of changes.

Updates a variety of nav/menus and interactive elements in both the Nexus and Portal, as well as some style, verbiage and title adjustments.
Apart from the portal Workflow Editor workflow and step selectors being switched, most changes should not be particularly evident to non-keyboard users.  However, updated regions should be tested to ensure no unexpected behavior has been introduced. 

Biggest areas of Impact / Testing

Nexus non admin and admin header navs
Nexus Orgchart Navigator and Orgchart Editor
Nexus OC Admin Panel View
Nexus Setup Wizard View
Portal admin nav
Portal Workflow Editor
Tables using the formGrid js class (eg Report Builder reports)


LEAF_Request_Portal/templates/view_session_expire.tpl
```diff
@@ -13,7 +13,7 @@
 <body>
     <div id="header">
         <div style="cursor: pointer" onclick="window.location='./'">
-            <span style="position: absolute"><img src="images/VA_icon_small.png" style="width: 80px" alt="VA logo and Seal, U.S. Department of Veterans Affairs" /></span>
+            <span style="position: absolute"><img src="images/VA_icon_small.png" style="width: 80px" alt="VA seal, U.S. Department of Veterans Affairs" /></span>
```
jampaul3, line 16: Just curious if this can or should be an environment variable. Seems to be in a lot of different places. If it were a global it could have been changed once and done.

jampaul3 (approved):

Pelentan (approved):

LEAF_Request_Portal/templates/view_session_expire.tpl
aerinkayne: Meant to reply to this.  If they are finally ok with the alt text it might not be a bad idea.

PR #2340, opened 2024-04-04
Test refractor

nk2136:
I have created git hub test actions to trigger test from git hub actions, I need to test it now if this works.


.github/workflows/run-automated-test.yml
```diff
@@ -0,0 +1,28 @@
+name: Run automated tests on push to master
+
+on:
+  push:
```
Pelentan, line 4: You should be able to remove the "branches" designation all together so it will run on any branch that gets pushed.  This will allow you to test your Actions without waiting for it to be merged into master.

Pelentan (changes requested):

jampaul3 (approved):

PR #2351, opened 2024-04-11
Hotfix/form grid issue s72c1

aerinkayne:
This update resolves some display issues caused by the introduction of a button related to header sorting.
The name value used to construct the aria-label will have html removed, and double quotes removed.
Also, not all headers are sortable: Those that are not will not contain a button element.

CSS targeting the new button has also been more broadly applied to resolve unintended style changes to the formGrid headers.  This update also syncs/brings up to date the version of formGrid in the app dir with the one in the portal/js dir, so the diff for the app file will show some additional changes.

**Impact**

Headers for formGrid specific output.  Some examples include the homepage search results, Report Builder reports, built-in LEAF secure form display, portal/admin Account Updater page, Nexus Summary page (lower most table)

Primary issue uncovered can be reproduced on the homepage by adding the following header line to a custom view_search template (function renderResult):

{name: '<input type="checkbox">', indicatorID: 'test code', editable: false},

Related styling issues can be reproduced or noticed by setting Report Builder label colors or visiting the Account Updater or Nexus Summary page

**Testing**

Fixes:
-buttons should not exist at all on non-sortable headers (Account Updater) but headers should still tab
-aria-label should not be truncated (dev tools -> header button element attribute),
-button (active sorting element) should have the same height as the header th
-buttons should inherit their color, background color and font size (Report Builder edit labels)
-Report Builder Virtual headers (fixed scrolling headers) should be aligned with the table cells

### DIFF
--- a/LEAF_Nexus/css/style.css
+++ b/LEAF_Nexus/css/style.css
@@ -438,6 +438,20 @@ table#tform {
     empty-cells: show;
     text-align: left;
 }
+table {
+    height: 1px; /* table will ignore, needed to set other content to 100% height */
+}
+table .btn_leaf_grid_sort {
+    cursor: pointer;
+    padding: 4px 2px;
+    margin: 0;
+    border: 0;
+    width: 100%;
+    height: 100%;
+    background-color: inherit;
+    color: inherit;
+    font-size: inherit;
+}
 
 table#tform th {
     border: 0;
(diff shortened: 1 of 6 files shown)


jampaul3 (approved):

Pelentan (approved):

KCN8 (approved):

jampaul3 (approved):

jampaul3 (approved):

Pelentan (approved):

aerinkayne:
Closing this one until all updates can be tested in more detail.
Sorting for custom header content that is set in callbacks will break sorting for the column unless the headerID is moved.  This will introduce additional complexity and potentially cause errors in some customized files.

PR #2356, opened 2024-04-15
Leaf 4291 - display workflow on creation

jampaul3:
**Summary:**

Create a workflow, the workflow that should be shown to the user should be the one created. Currently it will redirect to the first workflow in the list.

There was a change made to deal with the vue version of the workflow editor where url params are used to navigate to workflows. But with the current workflow editor the workflowID is passed to a function.

**Potential Impact:**

when the url param was introduced creating a new workflow, duplicating and renaming the workflow broke the display after save. It would always go back to the first workflow listed because the workflowID was not passed in the url. This ticket should fix that and pass the workflowID as a url param.

**Testing:

New workflow**

Navigate to the workflow editor

click New Workflow

supply a name for this workflow

click save

the workflow that is displayed should be the one just created. (check the dropdown to ensure that it is correct)

**Rename Workflow**

Navigate to the workflow editor

select a workflow to rename (preferably not the first one listed)

click Rename Workflow

change the name

click save

the same workflow should be displayed only the name should have changed (check the dropdown to ensure that it is correct)

**Copy Workflow**

Navigate to the workflow editor

select a workflow from the dropdown to copy

click Copy Workflow

supply a name for this workflow

click save

the workflow that is displayed should be the one just created. (check the dropdown to ensure that it is correct)

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_workflow.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_workflow.tpl
@@ -79,7 +79,10 @@
             let workflowID;
 
             postWorkflow(function(workflow_id) {
-                loadWorkflowList(workflow_id);
+                let url = new URL(window.location.href);
+                url.searchParams.set('workflowID', workflow_id);
+                window.history.replaceState(null, null, url.toString());
+                loadWorkflowList();
                 dialog.hide();
             });
         });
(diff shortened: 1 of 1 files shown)


Pelentan (approved):

SoftwareRenderer (approved):

mgaoVA (approved):

nk2136 (approved):

nk2136:
QA This bug ticket is working as expected, After creating the new workflow, the newly created workflow is displayed at the top of the list. When user selects any exisiting workflow from the list and rename the workflow, the edited workflow is shown at the top of the list.

shaneodd (approved):

PR #2369, opened 2024-04-24
LEAF 4315 checkpoint date headers edgecase

aerinkayne:
Covers an edge case in Report Builder where headers associated with Step fulfillment dates (Checkpoint Dates) continue to display the generic 'Checkpoint Date' text.

It adds a call to /steps to get the information if it's still needed after grid render.
However, this call should not typically be needed (the issue occurs if data for that step does not exist for any records returned by the query), so it should not make much impact.

**Impact/Testing**
Report Builder field selection.
Step 2, top right column, select options for Checkpoint Dates.
Headers should update to show the step name after generating a new report, or after opening an existing report using a short url (this is obtained by clicking the 'share report' button).

### DIFF
--- a/LEAF_Request_Portal/templates/view_reports.tpl
+++ b/LEAF_Request_Portal/templates/view_reports.tpl
@@ -547,12 +547,12 @@ function loadSearchPrereqs() {
             //toggle all subcheckboxes with parent indicator checkbox
             $('.indicatorOption > label > input').on('change', function() {
                 const indicatorIsChecked = this.checked;
-                $(`input[id^="indicators_${this.value}_columns"`).prop('checked', indicatorIsChecked);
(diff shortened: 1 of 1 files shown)


shaneodd (approved):

Pelentan (approved):

jampaul3 (approved):

KCN8 (approved):

aerinkayne:
Closing for now after meetup - it was decided that for now this situation is better simply explained that handled with additional logic.  Might possibly circle back.

mgaoVA:
Reopening. For the API call, please add `?x-filterData=workflowID,stepID,stepTitle`.

I have a separate PR which fixes an issue in formSearch.js, related to excess API calls.

PR #2379, opened 2024-04-30
LEAF-4170 - Developer environment and API endpoint testing

shaneodd:
move go test to a verbose model so we can see what tests are running. Other items for this ticket were already taken care of by other PRs.


aerinkayne (approved):

shaneodd:
The change here was a crucial to figuring what was going on with Carries machine I put it into the rc cleanup pr. The default go test will cover up error messages.

PR #2382, opened 2024-05-01
LEAF-4335 Resolve issues with integrated forms

mgaoVA:
This resolves a few issues with integrated forms, such as the LEAF-S Certification form.

Issue 1: The records_dependency table wasn't being updated correctly because integrated forms have a negative workflow ID, which prevented the record from showing up in the Inbox
Issue 2: The Inbox wasn't retrieving form labels correctly because it didn't include forms that were associated with a negative workflow ID.

The new GET parameter `includeStandardLEAF` was added to ./api/workflow/categories, which includes standard integrated forms in its output.

This also slightly improves organization of the Inbox's "Organize by Roles" view, so that sections are organized by a unique identifier that combines the dependencyID and stepID.

### Potential Impact
- Submitting new records

### Testing
- [ ] There are no issues submitting records
- [ ] Submitted records show up correctly for reviewers
- [ ] This Procedure works:
  1. Navigate to Admin Panel -> Site settings
  3. Submit a LEAF-S Certification
  4. Navigate to the Inbox
  5. The Certification request should show up in the Inbox

### DIFF
--- a/LEAF_Request_Portal/sources/Form.php
+++ b/LEAF_Request_Portal/sources/Form.php
@@ -1403,8 +1403,9 @@ class Form
                         LEFT JOIN workflow_steps USING (workflowID)
                         LEFT JOIN step_dependencies USING (stepID)
                         WHERE recordID=:recordID
-                        AND count > 0
-                        AND workflowID > 0';
+                            AND count > 0
+                            AND workflowID != 0
+                        GROUP BY dependencyID';
 
                 $res = $this->db->prepared_query($sql, $vars);
 
--- a/LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
+++ b/LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
@@ -299,7 +299,7 @@
             // index by roles
             for(let depID in dataInboxes[sites[i].url][j].unfilledDependencyData) {
                 let uDD = dataInboxes[sites[i].url][j].unfilledDependencyData[depID];
-                let roleID = depID.concat(dataInboxes[sites[i].url][j].stepID).toString();
+                let roleID = String(depID) + dataInboxes[sites[i].url][j].stepID;
                 let description = uDD.description;
                 if(roleID < 0 && uDD.approverUID != undefined) { // handle "smart requirements"
                     roleID = Sha1.hash(uDD.approverUID);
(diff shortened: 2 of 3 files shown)


LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
```diff
@@ -299,7 +299,7 @@
             // index by roles
             for(let depID in dataInboxes[sites[i].url][j].unfilledDependencyData) {
                 let uDD = dataInboxes[sites[i].url][j].unfilledDependencyData[depID];
-                let roleID = depID;
+                let roleID = depID + dataInboxes[sites[i].url][j].stepID;
```
Pelentan, line 302: This concerns me.  If both ids are ints, you could get a sum back.  And when I first scanned this, it's what I expected.  But pretty sure it's supposed to be a concat.  So, would probably be better as depID.concat('',  dataInboxes[sites[i].url][j].stepID)    Just need to make sure both are strings.  This is something that shouldn't be a problem now if at least one is a string, but could open up issues later on.  At the very least, it reduces confusion down the road.

Pelentan (changes requested):

LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
mgaoVA: Nice catch. Updated to make sure it's a string.

jampaul3 (approved):

Pelentan (approved):

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1403,8 +1403,9 @@ public function doSubmit(int $recordID): array
                         LEFT JOIN workflow_steps USING (workflowID)
                         LEFT JOIN step_dependencies USING (stepID)
                         WHERE recordID=:recordID
-                        AND count > 0
-                        AND workflowID > 0';
+                            AND count > 0
+                            AND workflowID != 0
```
shaneodd, line 1407: Is there a special use case that I am not seeing for workflow id 0?

shaneodd (changes requested):
Got a question about workflowID != 0 I am trying to see what special case would require that to still exist.

LEAF_Request_Portal/sources/Form.php
mgaoVA: workflowID == 0 means a workflow is not associated with the form.

Looking at this more broadly, it might be safe to remove it but I didn't want to expand the scope of this PR. In a separate story, it might be better to check for workflowID != 0 earlier in the function and prevent submission if the workflow was decoupled from the form.

shaneodd (approved):

aerinkayne (approved):

PR #2385, opened 2024-05-02
LEAF 4337 Simplify ./api/form/specified structure

mgaoVA:
Instead of wrapping content into an arbitrary structure, HTTP status codes should be used instead.

Note that the fetch() related to this already implements a catch(), capturing unsuccessful responses.

### Prerequisite
- The compiled Vue code may conflict with other PRs. After merging, `npm run build` should be executed to recompile it.

### Potential Impact
- This changes the response structure of `api/form/specified`, but it's apparently only used in one place.

### Testing
- [ ] `api/form/specified` isn't used anywhere else
- [ ] Form Preview works normally when the form contains a stapled form


aerinkayne (approved):

Pelentan (approved):

mgaoVA:
Note added prerequisite

jampaul3 (approved):

shaneodd (approved):

aerinkayne (changes requested):
Updating this, since the key change is here too.
We will need to address the sorting issue along with the numeric keys change.

aerinkayne (approved):
Approving with comment to make sure this goes along with 4336, which now has the sorting update

PR #2420, opened 2024-05-29
LEAF 4369 Report Builder stopword issues

aerinkayne:
Associated with INC33091103, user reported that a search using CONTAINS returned no results for a known entry.

The addition of '+' to words in the default stopword list (~36 non-indexed words or those less than 3 chars) for a MATCH ALL query makes them required despite being unsearchable.  Preventing them from being required helps resolve an issue where a search can return no results even when the entry contains the terms.  This change is specific to the CONTAINS (MATCH ALL) search.

Existence of an apostrophe was initially suspected to be the issue.  This did not affect Boolean searches but was found to cause issues with = and fragment searches.  htmlentities call was added for non-Boolean searches to improve hits for encoded data field entries.

**Testing / Impact**

Report Builder data field search filters

data field CONTAINS

Including words like 'a the in' should not prevent a hit.
eg if a textarea entry is: an apple on a tree
-searching for that should return the result.
-searching for the quoted phrase should also continue to return the result.

data field = or data field HAS FRAGMENT
eg if text field entry is: singin' in the rain
-searching = or HAS FRAGMENT should return the result


Pelentan (changes requested):
Looks good but needs Go tests

shaneodd (approved):

jampaul3 (approved):

KCN8 (approved):

PR #2423, opened 2024-05-31
Leaf 3135 - email priors on cancel

jampaul3:
Summary:
An email should be sent to notify people associated with prior steps

- The email should state the request has been cancelled, the reason for cancellation (if any), and a link to the request
- If a request has not been submitted, the email is not necessary

Potential Impact:
This should only affect cancelled requests.

Testing:
Setup a request that can be cancelled.
Cancel the request, (entering a comment as well as not entering a comment)
Observe that an email went to the previous recipients of emails from this request (check the view history to find a list of those previously email)
The email should have the comment if one was supplied.

### DIFF
--- a/LEAF_Request_Portal/sources/Email.php
+++ b/LEAF_Request_Portal/sources/Email.php
@@ -776,7 +776,7 @@ class Email
             $comments = $this->getDeletedComments($recordID);
 
             $comment = $comments[0]['comment'] === '' ? '' : 'Reason for cancelling: ' . $comments[0]['comment'] . '<br /><br />';
-            error_log(print_r($comments, true));
+
             $title = strlen($recordInfo[0]['title']) > 45 ? substr($recordInfo[0]['title'], 0, 42) . '...' : $recordInfo[0]['title'];
 
             $this->addSmartyVariables(array(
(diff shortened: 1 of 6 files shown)


shaneodd (approved):

Pelentan (approved):

KCN8 (approved):

aerinkayne (approved):

jampaul3:
@Pelentan I do see a test for the canceling of a request.

Pelentan (approved):

LEAF_Request_Portal/sources/Email.php
```diff
@@ -777,10 +771,91 @@ function attachApproversAndEmail(int $recordID, int $emailTemplateID, mixed $log
...
+
+        return $return_result;
+    }
+
+    private function processPriorStepsEmailed(array $email_recipients): void
+    {
+        $recipient_list = array();
+
+        foreach ($email_recipients as $recipient) {
+            $clean_list = str_replace('Recipient(s): ', '', $recipient['recipients']);
```
mgaoVA, line 849: Since we're introducing a dependency on the recipient string, there should be a comment on line ~369 where it's assigned to explain that the string has a specific format and shouldn't be changed.

Ideally the list of recipients should be maintained independent of any other text. Addressing this can be scoped to another story.

mgaoVA (changes requested):

LEAF_Request_Portal/sources/Email.php
jampaul3: Comment added. I also created a Jira ticket #4388 to address the ideal situation.

mgaoVA (approved):

PR #2427, opened 2024-06-03
Leaf 4386 - Report Builder: Add example in the JSON dialog

mgaoVA:
This updates the JSON dialog in the Report Builder:
- Remove 10,000 record limit
- Added "JavaScript Template" example
- "Copy to clipboard" button

### Potential Impact
Minimal because the changes are limited to the JSON dialog area.

### Testing
- [ ] Report builder functions normally
- [ ] "JavaScript Template" shows up correctly
- [ ] The "Copy to clipboard" button works

### DIFF
--- a/LEAF_Request_Portal/templates/view_reports.tpl
+++ b/LEAF_Request_Portal/templates/view_reports.tpl
@@ -896,9 +896,6 @@ function showJSONendpoint() {
         selection.addRange(range);
         return selection;
     }
-    document.querySelector('#exportPathContainer').addEventListener('click', (e) => {
-        selectExample();
-    });
 
     document.querySelector('#copy').addEventListener('click', () => {
         navigator.clipboard.writeText(selectExample().focusNode.innerText);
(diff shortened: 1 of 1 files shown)


shaneodd (approved):

jampaul3 (approved):

LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -860,15 +864,45 @@ function showJSONendpoint() {
         switch($('#format').val()) {
             case 'json':
                 $('#exportFormat').html('');
+                document.querySelector('#exportPath').style.display = 'inline';
+                break;
+            case 'leafFormQuery':
+                let buffer = "&lt;script&gt;<br />";
+                buffer += `async function main() {<br />
```
Pelentan, line 871: I'm very uncomfortable with this.  Inserting code via an html "pipeline" is frowned upon for many reasons.  One of the big reasons that jumped out at me on this one is "readability".  When I'm looking over code, I'm generally only glancing at things that are html related.  So I almost just skipped this section.  

One thing that jumps out at me is it looks like it may run but doesn't actually change anything.  The "query" is created but not used. 

Anyway, I'm thinking a better way to handle things like this is to build the function out in another section of this file, then just have this insert the call to it.

Pelentan (changes requested):

LEAF_Request_Portal/templates/view_reports.tpl
mgaoVA: This ends up being plain-text shown in a text area, not executed code. The idea is that after running a report, someone can easily copy/paste the code template for use in the programmer area.

I pushed another commit for this; it looks equally sketchy at first glance, but at least the "innerText" should make it more obvious that it's not executable.

Pelentan (approved):

PR #2424, opened 2024-06-01
Leaf 4381 - Optimize formGrid.js rendering

mgaoVA:
This improves rendering/sort performance by about 70% on my machine when using the test page below.

Subsequent sorts also load about 90% faster (1.9s vs 118ms).

Optimizations:
- Replace jQuery with vanilla JS in hotspots
- Moved [this querySelector so that it runs before](https://github.com/department-of-veterans-affairs/LEAF/compare/rc/2024-05-29/Sprint-75-c2...formGrid-ux?expand=1#diff-44a38b730a68b0a6678696a5a7b27c75fee777b14ae493226ebbc623db0c2c48R370) adding potentially millions of DOM elements.
- Moved JS assigned styles to the CSS stylesheet
- Cache rendered callbacks
- Avoid rendering records outside of the current viewport

This also includes minor updates for the Report Builder:
- Use formQuery.js's batching feature instead of its own
- Use formQuery.js's setAbortSignal() method to improve the reliability of the "Stop" button
- Resolve issue with "Create Row" button showing up when it shouldn't
- Renders the first set of records while loading the remaining data in the background

Before (2,288ms):
Screenshot 2024-06-01 101546

After (633ms):
Screenshot 2024-06-02 161225

Test page simulating a large report:
- To reproduce the test: Load this page, and click one of the headers to sort.
```html
<style>
#content {
    margin: 1rem;
}
</style>
<script>
async function main() {
    // Generate random data
    let numCols = 8;
    let data = {};
    for(let i = 0; i < 10000; i++) {
        data[i] = {};
        for(let j = 0; j < numCols; j++) {
            data[i][j] = Math.random() * 100000;
        }
    }

    let formGrid = new LeafFormGrid('grid');
    formGrid.enableToolbar();
    formGrid.setData(Object.keys(data).map(key => {
        data[key].recordID = key;
        return data[key];
    }));
    formGrid.setDataBlob(data);

    let headers = [];
    for(let j = 0; j < numCols; j++) {
        headers.push({name: `Col${j}`, indicatorID: `Col${j}`, editable: false, callback: function(data, blob) {
            // Simulate old report builder using jQuery
            /*
            $('#'+data.cellContainerID).html(blob[data.recordID][j]);
            $('#'+data.cellContainerID).on('click', function() {
                console.log(blob[data.recordID][j]);
            });
            */

            // Simulate new report builder without jQuery
            document.querySelector(`#${data.cellContainerID}`).innerHTML = blob[data.recordID][j];
[...]

PR #2429, opened 2024-06-05
formGrid.js 508 compliance

mgaoVA:
Separating this for better visibility, see https://github.com/department-of-veterans-affairs/LEAF/pull/2424


jampaul3 (approved):

mgaoVA:
Benefit of this can't be reproduced

PR #2435, opened 2024-06-11
LEAF  4373 (part1) - posting to data metadata field

aerinkayne:
Posts to new metadata field for orgchart employee format questions and outlines an initial related Go test.
After some additional discussion it was decided to stay with json data type.

Reading from the metadata field will be addressed by a future deployment to allow for more concentrated testing.

**Testing / Impact**

-Confirm that metadata is posted to portal table data.metadata for orgchart_employee format questions.
Information fields posted are: userName, firstName, middleName, lastName, email

-Confirm that API tester tests pass


shaneodd (approved):

jampaul3 (approved):

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1126,20 +1126,37 @@ private function writeDataField($recordID, $key, $series)
...
+        $validMetadataFields = array(
+            "firstName" => 1,
+            "lastName" => 1,
+            "middleName" => 1,
+            "email" => 1,
+            "userName" => 1,
+        );
+        $metadata = array();
+        if(isset($_POST[$key . "_metadata"])) {
+            $postMetadata = json_decode($_POST[$key . "_metadata"], true);
```
mgaoVA, line 1138: For backward compatibility, we cannot rely on the user providing this via HTTP POST.

See line ~1099 where logic should be included to handle orgchart_employee formats. Since this format typically only writes the empUID, we'll need to create the metadata based on a server-side empUID lookup.

From the API perspective, interacting with the metadata must be completely transparent to ensure this doesn't conflict with customizations.

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1126,20 +1126,37 @@ private function writeDataField($recordID, $key, $series)
...
                       ':series' => $series,
                       ':data' => trim($_POST[$key]),
+                      ':metadata' => count($metadata) > 0 ? json_encode($metadata) : null,
                       ':timestamp' => time(),
                       ':userID' => $this->login->getUserID(), );
 
-        $this->db->prepared_query('INSERT INTO data (recordID, indicatorID, series, data, timestamp, userID)
-                                            VALUES (:recordID, :indicatorID, :series, :data, :timestamp, :userID)
-                                            ON DUPLICATE KEY UPDATE data=:data, timestamp=:timestamp, userID=:userID', $vars);
+        $this->db->prepared_query('INSERT INTO data (recordID, indicatorID, series, data, metadata, timestamp, userID)
```
mgaoVA, line 1153: The action_history table will also need a `userMetadata` column

mgaoVA (changes requested):
Reading/Writing operations relating to metadata should be split into two deployments to make sure there's no situation where data is partially updated.

First Deployment:
1. Update all DB write operations so that writes to orgchart_employee formatted fields also writes the metadata
    - Also update action_history
2. Populate metadata for existing records

Second Deployment:
1. Prereq: Make extra sure that the metadata is correct
2. When reading from a orgchart_employee field, it should read directly from the metadata instead of performing an employee lookup.

jampaul3 (approved):

KCN8 (approved):

PR #2441, opened 2024-06-13
Leaf 4401 - File manager: last modified date

mgaoVA:
This adds a "Last Modified" column to the File Manager.

API Change:
- `./api/system/files` now accepts the optional GET parameter `getLastModified`. When set, it includes the last modified time for each file.

## Potential Impact
- Impact due to API change mitigated via use of backward compatible $_GET flag
- CSS style for leaf_grid was updated in the admin area. This might actually indirectly fix styling for the LEAF Library, which can only be checked in pre-prod.

## Testing
- [ ] Files and their last modified dates show up correctly in the File Manager

### DIFF
--- a/LEAF_Request_Portal/sources/System.php
+++ b/LEAF_Request_Portal/sources/System.php
@@ -589,8 +589,14 @@ class System
         return 1;
     }
 
-
-    public function getFileList($getLastModified = false)
+    /**
+     * getFileList retrieves filenames uploaded via Admin Panel -> File Manager
+     *
+     * @param bool getLastModified If set to true, the returned elements within the array include [file, modifiedTime]
+     * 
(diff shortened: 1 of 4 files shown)


jampaul3 (approved):

shaneodd (approved):

LEAF_Request_Portal/sources/System.php
```diff
@@ -590,7 +590,7 @@ public function setNationalLinkedPrimary()
     }
 
 
-    public function getFileList()
+    public function getFileList($getLastModified = false)
```
Pelentan, line 593: Please update with typehinting and return type.

Pelentan (changes requested):

PR #2448, opened 2024-06-14
Leaf 4406 - Custom Email Recipient List

jampaul3:
Summary:
A portal had entered recipients into a custom email template and separated them with a ;. This caused the email to not get sent. This fix prevents ; : and space.

Potential Impact:
Emails will go out when expected and won't get hung up on an improperly formatted recipient list

Testing:
Navigate to the email template
select a template and enter recipients in the To and CC fields, try entering any of the characters above, they will be changed to commas.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/mod_templates_email.tpl
+++ b/LEAF_Request_Portal/admin/templates/mod_templates_email.tpl
@@ -240,6 +240,17 @@
     const allowedToCcFormats = {
         "orgchart_employee": 1,
     }
+
+    $(function(){
+        $('#emailToCode').on("keyup", function(e) {
+            this.value = this.value.replace(/[;:\s]/, ',');
+        });
+
+        $('#emailCcCode').on("keyup", function(e) {
+            this.value = this.value.replace(/[;:\s]/, ',');
+        });
+    });
+
     /**
     * Force show or hide the right nav despite screen width
     * @param {bool} showNav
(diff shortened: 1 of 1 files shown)


Pelentan (approved):

LEAF_Request_Portal/admin/templates/mod_templates_email.tpl
```diff
@@ -240,6 +240,17 @@
     const allowedToCcFormats = {
         "orgchart_employee": 1,
     }
+
+    $(function(){
+        $('#emailToCode').on("keyup", function(e) {
+            this.value = this.value.replace(/[;:\s]/, ',');
```
mgaoVA, line 246: The instructions say to place an address "one per line". Commas should not be used.

mgaoVA (changes requested):
There's a few things to pay attention to:
1. Our instructions say to place an address "one per line"
2. The general convention is to use a semicolon to separate addresses


To improve usability:
1. The interface should be updated to make it more obvious when distinct emails have been successfully added (e.g. See Outlook or Gmail)
2. The semicolon convention should be maintained
    - We shouldn't change the storage format behind the scenes

shaneodd (approved):

aerinkayne (approved):

KCN8 (approved):

LEAF_Request_Portal/admin/templates/mod_templates_email.tpl
jampaul3: Modified the code to separate each email with a new line character to match our instructions. Also, made it so that you can't have more than one blank line.

PR #2449, opened 2024-06-14
LEAF 4402 update gridsplitter to use formQuery

aerinkayne:
Updates the grid-splitter report to use formQuery methods to query the selection rather than getting data for each record.
Minor style updates.

**Testing / Impact**

LEAF Programmer built-in report: LEAF_table_input_report

Confirm that grid format question entries for submitted requests can be split into columns and downloaded

### DIFF
--- a/LEAF_Request_Portal/templates/reports/LEAF_table_input_report.tpl
+++ b/LEAF_Request_Portal/templates/reports/LEAF_table_input_report.tpl
@@ -1,64 +1,61 @@
 <style type="text/css">
-    div#dataFieldContainer, div#progress, button#download {
+    #bodyarea {
+        padding: 1rem;
+        min-height: 100vh;
+        font-size:14px;
+    }
+    #tableInputExport {
+        display: flex;
+        flex-direction: column;
+        gap: 1rem;
+        align-items: flex-start;
+    }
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/templates/reports/LEAF_table_input_report.tpl
```diff
@@ -70,126 +67,74 @@ function populateIndicators(categoryID)
...
-        let dataRows = data[indicatorID]['value']['cells'];
-        if (dataRows !== null && dataRows !== undefined) {
-            for (let i = 0; i < dataRows.length; i++) {
-                bodyObj[recordID].push(dataRows[i]);
-            }
-            updateProgress();
-        } else {
-            //if the data for this request isn't in table format, just mark as processed
-            updateProgress();
+function addRecordData(recordID, indicatorID, gridInput = {}) {
```
Pelentan, line 109: Do we want some defaults here?  Additionally, is indicatorID even being used?  If it's here for legacy sake, we may want to indicate that.

Pelentan (changes requested):

jampaul3 (approved):

shaneodd (approved):

Pelentan (approved):

PR #2453, opened 2024-06-20
Leaf 4399 - form editor links and form import behavior

aerinkayne:
Updates the links on the LEAF Library page to navigate to the new form editor.

Minor style changes to the import form dialog:  Drop zone is more obvious, button text is updated, and there is additional feedback during import.

User is navigated to the imported form after import.

**Testing / Impact**

-links on Form Library page should go to the new Form Editor
-Importing a form should nav user to that form.  An import of a packet from the form editor can be tested locally, but a library download will need to be tested on preprod.
-old form editor still behaves as it used to.


LEAF_Request_Portal/admin/templates/view_form_library.tpl
```diff
@@ -33,15 +33,15 @@ function showPreview(recordID) {
...
-    	$.ajax({
-    		type: 'POST',
-    		url: 'ajaxIndex.php?a=importForm',
-    		data: { formPacket: JSON.stringify(preview.getRawForm()),
-    			    formLibraryID: recordID},
-    		success: function() {
-    			window.location = '?a=form&referFormLibraryID=' + recordID;
-    		}
-    	});
+        $.ajax({
```
Pelentan, line 36: Needs a fail catch

Pelentan (changes requested):

Pelentan (approved):

LEAF_Request_Portal/admin/templates/view_form_library.tpl
shaneodd: I would be curious how this works since error is removed according to the docs in jquery 3 which is what we are using from what I can see. https://api.jquery.com/Jquery.ajax/

aerinkayne: The method on the ajax object is removed and replaced with fail.  eg ajax({options}).done().fail().always()
The config option in the setup is still error.  ajax({success, error})
This isn't confusing at all and is why there have been fails sneaking into the config (where it gracefully does nothing).

shaneodd (approved):

PR #2463, opened 2024-07-02
LEAF-4421 Simplify formGrid and example

mgaoVA:
This simplifies formGrid.js usage and updates the example.

Before:
```js
    formGrid.setData(Object.keys(queryResult).map(key => queryResult[key]));
    formGrid.setDataBlob(queryResult);
```

After:
```js
    formGrid.setDataBlob(queryResult);
```

To maintain backward compatibility: If setData() was not previously called, setDataBlob() will apply the same transformation from before.

## Potential Impact
Dependencies of formGrid.js

## Testing
- [ ] Homepage and Report builder works normally
- [ ] Custom implementations of formGrid.js such as LEAF Data Dictionary works correctly


jampaul3 (approved):

shaneodd (approved):

LEAF_Request_Portal/js/formGrid.js
```diff
@@ -839,22 +840,40 @@ var LeafFormGrid = function (containerID, options) {
...
    * @memberOf LeafFormGrid
    */
   function setDataBlob(data) {
     dataBlob = data;
+    if(!usingSetData) {
+      if(Array.isArray(data)) {
+        currentData = data;
+      }
+      else {
+        currentData = Object.keys(data).map(key => data[key]);
```
Pelentan, line 870: What happens when "data" is a string or null?

app/libs/js/LEAF/formGrid.js
```diff
@@ -861,20 +840,40 @@ var LeafFormGrid = function (containerID, options) {
...
    * @memberOf LeafFormGrid
    */
   function setDataBlob(data) {
     dataBlob = data;
+    if(!usingSetData) {
+      if(Array.isArray(data)) {
+        currentData = data;
+      }
+      else {
+        currentData = Object.keys(data).map(key => data[key]);
```
Pelentan, line 870: This function looks identical the the one I commented on above.  Do they need to be pulled out an put in a different file for both to reference one source?

libs/js/LEAF/formGrid.js
```diff
@@ -861,20 +840,40 @@ var LeafFormGrid = function (containerID, options) {
...
    * @memberOf LeafFormGrid
    */
   function setDataBlob(data) {
     dataBlob = data;
+    if(!usingSetData) {
+      if(Array.isArray(data)) {
+        currentData = data;
+      }
+      else {
+        currentData = Object.keys(data).map(key => data[key]);
```
Pelentan, line 870: Here it is again.  Normally the rule is if you have the same function that gets used 3 or more times, you want to pull it out to a central location, unless there is a good reason not to.

Pelentan (changes requested):

aerinkayne (approved):

LEAF_Request_Portal/js/formGrid.js
mgaoVA: Updated to add a console error

app/libs/js/LEAF/formGrid.js
mgaoVA: The directory for this file is part of an incomplete migration

libs/js/LEAF/formGrid.js
mgaoVA: The directory for this file is part of an incomplete migration. Ideally it'd be a symlink.

Pelentan (approved):

PR #2469, opened 2024-07-09
LEAF-4427 Update rules for cancelling requests

mgaoVA:
This changes the rules when cancelling requests and updates the relevant email template.
 
     * Only admins should be able to cancel resolved records.
     * Requestors can cancel at any time.
     * People with access to the current step can cancel.
     * 
     * A comment must be provided for cancellation unless it's an unsubmitted request.

Previously, only admins were allowed to cancel submitted records. Cancelling a request in the middle of processing can cause confusion if people are expecting to see the record, however the notification provided via https://github.com/department-of-veterans-affairs/LEAF/pull/2423 mitigates this.

This also changes the dialogController.js .show() method to return a Promise to enable focusing on the cancellation's comment field.

Other changes:
- The email sender for cancellations and reminders was changed to the actual sender instead of "leaf.noreply@va.gov". This makes it obvious who triggered the email.
- getDirectory() was ported into Form.php from FormWorkflow.php to reduce duplicate initializations of VAMC_Directory.
- The Confirm dialog template was changed from `height: 120px` height to `min-height: 120px`.

## Potential Impact
- The new rules can cause confusion, and they need to be explained to the community.

## Testing
[ ] Automated tests pass
[ ] The rules above work
[ ] Dialog boxes work normally

### DIFF
--- a/LEAF_Request_Portal/sources/Form.php
+++ b/LEAF_Request_Portal/sources/Form.php
@@ -772,7 +772,9 @@ class Form
     /**
      * cancelRecord marks a record as cancelled.
      *
-     * Only admins should be able to cancel submitted records.
+     * Only admins should be able to cancel resolved records.
+     * Requestors can cancel at any time.
+     * People with access to the current step can cancel.
      *
      * @param int $recordID
      * @param string $comment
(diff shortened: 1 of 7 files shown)


LEAF_Request_Portal/sources/Form.php
```diff
@@ -784,21 +804,28 @@ public function cancelRecord(int $recordID, ?string $comment = ''): int|string
...
         $vars = array(':recordID' => $recordID);
-        $sql = 'SELECT `submitted`
+        $sql = 'SELECT `userID`, `submitted`, `stepID`
                     FROM `records`
+                    LEFT JOIN records_workflow_state USING (`recordID`)
                     WHERE `recordID` = :recordID';
-        $resIsSubmitted = $this->db->prepared_query($sql, $vars);
+        $res = $this->db->prepared_query($sql, $vars);
+
+        if ($res[0]['submitted'] > 0 && $res[0]['stepID'] == null && !$this->login->checkGroup(1)) {
```
Pelentan, line 813: Two concerns with this if operation.  First, the "else if" does a hard return.  This should just be putting a value to $return_value; each function should only have a single "return()".  Second, both conditionals are pretty complex.  I'm not seeing a final "else" for a catch-all.  I see that $return_value is declared above, but it might be a good idea to include the final "else" for clarity's sake and just in case.

LEAF_Request_Portal/templates/print_form.tpl
```diff
@@ -686,9 +686,23 @@ function doSubmit(recordID) {
...
 
     function cancelRequest() {
+        dialog_confirm.setTitle('Confirm');
+        <!--{if $submitted > 0}-->
+        dialog_confirm.setContent(
+            '<img src="dynicons/?img=process-stop.svg&amp;w=48" alt="" style="float: left; padding-right: 24px" /> Are you sure you want to cancel this request?<br style="clear: both" /><br />The following reason will be emailed as a notification:<br /><textarea id="cancel_comment" style="width: 100%" rows=3></textarea>'
+        );
+        dialog_confirm.setRequired(<!--{$recordID|strip_tags|escape}-->, function() {
+            let comment = document.querySelector('#cancel_comment').value;
+            if(comment.trim() == '') {
```
Pelentan, line 696: As opposed to here where adding the code to give a single return would only muddy things up.  Always exceptions to every "rule".

Pelentan (changes requested):

jampaul3 (approved):

shaneodd (approved):

LEAF_Request_Portal/sources/Form.php
mgaoVA: I understand the theory of single-return, I'm personally not a fan of always using it. In this situation, using a single return would make the actual logic changes harder to review.

For example, note that all the "changes" due to indentations make it harder to visualize here: https://github.com/department-of-veterans-affairs/LEAF/pull/2481/files

Pelentan (approved):

mgaoVA:
Changing back to draft based on new feedback

aerinkayne (approved):

PR #2470, opened 2024-07-10
Leaf-4445 Promote x-filterData in the dev example

mgaoVA:
This includes the x-filterData parameter as an example for developers.

## Potential Impact
No dependencies

## Testing
- [ ] The example code works when run in the LEAF Programmer (observable via network view)

### DIFF
--- a/LEAF_Request_Portal/templates/view_reports.tpl
+++ b/LEAF_Request_Portal/templates/view_reports.tpl
@@ -816,7 +816,8 @@ function showJSONendpoint() {
     delete query.limit;
     delete query.limitOffset;
     let queryString = JSON.stringify(query);
-    let jsonPath = pwd + leafSearch.getLeafFormQuery().getRootURL() + 'api/form/query/?q=' + queryString + '&x-filterData=recordID,'+ Object.keys(filterData).join(',');
+    let xFilterData = '&x-filterData=recordID,'+ Object.keys(filterData).join(',');
+    let jsonPath = pwd + leafSearch.getLeafFormQuery().getRootURL() + 'api/form/query/?q=' + queryString + xFilterData;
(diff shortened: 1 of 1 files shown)


LEAF_Request_Portal/templates/view_reports.tpl
```diff
@@ -816,7 +816,8 @@ function showJSONendpoint() {
     delete query.limit;
     delete query.limitOffset;
     let queryString = JSON.stringify(query);
-    let jsonPath = pwd + leafSearch.getLeafFormQuery().getRootURL() + 'api/form/query/?q=' + queryString + '&x-filterData=recordID,'+ Object.keys(filterData).join(',');
+    let xFilterData = '&x-filterData=recordID,'+ Object.keys(filterData).join(',');
+    let jsonPath = pwd + leafSearch.getLeafFormQuery().getRootURL() + 'api/form/query/?q=' + queryString + xFilterData;
```
Pelentan, line 820: Just a quick check.  Unsure whether I missed this before or just don't remember the answer.  Is pwd accessing the client computer or the server?  If server, it looks like we're sending infrastructure information to the client?  Something that's a no-no.

Pelentan (changes requested):

LEAF_Request_Portal/templates/view_reports.tpl
shaneodd: It looks like it is information that would be present to the client. 

    let pwd = document.URL.substr(0,document.URL.lastIndexOf('?'));

jampaul3 (approved):

aerinkayne (approved):

Pelentan (approved):

PR #2472, opened 2024-07-11
LEAF 4430 script to handle request and WF issues associated with 4363

aerinkayne:
Separate script intended to resolve existing issues for requests in an inconsistent 'returned to requestor state'.

Find and update 'sendback' type workflow_routes that are incorrectly configured with a nextStepID of -1.
Find and update requests in a -1 workflow step state (essentially runs Form->openForEditing on these requests).

**Impact / Testing**

Workflow Editor
- use any workflow, but add a 'return to requestor' route if it does not yet have one.
- open a stepInfo modal by clicking on any step.
- without refreshing the page, copy the workflow (note in Adminer, sendback actionTypes for this workflow have -1 as their next step).
- in the Form Editor, set a form to this workflow.
- submit a request and move it as needed to use the 'return to requestor' action.
- open returned request to see error state.

running this script should:
- set the nextStepID of the misconfigured sendback workflow routes to 0.
- put affected 'returned to requestor' requests in a proper state.

### DIFF
--- a/LEAF_Request_Portal/scripts/checkNextStepID_and_update.php
+++ b/LEAF_Request_Portal/scripts/checkNextStepID_and_update.php
@@ -6,7 +6,8 @@ require_once getenv('APP_LIBS_PATH') . '/../Leaf/Db.php';
 $db = new App\Leaf\Db(DIRECTORY_HOST, DIRECTORY_USER, DIRECTORY_PASS, 'national_leaf_launchpad');
 
 /**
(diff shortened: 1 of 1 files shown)


jampaul3 (approved):

shaneodd (approved):

Pelentan (approved):

aerinkayne:
Closing since this was review only and corresponding update is deployed

PR #2491, opened 2024-07-26
Leaf 4449 - Cleanup of orgchart refresh

jampaul3:
Summary:
The nightly orgchart refresh was causing the employee's empUID to increment even though there wasn't a new record added. It was determined that this was because that table has an auto increment as well as a separate column that was unique. Testing the current script revealed that there were instances where the userName from the national orgchart and local orgchart were spelled differently case wise. This was causing an issue where the empUID from the local orgchart was not being included in the update query. So when MySql hit this record it would increment the db auto increment but when it tried to insert it hit the unique index and found a match and therefore updated that record. The auto increment still increased but would not be noticeable until a new employee was added.
The changes made here make the userName all upper case when looping through so that they will match every time. Running this change resulted in no userNames without an empUID.
Adding deleted_ to the front of userNames that are deleted or absent from the national orgchart is also in this script.

Potential Impact:
The main impact here will be that the db's will not hit the upper limit of the mediumint for the empUID in the employee table. With adding deleted_ to the front of userNames will prevent recycled userNames from being used for the wrong person. A person that has been disabled (deleted_ added to the beginning of their userName) will cause read issues with requests. (Specifically, when initiator is disabled the backup will not be able to take action on their requests)

Testing:
This can not be tested locally, it needs to be pushed to staging and tested there.
When testing, run the script. If there is a problem a file will be created here /var/www/php-logs/refresh.log. If no file is created then everything should be good.


jampaul3:
Removed a commit and can't sync.

PR #2500, opened 2024-08-06
LEAF-4425 prevent built-in email template changes

aerinkayne:
Resolves an issue where custom event descriptions could cause email template collision and allow inadvertent editing or deletion of built-in email_templates records if the description matched the value of an email_templates label.  

A number of other issues and optimizations have also been addressed:

- Front end updated to handle non-error return values when events cannot be created (previously incorrectly alerted that the event was successfully created).

- The alert for a success has been removed.  If an event is created from a step action, the event is added and the actionInfo modal re-opens to show the added event (previously users would need to re-open the modal, select the newly created event and add it).

- Updated system event checking to use prefixes to accommodate potential future built-in templates.

- Event creation and editing methods do not try to parse data associated with "AutomateEmailGroup", "DaysSelected", "DateSelected" and "AdditionalDaysSelected".  These properties are associated with step-specific workflow_steps.stepData not events.eventData.  

- Created Go tests associated with event creation, reserved prefixes and duplicate descriptions

**Testing / Impact.**
Workflow Editor - Creating and Editing events.
Email Template Editor - Display and Editing of custom email templates.

Events are created from either the left side menu (edit events - events modal has its own 'create event' button) or through workflow actions bubbles -> Add Event -> Create Event.

- Confirm events can be created, and that they can be added and removed from workflow actions.
- Confirm duplicate descriptions or those that match built-in email_templates labels are not allowed (user should be alerted).
- Confirm the expected template displays in the Email Template Editor (label is based on the event description).
- Confirm API tests pass


Pelentan (changes requested):
This is changing a function to accept a params variable instead of specific variables.  We need to ensure that no legacy code will be affected by this.  

Please give some detailed testing instructions and plan to get with Nikesh and Margo on the testing.

shaneodd (approved):

Pelentan (approved):

PR #2501, opened 2024-08-07
Leaf 4468 - updated disabled userName

jampaul3:
Summary:
After initial deployment of the refresh orgchart refactor it was observed that we could end up with a db unique key violation should a userName be recycled more than once. This is to ensure that doesn't happen.

Potential Impact:
This addresses the potential of a duplicate userName being attempted to be created. No other impact is known at this time.

Testing:
Testing is not possible on local. Extensive testing will need to be accomplished on staging before implementing on prod.

### DIFF
--- a/app/Leaf/VAMCActiveDirectory.php
+++ b/app/Leaf/VAMCActiveDirectory.php
@@ -240,7 +240,7 @@ class VAMCActiveDirectory
     private function preventRecycledUserName(): void
     {
         $sql = 'UPDATE `employee`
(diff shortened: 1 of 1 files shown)


shaneodd (approved):

aerinkayne (approved):

Pelentan (approved):

jampaul3:
This will not go out for a few weeks most likely so closing this until we are ready to deploy.

PR #2512, opened 2024-08-09
Leaf 4474 - prevent phantom increments

jampaul3:
Summary:
With the Employee table having an auto increment as well as a unique index we had a situation where the auto increment would happen and then the unique index would get updated. The reason for this is because the userName in the national orgchart and the userName in the local orgchart don't always match case. When they don't match case it would cause a situation where the userName didn't look like it was avaliable and therefore no empUID would be supplied. This will remove that scenario by making all userNames be compared as all lower case, ensuring a match and therefore an empUID being supplied to update.

Potential Impact:
I think we have everything covered now. Adding a timestamp to the userName when disabled will ensure that a second recycle of a userName doesn't cause conflicts.

Testing:
Testing is difficult locally. It requires a national orgchart that is different from the local orgchart. Locally they are virtually the same.


Pelentan (approved):

mgaoVA (changes requested):
It's not clear how user accounts get marked as disabled.

jampaul3:
This Leaf ticket was redone to stay within scope and another ticket created to complete everything that was done here.

PR #2514, opened 2024-08-12
LEAF 4455 - reading metadata for form requests (orgchart employee format)

aerinkayne:
Updates orgchart_employee format form questions and report builder cells to use saved metadata from the data table rather than calling lookupEmpUID.

Adds the metadata field values to the API-testing portal db initial boilerplate (this db is otherwise already up to date with table structure).

other: script to update data.metadata added to leaf-scripts.  This is potentially needed for local/dev envs with older volumes of leaf_portal data.  Orgchart employee entries without metadata would otherwise show as disabled accounts.  Pending some further discussion.

Finish associated Go test

**Impact / Testing**

Impact is to orgchart employee format questions on form requests and Report Builder reports.  Impact is high and could cause data for orgchart_employee format questions to display incorrectly in form requests and Report Build reports.

- Confirm orgchart employee format entries display correctly in form requests and in Report Builder reports.

- Run Go tests on branch **issue/LEAF-4478/orgchart_info_go_test** as well as on this one and confirm they pass.  On 4478, orgchart values come from a query of the orgchart, on 4455, they come from the metadata field.

- Confirm employee entries with metadata continue to display if they are subsequently disabled.
Note: entries associated with accounts that were **already disabled** when the metadata update script was run should continue to display disabled account information.


Pelentan (approved):

Pelentan (approved):

shaneodd (approved):

scripts/leaf-scripts/update_data_metadata.php
```diff
@@ -0,0 +1,134 @@
+<?php
```
mgaoVA, line 1: Please rename to update_devenv_data_metadata.php or similar.

This is to improve the distinction between a script meant to be run in development or prod environments, and reduce potential mishaps on the command line.

LEAF_Request_Portal/templates/print_subindicators_ajax.tpl
```diff
@@ -228,7 +228,7 @@
             <!--{if trim($indicator.value) != ''}-->
             <div style="padding: 0px">
                 <!--{if $indicator.displayedValue != ''}-->
-                    <a href="<!--{$orgchartPath}-->/?a=view_employee&empUID=<!--{$indicator.value|escape}-->"><!--{$indicator.displayedValue|sanitize}--></a>
```
mgaoVA, line 231: Any reason why this link can't be retained? This represents a regression in functionality.

Side note: An e2e test for something like this would help highlight potential regressions. @Pelentan

mgaoVA (changes requested):

LEAF_Request_Portal/templates/print_subindicators_ajax.tpl
aerinkayne: The concern for this was that the link is based on the data value / empUID.  If the number is recycled the link would go to an incorrect page.  It had been mentioned and the thinking at that time was that it should be avoided for that reason.

aerinkayne:
Tagging this as 'review / comment only' for now because it was decided that a similar column was also needed for the records and notes tables and that the changeover for highly visible items should be made at the same time.

aerinkayne:
closing until additional database and associated post behavior is updated for records, notes and data_history

PR #2528, opened 2024-08-21
Leaf 4465 - Nexus 508 updates

jampaul3:
Nexus Orgchart navigator and editor

Page needs a number of improvements

Additional info to user on overview page about what the page is conveying and what can be edited (brief ok)

Drag behavior in the editor should be available from keyboard

Tab order at least appears unintuitive.  There is a number associated with each cell that might mediate this.  Tabbing should be tree breadth so presentation alone might resolve this

Doublecheck how / when associated dropdowns collapse - tabbing away and full mouse off should close them

Potential Impact:
These changes should make it easier for people to navigate the orgchart navigator and editor with the keyboard.

Testing:
Open the orgchart and click browse.
using only the keyboard tab through the page. Once you get to the different positions they should tab in hierarchy order, Level 1 to level 2 to level 3 and so on.
If you press enter on the edit orgchart from the navigator page you enter the editor page. 
tabbing through the page again should tab through the hierarchy order. 
There is a new button name move coordinates for each of the positions.
Pressing enter on move coordinates will allow you to use the arrow keys to move the position around the screen.
Pressing Esc will return the position to it's original coordinates and exit the eventListener.
Pressing Enter will save the newcoordinates and exit the eventListener.

### DIFF
--- a/LEAF_Nexus/js/ui/position.js
+++ b/LEAF_Nexus/js/ui/position.js
@@ -5,30 +5,30 @@
 */
 
 function position(positionID) {
-  this.positionID = positionID;
-  this.rootID = 0;
-  this.parentID = 0;
-  this.parentContainerID;
-  this.containerHeader;
+  this.positionID = positionID
+  this.rootID = 0
+  this.parentID = 0
+  this.parentContainerID
+  this.containerHeader
 
-  this.prefixID = "pos" + Math.floor(Math.random() * 1000) + "_";
-  this.data = new Object();
+  this.prefixID = 'pos' + Math.floor(Math.random() * 1000) + '_'
+  this.data = new Object()
 }
 
 position.prototype.initialize = function (parentContainerID) {
-  var t = this;
-  this.parentContainerID = parentContainerID;
+  var t = this
+  this.parentContainerID = parentContainerID
 
-  var prefixedPID = this.prefixID + this.positionID;
-  this.containerHeader = prefixedPID + "_title";
-  var buffer = "";
+  var prefixedPID = this.prefixID + this.positionID
+  this.containerHeader = prefixedPID + '_title'
+  var buffer = ''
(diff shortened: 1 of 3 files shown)


LEAF_Nexus/js/ui/position.js
```diff
@@ -37,171 +37,216 @@ position.prototype.initialize = function (parentContainerID) {
...
 
 position.prototype.prepContent = function (response) {
-  this.data = response;
-  this.setNumFTE(response[11].data);
-  this.setTitle(response.title);
-  var pd = response[9].data != "" ? "PD#" + response[9].data : "";
+  this.data = response
+  this.setNumFTE(response[11].data)
+  this.setTitle(response.title)
+  var pd = response[9].data != '' ? 'PD#' + response[9].data : ''
```
Pelentan, line 136: I'm seeing a whole lot of var's all through this.  Are they required?

LEAF_Nexus/js/ui/position.js
```diff
@@ -37,171 +37,216 @@ position.prototype.initialize = function (parentContainerID) {
...
-  if (response[15].data != "") {
-    var layout = $.parseJSON(response[15].data);
-    $("#" + this.prefixID + this.positionID).css("position", "absolute");
-    var y = 120;
-    var x = 40;
+  )
+  if (response[15].data != '') {
+    var layout = $.parseJSON(response[15].data)
+    $('#' + this.prefixID + this.positionID).css('position', 'absolute')
+    var y = 120
```
Pelentan, line 149: I know they aren't needed here.

Pelentan (changes requested):
Whole lotta changes in here that I'm not sure are needed.  As well as "vars".

jampaul3:
redoing this to keep formatting where it was.

PR #2530, opened 2024-08-21
Rc/2024 08 21/sprint 80 c1

Pelentan:
LEAF tickets:

4425
4106
4464


mgaoVA (changes requested):
Results from the automated test suite:
- Selenium: 5 failed, 15 skipped
- Playwright: 2 failed


Couple highlights for consideration: @department-of-veterans-affairs/leaf-developers-team 
- While resolving issues flagged by tests, keep in mind that the tests represent the expected outcome. If the expectations change and it's not explicit in the original story, we should briefly explain why the test was changed in the commit log or PR. This helps clarify how the test was resolved and can also reveal other issues.
  - E.g. In docker/vue-app/src/form_editor/views/FormEditorView.js:
    - For the "Add Section" button, the original title attribute was incorrect. Instead of "Add new form section", it should've been "Add section".
    - In contrast on the same page, the "Add Question to Section" has the correct title. Note the button label matches the aria-label attribute.
- Maybe `git merge master` or `git rebase master` should be run more frequently within branches

PR #2534, opened 2024-08-26
E2E-Tests:  New Service Request

nk2136:
On this Work Flow , user selects the Simple Form and completes new service Request,  Assertion on most the active elements and actions.

Documented every steps of the workflow.

image


x-test/tests/newServiceRequest.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  await expect(page.getByRole('button', { name: 'Click here to Proceed' })).toBeVisible();
+  await page.getByRole('button', { name: 'Click here to Proceed' }).click();
+
+  // Click on the "single line text" input field and fill it with a response.
+  await page.getByLabel('single line text').click();
+  await page.getByLabel('single line text').fill('test question 1');
+
+  // Click on the "Next Next Question" button to move to the next step of the form.
+  await page.getByRole('button', { name: 'Next Next Question' }).click();
+
```
aerinkayne, line 37: I think using getByRole is a good idea since it gives better insight into how these are likely to be read.  It also shows the need for some agreement on html consistency, clarity on how the name value is determined and some coordination -- especially in situations like this where the name value is already a red flag. 
There are some img tags with attributes that should be removed (like here) but they cannot be removed without also updating the selenium and playwright tests

```html
    <!-- NVDA screen-reader output -->
    
    <!-- 'Test button button' -->
    <button type="button">Test button</button>
    <button type="button" title="Test button">Test button</button>
    <button type="button" aria-label="Test button">Test button</button>
    <button type="button" aria-label="Test button" title="Test button">Test button</button>
    
    <!-- 'this is what the screen reader says button' -->
    <button type="button" aria-label="this is what the screen reader says" title="this is what the screen reader says">Test button</button>
    
    <!-- 'this is what the screen reader says button and usually this too if it's different' -->
    <button type="button" aria-label="this is what the screen reader says" title="and usually this too if it's different">Test button</button>
    
    <!-- 'Test button button Test' -->
    <button type="button" title="Test">Test button</button>
    
    <!-- 'Next graphic Next Question button ' -->
    <button id="nextQuestion2" type="button" class="buttonNorm nextQuestion"><img src="dynicons/?img=go-next.svg&amp;w=22" alt="" aria-label="Next"> Next Question</button>```

mgaoVA (changes requested):
1. Pull Requests must explain changes, especially reasoning if there's no existing documentation (e.g. JIRA story). This PR makes changes to folder structure and configuration without explanation.

2. The change to the folder structure negatively impacts organization, and creates a `gitignore` issue. Note that the previous structure provided clear separation between API and End-to-end test suites. This new structure makes it harder to see where the boundaries are.

3. Test results should never be committed to the repository. They'd be immediately overwritten, it takes unnecessary space, and makes it more difficult to review real changes.

4. The Playwright config is meant for automated test runs, and these changes should be removed:
    - `workers: 4` - Playwright auto-selects workers based on the number of available CPU cores. 
    - `headless: false` If a UI is needed, please refer to: https://playwright.dev/docs/running-tests#run-tests-in-ui-mode

5. Tests should avoid duplicating existing tests. For example, creating a new request has already been done, but commenting has not. So adding a new test that exercises the comment system would be appropriate. @Pelentan Maybe PRs for new tests should also explain why it's unique?

6. The filename of the test should reflect the page or feature it's testing. For example, the section relating to comments is on the page where you're viewing the record, so `viewRecord.spec.ts` would be more appropriate. If the test involves a broader scope, we should prefix it with a word (for example `lifecycle`).

7. If there's a need to reference unique text, either generate random numbers or use a more unique fragment. For example this test includes a direct reference to `new request test`. This can potentially conflict with other text because it's very generic. A more unique fragment might be `new request test nk20240827`.

8. It's not necessary to define `ignoreHTTPSErrors` because this is already defined in the global configuration.

PR #2536, opened 2024-08-27
Leaf 4500 - avoid phantom increments

jampaul3:
Test to avoid phantom increments. Running the refresh on check national employee updated the userNames and set the auto increment to a higher number and any subsequent refreshes wouldn't cause the phantom because the userNames were already updated so there wasn't any mismatched userNames.

Potential Impacts:
The TestEmployee_AvoidPhantomIncrements needs to be completed before TestEmployee_CheckNationalEmployee because they both run refreshOrgchartEmployee and running them out of sequence will cause the national and local orgcharts to be sync'd up and their userNames be updated so they match on both sides.

**Testing:**

Create a new user in National Orgchart
Create a the same new user in Local Orgchart (change the case of the userName)
get these two userNames to ensure they are in the database
compare the userNames - they should be the same except the case should be different
refresh Orgchart Employees
create a new user in local orgchart (different from the the previous users added)
compare the empUID from the first local to the second local (the second local empUID should be 1 higher than the first)


x-test/API-tests/employee_test.go
```diff
@@ -91,6 +92,79 @@ func disableEmployee(postUrl string) error {
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	if empUID1 == "" {
+		t.Error("no user id returned")
+	}
+
+	// run refresh Orgchart
+	err = updateEmployees(RootOrgchartURL + `scripts/refreshOrgchartEmployees.php`)
```
mgaoVA, line 132: go-staticcheck: this value of err is never used

mgaoVA (changes requested):
We should expect this test to fail when its run in an older un-patched version (prior to resolving the phantom increment issue) of LEAF.

Example:
1. git checkout [old version prior to the fix]
2. Copy/Run this test on the old version
3. Test should fail because the old version doesn't contain the fix for phantom increments

x-test/API-tests/employee_test.go
jampaul3: Added the check.

jampaul3:
@mgaoVA  I did exactly what you commented and it did fail. I did it before you said that here because I wanted to be sure the test worked as expected and that my code changes worked as expected.

mgaoVA:
> @mgaoVA I did exactly what you commented and it did fail. I did it before you said that here because I wanted to be sure the test worked as expected and that my code changes worked as expected.

Mind elaborating which branch you used?

I checked it against `rc/2024-08-21/Sprint-80-c1` because it's the first merge that precedes [the fix](https://github.com/department-of-veterans-affairs/LEAF/pull/2522), which was merged in c2.

Based on the procedure in https://github.com/department-of-veterans-affairs/LEAF/pull/2522 it seems like the following is missing in the automated test:
 > at least one userName needs to be spelled the same but with different cases

jampaul3:
> > @mgaoVA I did exactly what you commented and it did fail. I did it before you said that here because I wanted to be sure the test worked as expected and that my code changes worked as expected.
> 
> @jampaul3 Mind elaborating on which branch you used?
> 
> I checked it against `rc/2024-08-21/Sprint-80-c1` because it's the first merge that precedes [the fix](https://github.com/department-of-veterans-affairs/LEAF/pull/2522), which was merged in c2.
> 
> Based on the procedure in #2522 it seems like the following is missing in the automated test:
> 
> > at least one userName needs to be spelled the same but with different cases

@mgaoVA I branched off of master (which should have Sprint-80-c1) and pasted the test into the same file and ran it. It failed for me as expected. I know that there are several userName's spelled the same but with different cases so changing userNames isn't in the test it self. That is why this test needs to be before the Check National Employee. I can try and figure out how to make the userName change in the test, that would make it better for sure to ensure a future update didn't make everything the same on both National and Local orgcharts.

shaneodd (dismissed):

mgaoVA:
@jampaul3 Thanks, after the latest commit, the test behaves as expected.

It's important that tests replicate the conditions that caused the original issue. We should be able to demonstrate that the test fails/passes before and after the fix. I'd have to guess there's a process-related inconsistency since we got different results, and this would either need to be reviewed or an automated process should be developed to simplify it.

The username with different case sensitivity is fundamental in replicating the issue.

From https://github.com/department-of-veterans-affairs/LEAF/pull/2522:
> If they don't match case wise then it would attempt an insert and upon finding a userName that matched (case insensitive at this point) it would update.

mgaoVA (dismissed):

KCN8 (approved):

aerinkayne (dismissed):
Test runs ok if I put it into 4474 but just a heads up git doesn't like an attempt to pull this - not sure if it's something we need to consider or not

Pelentan (dismissed):

PR #2540, opened 2024-08-31
Resolve JS error (dev console approval), Improve tests

mgaoVA:
Note: This does not need to be rolled out as a hotfix, because functionality is not impacted.

This PR:
- Resolves an error message that shows up in the JS console during the LEAF Developer Console approval workflow
  - Added automated test that reproduces the issue, and checked for validity against `rc/2024-08-27/Sprint-80-c2`
  - Also resolves a JS warning on the same page
- Expands the work to mitigate page hydration issues from https://github.com/department-of-veterans-affairs/LEAF/pull/2532
- Reduces test flakiness
  - `x-test/end2end/tests/lifecycleSimpleTravel.spec.ts` now checks the results of a key async request. Previously, the UI would occasionally render two workflows on the screen during repeated test runs.

### Potential Impact
Similar to #2532: 
> pages that utilize [dialog templates] depend on the updated dialogController.js file to be up-to-date on client machines, which may be impacted by cache controls.

### Testing
- [ ] Automated tests pass

### DIFF
--- a/x-test/end2end/tests/lifecycleDeveloperConsoleApproval.spec.ts
+++ b/x-test/end2end/tests/lifecycleDeveloperConsoleApproval.spec.ts
@@ -9,5 +9,9 @@ test('no javascript errors on supervisor selection page', async ({ page }) => {
   await page.goto('https://host.docker.internal/Test_Request_Portal/report.php?a=LEAF_start_leaf_dev_console_request');
   await page.getByRole('button', { name: 'I understand and accept' }).click();
   await page.getByRole('button', { name: 'Next Next Question' }).click();
+
+  // wait for async request to complete
+  await expect(page.locator('label')).toContainText('Supervisor* Required');
+
   expect(errors).toHaveLength(0);
 });
--- a/x-test/end2end/tests/lifecycleDeveloperConsoleApproval.spec.ts
+++ b/x-test/end2end/tests/lifecycleDeveloperConsoleApproval.spec.ts
@@ -1,0 +1,13 @@
+import { test, expect } from '@playwright/test';
+
+test('no javascript errors on supervisor selection page', async ({ page }) => {
+  let errors = new Array<Error>;
+  page.on('pageerror', err => {
+    errors.push(err);
+  });
+
+  await page.goto('https://host.docker.internal/Test_Request_Portal/report.php?a=LEAF_start_leaf_dev_console_request');
+  await page.getByRole('button', { name: 'I understand and accept' }).click();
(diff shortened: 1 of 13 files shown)


shaneodd (changes requested):
For the testing portion would we also want to check if the disabled actually gets removed in these sections? Since that is something that has been added and has code to remove the disabled.

jampaul3 (dismissed):

x-test/end2end/tests/lifecycleDeveloperConsoleApproval.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+  });
+
+  await page.goto('https://host.docker.internal/Test_Request_Portal/report.php?a=LEAF_start_leaf_dev_console_request');
+  await page.getByRole('button', { name: 'I understand and accept' }).click();
+
+  // Wait for load
+  await expect(page.locator('#xhr')).toContainText('This is a request to access the LEAF Developer Console.');
+
+  await page.getByRole('button', { name: 'Next Next Question' }).click();
+
```
aerinkayne, line 16: This button name is part of the 508 updates and will become 'Next Question'.  There is an image element with a 'next' label causing readers to announce both the label and the text.

I think getByRole('button', { name: 'Next Question' }).last().click() should already work now on the other button - last() since after the update there will be two of them.

mgaoVA:
> For the testing portion would we also want to check if the disabled actually gets removed in these sections? Since that is something that has been added and has code to remove the disabled.

This would probably be redundant because the function is shared in the previously updated `dialogController.js` file, where the removal actually happens and has test coverage.

For the end-to-end tests, we'd want to focus on and think in terms of user stories. Creating a test for "As an admin, I should be able to cancel a request" would be a good exercise since it represents a truly new test and would provide supporting evidence that the `disabled` attribute is being removed correctly (since it involves a different dialog template).

shaneodd (dismissed):

x-test/end2end/tests/lifecycleDeveloperConsoleApproval.spec.ts
mgaoVA: Updated. Generally we'd want to leverage Playwright's code generator as much as possible.

aerinkayne (dismissed):

Pelentan (dismissed):

Pelentan (approved):

aerinkayne (approved):

shaneodd (approved):

PR #2543, opened 2024-09-06
Leaf 4501 - Database update

jampaul3:
Summary:
This is to prepare the database to accept a longer userName for those that need to be disabled.

Potential Impact:
The biggest thing is this will allow us to keep users even after they are no longer with the VA and their userName is recycled and given to someone else.

Testing:
Go to the Nexus
Click on the OC Admin Panel
Click on Other Tools
Click on Update Database
Verify in the database that the field size was increased for expected tables.


mgaoVA:
Have the Request Portal tables been updated already?

Pelentan (approved):

jampaul3:
@mgaoVA I changed the nexus fields to varchar(50) explained in the jira ticket why I chose 50. Portal side where ever the userName is stored is already at varchar(50) so there is no need to make a change there.

PR #2546, opened 2024-09-13
LEAF-4487 - large queries local server

shaneodd:
image

This is for the container changes. I broke this out since the large queries change was small and I did not want it and its tests to get squashed by changes here. We can merge them later if we find this too confusing. I would like to discuss how the headers are setup since having that information on the api call is helpful to see where the code was run.

#### Reverse Proxy NGINX
The rpnginx is similar to our production reverse proxy. I tried to just copy what is needed to get this to work. I can make this more like our production server if needed. The proxy_next_upstream is set to handle on 503's I was not able to get this into a custom response code. The next upstream has limited headers that it can work with. https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream

#### NGINX API
The nginx-api is the api that will route to php for the larger queries to run. Below are  settings that I changed that I need to point out. These we will want to discuss what we want for production. About 20 minutes seems to be enough to pull the data on staging. That could change when we have 20 people trying to pull large queries along with the normal traffic to the site and database. Before we start working on production configurations I would like to talk about how we should break that problem down.
```
    fastcgi_read_timeout 3000;
    fastcgi_send_timeout 3000;
```
#### PHP-API
The php-api is the php portion that does the processing. Memory limit is set to 2gb, max requests are set to 1, this helps with memory leak issues (see below). On staging I had this set to 14GB however I am now wondering if I was running into an issue with memory, locally I was having it use 8GB of memory to return 33MB of data. I would then restart the containers then it would use less than 2GB.

www.conf
```
pm.max_requests = 1
```

php.ini
```
memory_limit 2048M
```
### Differences 
Our staging and prod have the nginx and php-fm in one container. Reverse proxy is on bare metal in front of our application server. Here we have the nginx and php containers separated and the reverse proxy server is in a docker container.

### Memory Limit issues
Set your max_request for api and normal php fpm to be be 100. Pull large queries, the large queries container will eventually no longer garbage collect properly and start eating up memory. The memory used will go beyond what php would be set to pull. I have not been able to figure out how to reliably get this to happen, it will just happen. If I set it to 1 or 2 it will clear memory. Setting it to 0 will need more testing, I was hoping to find out how to trigger this before going down a path of figuring out all of these options. 
https://www.php.net/manual/en/features.gc.collecting-cycles.php
https://tideways.com/profiler/blog/what-is-garbage-collection-in-php-and-how-do-you-make-the-most-of-it

image
In this screenshot I ran command `watch free -mh`, when I ran the process used would get close to equaling total and available would be down in the MB before the system locked up.


jampaul3 (dismissed):

aerinkayne (approved):

Pelentan (approved):

mgaoVA (changes requested):
1. Please include a brief summary of the change in the PR's comment as the first line. This will help in the future because all of these PRs are incorporated into reports and changelogs. E.g. "This routes large queries to a specialized high-resource container"

2. The flowchart needs to be updated to clarify how the headers work, and what they specifically are. See https://info.aiim.org/aiim-blog/flowcharting-in-business-process-management for common flowchart conventions. Switching to a sequence diagram might be better overall since there's some back-and-forth: https://mermaid.js.org/syntax/sequenceDiagram.html

3. The headers should be simplified

    This is unclear:
    `LEAF_Large_Queries`
    - pass_onto_large_query_server 
    - process_ran_on_large_query_server
    
    Simplified:
    `LEAF_Is_Large_Query`
    - true
    - false

4. Regarding slow rollout: I think we should rely on our autoscaling infrastructure to handle load.

5. Additional considerations involving monitoring/deployment: The Large Query Container must be monitored as a separate service so that the response time metrics don't interfere with typical queries.

shaneodd:
1. I agree, I have updated that let me know how that looks to you.
2. I have this updated, I am not entirely  happy with it, but that is what I have
3. I was thinking on changing pass_onto_large_query_server -> process_ran_on_api_server. When looking from the front end you will know where this process ran so if it is spitting other errors out you can pinpoint. For me trying to translate true/false into location of where it would run is a bit harder. This was something I have been trying to get some thoughts from Pete on.
4. Yeah this was when the server landscape was different, I replaced/removed it. #5 is what I ended up putting there since I think that fits perfectly.

jampaul3 (approved):

Pelentan (approved):

jampaul3 (approved):

jampaul3 (approved):

Pelentan (approved):

jampaul3 (approved):

PR #2553, opened 2024-09-18
LEAF 4526 (INC35161094) prevent copying of request if form is hidden

aerinkayne:
This issue is associated with a help desk ticket where users were able to create hidden forms.  
Forms had been created through Tools -> Copy Request convenience feature.  

Database update:
Updates the default value for new categories.visible to -1 (Unpublished).  Existing settings are not being changed.

Server/API
Adds logic to Form.php method newForm to first check if any forms posted are set to unpublished before creating the record.

Front:
Makes the new visible selection option ('Unpublished') available in the new and old Form Editor.
Copy Request modal:
Displays an error message instead of creating a new request if the request is associated with unpublished forms.
Addresses some label issues on the request copy modal. 
Change Form(s) modal:
Displays additional information if the form is unpublished.  This modal was updated to match checked status by formID rather than by form name,

**Testing / Impact**

Form Editor - Visible Form setting (form properties panel)

Form View - Request Copy Feature (Tools -> Copy Request).

Confirm that the request copy feature continues to work as expected when no forms are set to 'unpublished'.

Confirm that a warning message displays if forms associated with the request are set to unpublished.

Confirm all form elements in the modal are associated with labels.

Confirm functionality of the Change Form modal is unchanged.


Pelentan (approved):

shaneodd (approved):

mgaoVA (changes requested):
A different approach needs to be taken.

The core issue is:
> users were able to create [requests based on] hidden forms

We should prevent requests from being created if the form is marked as hidden.

Therefore:
1. Form.php -> newForm() should be updated to return an error string if the form is marked as hidden.
2. The "Copy Request" feature should show the error message.

jampaul3 (approved):

jampaul3 (approved):

Pelentan (approved):

shaneodd (approved):

PR #2613, opened 2024-12-03
LEAF-4604 update form progress checking to account for conditional logic

aerinkayne:
The current Form getProgress method still compares a flat array of required questions with data entries in order to determine form progress percentage.  Individual questions account for direct conditional logic.  However, since there is no information about form structure, using conditional logic when the conditional indicator has required sub-questions is problematic / tedious because the logic needs to be repeated on the subquestion.

This update will take form structure into account (when needed) during the progress count so that any ancestor in a hidden state is sufficient.

**Front end:**
Updates to form.js to handle conditional display states and to toggle front end validation for required questions

**Server-Side:**
Updates to form method getProgress to account for structure.
Following a check for submission and for the number of required questions, method will progress with a depth search assessment of question display state, visible required question, and total required question totals.  Required questions that are not in a visible state because they OR one of their ancestors are hidden by conditional do not count towards the visible total.

**Impact / Testing.**

Associated go test PR
https://github.com/department-of-veterans-affairs/LEAF-Automated-Tests/pull/20

This has high impact to the initial request submission process.
If the total is inaccurate
-data entry or submission could be prevented without a visible indication to the user as to why.
-submission could be allowed despite the user not having answered all required questions.

Test form data entry (initial request) and confirm that the percentage (form progress) reflects the percentage of answered required questions that are visible on the form.  Required questions hidden by ifthen logic directly, or because they are subquestions of hidden questions should not count towards the visible total.

### DIFF
--- a/LEAF_Request_Portal/sources/Form.php
+++ b/LEAF_Request_Portal/sources/Form.php
@@ -1509,206 +1509,208 @@ class Form
         return $return_value;
     }
 
-/**
+    /**
      * Get the progress percentage (as integer), accounting for conditinally hidden questions
      * @param int $recordID
      * @return int Percent completed
      */
     public function getProgress($recordID)
     {
-        $returnValue = 0;
-
-        $vars = array(':recordID' => (int)$recordID);
-        //get all the catIDs associated with this record and whether the forms are enabled or submitted
-        $resRecordInfoEachForm = $this->db->prepared_query('SELECT recordID, categoryID, `count`, submitted FROM records
-                                                    LEFT JOIN category_count USING (recordID)
-                                                    WHERE recordID=:recordID', $vars);
-        $isSubmitted = false;
+        $subSQL = 'SELECT submitted FROM records
+            LEFT JOIN category_count USING (recordID)
+            WHERE recordID=:recordID';
+        $subVars = array(':recordID' => (int)$recordID);
+        //First check if submitted and just return 100 if it is.
+        $resRecordInfoEachForm = $this->db->prepared_query($subSQL, $subVars);
         foreach ($resRecordInfoEachForm as $request) {
             if ($request['submitted'] > 0) {
-                $isSubmitted = true;
-                break;
+                return 100;
             }
         }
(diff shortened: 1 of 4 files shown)


shaneodd (approved):

jampaul3 (approved):

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1509,207 +1509,260 @@ public function doSubmit(int $recordID): array
...
+     * @param array dataTable (ref) lookup table of form data.  Used to check if question is answered and for conditional logic assessment
+     * @param array formNode form question with potential substructure
+     * @param bool parentOrSelfHidden is question and children in hidden state
+     * @param int required_visible (ref) number of required questions that are visible
+     * @param int required_answered (ref) number of required questions that are answered
+     * @param int required_total (ref) total number of required questions found during checking
+     * @param int res_max_required total number of required questions on the form
      */
-    public function getProgress($recordID)
+    private function count_required(
```
Pelentan, line 1523: Where's my return type?  ;-)

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1509,207 +1509,260 @@ public function doSubmit(int $recordID): array
...
-                } else {
-                    //finally, check if there are conditions that result in required questions being in a hidden state, and adjust count and percentage
-                    $multiChoiceParentFormats = array('multiselect', 'checkboxes');
-                    $singleChoiceParentFormats = array('radio', 'dropdown', 'number', 'currency');
+    /**
+     * Get the progress percentage (as integer), accounting for conditinally hidden questions
+     * @param int $recordID
+     * @return int Percent completed
+     */
+    public function getProgress($recordID)
```
Pelentan, line 1706: Type cast and return type, please.

Pelentan (changes requested):

Pelentan (dismissed):

shaneodd (approved):

jampaul3 (approved):

PR #2615, opened 2024-12-06
LEAF 4455 metadata and userMetadata reading

aerinkayne:
Uses data saved in the new metadata and userMetadata columns to replace display related orgchart lookups.
Lookups are still used for requests pending direct action and for emailing. 

**server side:**

Replaces lookupLogin calls with available metadata.
Replaces an orgchart join when getting initiator information with Report Builder queries

**front:**

data.metadata
Form print view, Report Builder.  Display of user info associated with orgchart employee format data entries.
For orgchart_employee format entries with metadata, the entered user name will continue to display even if the associated account is disabled.  Accounts disabled prior to metadata entry will continue to display disabled account information.

action_history.userMetadata
Request history modal, Report Builder (Approval History, Resolved By, Cancellation Date).  Display of user names associated with workflow and admin actions.

notes.userMetadata
Notes panel.  Request history modal (notes).  Display of user names.

records.userMetadata
Form print view.  Some workflow form fields (text only).  Report Builder Initiator queries.
The previous join used for initiator information looked up disabled accounts.  Records with empty records.userMetadata will display userID, (inactive account)  

 
**Testing / Impact**

These updates should not impact access but do have high visibility.
Confirm no unusual function or display changes. 
See Jira LEAF-4455 for more details about the above listed areas.

Associated Test DB branch
https://github.com/department-of-veterans-affairs/LEAF-Automated-Tests/pull/21
This brings the dev boilerplate up to date with a version containing all metadata columns and adds appropriate metadata values to initially inserted records


LEAF_Request_Portal/sources/Form.php
```diff
@@ -2605,16 +2606,14 @@ public function getCustomData(array $recordID_list, string|null $indicatorID_lis
...
-                            {
-                                $item['data'] = "{$empRes[0]['firstName']} {$empRes[0]['lastName']}";
-                                $item['dataOrgchart'] = $empRes[0];
-                            }
-                            else
-                            {
-                                $item['data'] = '';
+                        case 'orgchart_employee': //report builder cells
+                            $item['data'] = '';
+                            if (isset($item['metadata'])) {
```
Pelentan, line 2611: What happens if not "isset"?  Do we want a silent fail here?  Or do we want to catch it and provide something back?

LEAF_Request_Portal/sources/Form.php
```diff
@@ -3712,14 +3703,23 @@ public function query(string $inQuery): mixed
...
         }
 
-        if ($joinInitiatorNames)
-        {
-            $joins .= "LEFT JOIN (SELECT userName, lastName, firstName FROM {$this->oc_dbName}.employee) lj_OCinitiatorNames ON records.userID = lj_OCinitiatorNames.userName ";
-        }
+        //backwards compat: userMetadata properties are empty for accounts that were inactive when prior md values were updated.
+        //userMetadata alternatives here prevent the initiator field from displaying 'null, null' if metadata is empty.
+        //Handled here due to high customization of view_reports, view_search and other reports
+        $resSQL = 'SELECT *,
```
Pelentan, line 3709: This concerns me.  Not because of the select *, but because of the amount of processing it has to do on each line.  How much is this going to slow down db responses, especially during the middle of the day?  It might be better to do the pull for * first, process it on the machine designed for these operations, and then reach back out for the data it needs in addition.

LEAF_Request_Portal/sources/Form.php
```diff
@@ -1008,9 +1009,9 @@ public function getRecordInfo($recordID)
             );
         }
 
-        $dir = new VAMC_Directory;
-        $user = $dir->lookupLogin($res[0]['userID']);
-        $name = isset($user[0]) ? "{$user[0]['Fname']} {$user[0]['Lname']}" : $res[0]['userID'];
+        $userMetadata = json_decode($res[0]['userMetadata'], true);
+        $name = isset($userMetadata) && trim("{$userMetadata['firstName']} {$userMetadata['lastName']}") !== "" ?
+            "{$userMetadata['firstName']} {$userMetadata['lastName']}" : $res[0]['userID'];
```
Pelentan, line 1014: {$userMetadata['firstName']} {$userMetadata['lastName']} gets used a lot in this scope.  Both as a variable and as a determinant.  I'm thinking the first time it comes up, two variables should be created:  A string with the value and a binary for if it exists.

LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -494,8 +494,11 @@ public function getRecordsDependencyData(object $form, array $records, bool $sel
                         $approver = $dir->lookupLogin($res[$i]['userID']);
 
                         if (empty($approver[0]['Fname']) && empty($approver[0]['Lname'])) {
-                            $res[$i]['description'] = $res[$i]['stepTitle'] . ' (Requestor followup)';
-                            $res[$i]['approverName'] = '(Requestor followup)';
+                            $approverMetadata = json_decode($res[$i]['userMetadata'], true);
+                            $nameInfo = isset($approverMetadata) && trim($approverMetadata['firstName'] . " " . $approverMetadata['lastName'] ) !== '' ?
```
Pelentan, line 498: Same for this scope.

Pelentan (changes requested):

LEAF_Request_Portal/sources/Form.php
aerinkayne: Item data will remain empty.  This is what it had done previously, but since it also needs to be checked for a valid name it was simpler to set it at the top and update it if it had a valid name rather than set it afterwards.

LEAF_Request_Portal/sources/Form.php
aerinkayne: It looks like it might have added 10-25% to a 'select * ' query.  Going to rework this to keep the initiatorNames "join" logic, but only use it in a backwards compatible way to add the additional SQL if it's needed.  This will keep queries where initiator information is not requested the same speed, but still use json instead of the join when it is (where it should be considerably faster)

shaneodd (approved):

LEAF_Request_Portal/sources/Form.php
Pelentan: That doesn't resolve the issue I brought up.  The efficiency wasn't the concern, it was the repeated code.

jampaul3 (approved):

Pelentan (approved):

crystalkirscht:
Ran all Playwright tests in test/LEAF-4455/dev_database_info_updates on the code in enhance/LEAF-4455/metadata_read_all. Had to replace the reportBuilder.spec.ts in dev_database_info_updates with the version in main since it has been updated and now passed.

PR #2616, opened 2024-12-10
Leaf 4376 - update user name part 2

jampaul3:
Summary:
A way is needed to prevent a userName from being recycled and the new person being able to view data that does not belong to them.

Potential Impact:

Testing:
The only way to test this is on staging. And that is to run the script. Ensure that only those users who are disabled actually get their userName updated.


app/Leaf/VAMCActiveDirectory.php
```diff
@@ -0,0 +1,489 @@
...
+
+    // Connect to the database
+    public function __construct($national_db)
+    {
+        $this->db = $national_db;
+    }
+
+    // Imports data from \t and \n delimited file of format:
+    // Name	Business Phone	Description	Modified	E-Mail Address	User Logon Name
+    public function importADData($file)
```
shaneodd, line 45: Return types?

app/Leaf/VAMCActiveDirectory.php
```diff
@@ -0,0 +1,489 @@
...
+                FROM `employee`
+                WHERE `deleted` = 0
+                AND `lastUpdated` < (UNIX_TIMESTAMP(NOW() - 108000) ';
+
+        $return_value = $this->db->prepared_query($sql, array());
+
+        return $return_value;
+    }
+
+    private function getNewlyDeletedUsers($time): array
```
shaneodd, line 260: Input type?

shaneodd (changes requested):
Looks like some input/return types are missing? Looks like some areas have it.

shaneodd (approved):

aerinkayne (approved):

Pelentan (approved):

PR #2640, opened 2025-01-07
Leaf 4624 - db update for disabled users

jampaul3:
update cache table in the orgcharts to use utf8mb4 for the data field and table

testing:
once the update is complete the field should be utf8mb4 compatible.


shaneodd (approved):

aerinkayne (approved):

Pelentan (approved):

jampaul3:
With the pulling back of 4486 this one is no longer needed in the present. This will go away in favor of 4632 which will update the whole database to utf8mb4.

PR #2672, opened 2025-02-04
LEAF 4664 workflow module loading issue

aerinkayne:
Adds some comments to drawWorkflow to clarify that it is potentially called multiple times and that its step param also includes dependency information.
Adds a param to drawWorkflow to identify the requirement that the form fields are loaded to.

Fields are only loaded on the first requirement - running init on more than one was associated with breaking things.
The else clause has been updated so that it should reinit the module if
 -it is already flagged as loaded and
 -the depID/requirement currently being drawn is also the first one.

Updates ajax config object callback name to 'error'

**Testing / Impact**

This item impacts request workflow fields.  If these fields do not load or otherwise function correctly, the workflow process could be stopped and/or data could not correctly save.
-confirm fields load as expected, for both single and multi-requirement steps
-confirm that edited data and comments are consistently saved regardless of which requirement takes action

### DIFF
--- a/LEAF_Request_Portal/js/workflow.js
+++ b/LEAF_Request_Portal/js/workflow.js
@@ -205,9 +205,10 @@ var LeafWorkflow = function (containerID, CSRFToken) {
      * @memberOf LeafWorkflow
      * Called for each requirement with access. Initializes the step module if one exists for the step ID.
      * @param {object} step includes both step and depencency information for a specific requirement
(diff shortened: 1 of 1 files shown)


jampaul3 (approved):

Pelentan (approved):

shaneodd (approved):

shaneodd:
image
image

PR #2674, opened 2025-02-04
INC37469608 LEAF 4661 catch nexus side export error

aerinkayne:
On export, there is an endpoint call to pull indicator format information to avoid some issues with excel exports.

The orgchart does not have this endpoint, which results in a breaking error when attempting to export vacancy summary.

This update adds a try catch block so that potential errors will not affect the export.
It also adds pattern checking of the exported header ids to avoid trying to get format information if indicator patterns are not present.

Impact/Testing
formGrid export should work wherever it is used (Nexus Vacancy Summary page, Report Builder, Account Updater, Custom Reports)


shaneodd (approved):

Pelentan (approved):
This looks like it needs an api-test update or new test.

PR #2686, opened 2025-03-01
Clickable Links in User Content (Variation A)

mgaoVA:
This takes a slightly different approach than #2685 to enable clickable links in user content.

The main difference is that this parses user content into segments instead of processing the entire content at once. One benefit of this is that the same URL may appear more than once in the user's content.

Additionally, this:
- Organizes the code in a way that can potentially provide more compatibility
- Removes a couple IE polyfills

This variant minimizes changes in `print_subindicators_ajax.tpl`, which may benefit the stability of field-developed code at the cost of needing more code.

## Impact
Similar to #2685, this performs client-side text substitution, and impact is isolated in the printview context.

## Testing
Automated Test: https://github.com/department-of-veterans-affairs/LEAF-Automated-Tests/pull/55

PR #2687, opened 2025-03-01
LEAF-4720 - Clickable Links in User Content (Variation B)

mgaoVA:
See https://github.com/department-of-veterans-affairs/LEAF/pull/2686

This variant triggers the feature from individual data fields. It requires less overall code and has more forward compatibility, at the cost of having a higher chance of impacting field-developed code.

This variant is likely to be preferred, pending discussion.


shaneodd (approved):
I would say since this keeps things simpler even though it could nudge field code a bit. I think this may be the better option.

Pelentan (approved):

jampaul3 (approved):

aerinkayne (approved):

Pelentan (approved):

shaneodd (approved):

aerinkayne (approved):

PR #2688, opened 2025-03-03
LEAF 4685 - limit data importer requests with interval queue

aerinkayne:
## Summary

The initial scope of this update was to add the interval queue to the Toolbox - Import Spreadsheet (LEAF_import_data ) utility report.  Unlimited rates would cause imports with a large number of rows (~1000+) to fail on production.

The scope was expanded to resolve column order issues, fix an unresolved promise that would stop the import process, and add a needtoknow update if a question on a new form is marked as sensitive.
Minor template logic was also added to restrict display of the 'new form' choice to admins.

All changes are to the front end LEAF_import_data template file.

## Impact

This update affects the portal admin Import Spreadsheet utility.
Users with a moderate-large number of updates (~1000) will find that the process is slower.
This should correct breaking issues for imports with a large number of requests

## Testing

This report can be navigated to from
Portal Admin -> Toolbox -> Import spreadsheet

See attachments to Jira-4685 for a file with
a ~1200 row import
a file for testing orgchart errors
a local file for testing orgchart errors

The report import process is guided once a file is chosen.

Select 'new form' for the initial tests of the 1200 import and orgchart import (the import will create the associated form).
For the orgchart file, specify that 'Org Emp' is an orgchart employee format.  The other formats can be left as single line text.
Once a form exists, followup testing can use the existing form option.

- New Form option: indicators/questions should reflect the column order on the spreadsheets used to create them.
Go to the Form Editor to see the resulting form

- New Form option: If any question is marked as sensitive, form need to know will be turned on after questions have been added. 

- The unordered import of 1200 rows should succeed.  Use the resulting report button/link (it will open a Report Builder view) to get a list of the imported requests.  Modify the search and include the form's questions to confirm that data saved for the requests.  This should be re-tested on Academy/Demo1 (production) when the update is made live.

- Check the 'preserve row order' checkbox) to test importing requests in order (this process will be slower).
When an import is completed, a button will appear at the bottom of the report. Navigation will go to a Report Builder report. Choose 'Modify Search' and on step 2 select the letter and number columns for the associated form. Sort by UID - the number and letter columns can be used to judge whether the requests imported in order.

- The import of _orgchart_error file should finish.
A request will not be made for the row that errored, but the process itself should complete.  Previously, failure to locate an employee email would result in an unresolved promise and the import to stop at that row.

- Non-admins should not see the 'new form' option.
Go to adminer, leaf_portal_API_testing.users and change the row with tester tester groupID 1 to another number so they are not in admin. Change it back to 1 after testing.

### DIFF
--- a/LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
+++ b/LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
@@ -14,12 +14,15 @@
     #category_indicators td, #new_form_indicators td {
         padding: 7px;
     }
+    #category_select {
+        max-width: 600px;
+    }
 
     .modalBackground {
         width: 100%;
         height: 100vh;
         z-index: 5;
-        position: absolute;
+        position: fixed;
         background-color: grey;
         margin-top: 0px;
         margin-left: 0px;
@@ -81,7 +84,7 @@
     <br/><br/>
 </div>
 <div id="import_data_existing_form" style="display: none;">
-    <h4>Select a Form</h4>
+    <label for="category_select"><b>Select a Form</b></label>
     <select id="category_select"></select>
 
     <button id="import_btn_existing" type="button">Import</button>
@@ -392,10 +395,13 @@
         }
     }
 
-    function makeRequests(categoryID, requestData) {
-        const preserveOrder = $("#preserve_new").prop("checked");
-        requestData.sendRequestsSynchronously = preserveOrder;
-        
+    function handleMakeRequestFail(requestData, title, msg = "") {
+        failedRequests.push(requestData);
+        updateReportStatus(title);
+        console.log(msg)
+    }
+
+    function makeRequests(categoryID, requestData, preserveOrder) {
         const title = $('input[name="toggle"]:checked').attr('id') === 'newFormToggler' ? titleInputNew.val() : titleInputExisting.val();
 
         return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


shaneodd (dismissed):

Pelentan (dismissed):

jampaul3 (dismissed):

LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
```diff
@@ -561,183 +638,168 @@
...
-                                                        var emp = user[res[0]];
-                                                        if (typeof (emp) !== "undefined" && emp !== null && res.length === 1) {
-                                                            nexusAPI.Employee.importFromNational({
-                                                                'onSuccess': function (results) {
-                                                                    if (!isNaN(results)) {
-                                                                        requestData[indicatorArray[completed]] = parseInt(results);
-                                                                    } else {
-                                                                        requestData['failed'] = currentCol + titleIndex + ': Employee ' + sheetEmp + ' not found. Error: '+ results;
-                                                                    }
+                                return new Promise(function(resolve, reject) {
```
Pelentan, line 641: Convince me to allow this.  A return that returns promise that returns a promise and/or does a hundred lines of work, all inside a return.

LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
```diff
@@ -746,203 +808,189 @@
...
+
             totalImported = 0;
             $('#status').html('Processing...'); /* UI hint */
-
             requestStatus.html('Parsing sheet data...');
+            const preserveOrder = $("#preserve_existing").prop("checked");
 
+            let rowsCompleted = 0;
             function selectRowToAnswer(i) {
+                return new Promise(function(resolve,reject) {
```
Pelentan, line 823: Same here.

jampaul3 (approved):

shaneodd (approved):

shaneodd (changes requested):
This is the one where there are functions inside of the return. I had this confused with another one when I went through approved. This will need another pass.

LEAF_Request_Portal/templates/reports/LEAF_import_data.tpl
aerinkayne: I had mentioned this to Michael Gao too.  The page as a whole can use a more extensive refactor.  He wants to fix the major issues now and then address further refactoring in a future story.  But yes, it's functions that return promises that have function declarations in them that return promises that have function declarations that return more promises.

I pulled some methods out, but untangling everything would require a more significant amount of time.

Pelentan (approved):

shaneodd (approved):
Discussed this change, now I understand what is going on.

PR #2698, opened 2025-03-14
LEAF-4707 Avoid playwright warning

mgaoVA:
## Summary
This avoids an issue in the development environment, where on first run, the end2end test suite would error with "Looks like Playwright Test or Playwright was just installed or updated...".

## Impact
The changes are limited to the development environment.

## Testing
After deleting all development docker volumes, containers, and images, following the instructions in https://github.com/department-of-veterans-affairs/LEAF/blob/master/docs/InstallationConfiguration.md results in a working environment. Then, following prompts to run the end2end test suite should not result in any errors or failures.

### DIFF
--- a/docker/playwright/playwright.dockerfile
+++ b/docker/playwright/playwright.dockerfile
@@ -6,4 +6,4 @@ RUN npm install -D mysql2
 
 WORKDIR /usr/app/leaf
 RUN npx playwright install --with-deps
-CMD node main.js
+CMD npx playwright install && node main.js
--- a/docker/playwright/playwright.dockerfile
+++ b/docker/playwright/playwright.dockerfile
@@ -6,4 +6,5 @@ RUN npm install -D mysql2
 
 WORKDIR /usr/app/leaf
 RUN npx playwright install --with-deps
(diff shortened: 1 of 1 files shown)


jampaul3 (approved):

shaneodd:
> So the solution is to run npx playwright install twice?
I am not sure why it is an intermittent thing for me but I sometimes have to run it a second time. I do not remember seeing the text he is saying but it is not always an issue but sometimes get that error.

shaneodd (changes requested):
Because this one takes care of an error that sometimes happens and is an oddball thing. I would ask for a comment in the code. I can see someone looking at this down the road going why is this here.

aerinkayne (dismissed):

aerinkayne:
This does seem to resolve the issue, but I had been wondering if it could be volume related.  For the other containers, the installs stay there

mgaoVA:
> Because this one takes care of an error that sometimes happens and is an oddball thing. I would ask for a comment in the code. I can see someone looking at this down the road going why is this here.

Updated

shaneodd (approved):

aerinkayne (approved):

Pelentan (approved):

PR #2705, opened 2025-03-25
LEAF-4700 - if need to know is enabled, replace request titles in emails with the type of form.

aerinkayne:
## Summary
This update helps ensure sensitive data is not potentially displayed in emails.  The values of the 'fullTitle' and 'truncatedTitle' variables in the email template normally refer to the user-entered request title.  These values are replaced with the type of form if the form's need to know setting is turned on.

Since it has been requested by users in the past, formType has also been added to available smarty variables.

## Impact
Users should be notified of this update.  Template variables 'fullTitle' and 'truncatedTitle' will have unexpected values if the form's need to know setting is on.

## Testing
Within emails:
- If need to know is off, fullTitle and truncatedTitle should contain the title of the specific request.
- If need to know is on, fullTitle and truncatedTitle should contain the type of form (categoryName).
- Confirm current automated tests pass

See portal admin **Email Template Editor** for different types of email events:
_notify next approver, notify requestor of completion, custom email events, cancel notification, send back, automated email reminder, mass-action email reminders._

These are processed separately so each should be tested.
The original email templates contain truncatedTitle and fullTitle references.  Confirm these are still there if the templates have been customized.
Add a formType reference to the email content.

Update or add workflow events as needed and step a request through a workflow.
Cancel the request to trigger a cancel notification email.
Mass-action Email Reminders can be sent from portal admin -> **Mass Actions**.

Use the **SMTP Server** to confirm that emails send as expected and that they contain expected values.

### DIFF
--- a/LEAF_Request_Portal/sources/FormWorkflow.php
+++ b/LEAF_Request_Portal/sources/FormWorkflow.php
@@ -1344,16 +1344,17 @@ class FormWorkflow
             $strSQL = 'SELECT stepTitle FROM workflow_steps WHERE stepID = :stepID';
             $groupName = $this->db->prepared_query($strSQL, $vars);
 
-            $fullTitle = $record[0]['title'];
-            $formType = $record[0]['categoryName'];
+            $formType = trim(strip_tags(
+                htmlspecialchars_decode($record[0]['categoryName'], ENT_QUOTES | ENT_HTML5)
+            ));
 
+            $fullTitle = trim(strip_tags(
+                htmlspecialchars_decode($record[0]['title'], ENT_QUOTES | ENT_HTML5)
+            ));
             if($record[0]['needToKnow'] === 1) {
-                $title = $formType;
                 $fullTitle = $formType;
-            } else {
-                $title = strlen($fullTitle) > 45 ? substr($rfullTitle, 0, 42) . '...' : $fullTitle;
             }
-            $truncatedTitle = trim(strip_tags(htmlspecialchars_decode($title, ENT_QUOTES | ENT_HTML5 )));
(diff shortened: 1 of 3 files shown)


LEAF_Request_Portal/sources/FormWorkflow.php
```diff
@@ -1443,12 +1455,22 @@ public function handleEvents(int $workflowID, int $stepID, string $actionType, ?
...
-                        $title = strlen($requestRecords[0]['title']) > 45 ? substr($requestRecords[0]['title'], 0, 42) . '...' : $requestRecords[0]['title'];
-                        $truncatedTitle = trim(strip_tags(htmlspecialchars_decode($title, ENT_QUOTES | ENT_HTML5 )));
+                        $formType = trim(strip_tags(
+                            htmlspecialchars_decode($requestRecords[0]['categoryName'], ENT_QUOTES | ENT_HTML5)
+                        ));
+
+                        $fullTitle = trim(strip_tags(
+                            htmlspecialchars_decode($requestRecords[0]['title'], ENT_QUOTES | ENT_HTML5)
+                        ));
+                        if($requestRecords[0]['needToKnow'] === 1) {
```
Pelentan, line 1465: Little nervous here.  Are we sure that "needToKnow" is getting set as an int?  'Cause a string "1" will fail here.

Pelentan (changes requested):

LEAF_Request_Portal/sources/FormWorkflow.php
shaneodd: It is set to int in database, it should come back as int. I would also say we should parse it as an int for belt and suspenders

shaneodd (changes requested):

jampaul3 (approved):

Pelentan (approved):

PR #2710, opened 2025-03-27
Leaf 4711 - Mass action: Add "Take Action" feature

jampaul3:
## Summary
This change helps the customer take an action on several requests at the same time. It adds to the mass action functionality to allow a user to approve, send back to requester, etc in one action.

## Impact
The only changes made were to the mass action page, so impact should be very minimal to the rest of the platform. 

## Testing
Navigate to the Mass Action Page
Select "Take Action" from the actions dropdown
A dropdown of Forms should now be showing, select a form you wish to take action on
A dropdown for Steps used on this form shows, select a step to take action on
If this step has more than one requirement a dropdown of requirements will show, select the requirement to take action on
If the step only has one requirement you will not see this dropdown. The only requirement will automatically be chosen for you.
Now a dropdown with available actions for that requirement shows, select the action you wish to take
After all these steps have been completed a comments area shows, this is required, also all the requests that meet the selected criteria show up.
Select all the requests you want to take this action on and click the "Take Action" button
If you selected all the requests then the search results will be empty and tell you that there are no results.
If you only selected some of the requests then your list will get renewed and the ones you selected will be gone.


LEAF_Request_Portal/templates/reports/LEAF_mass_action.tpl
```diff
@@ -89,11 +95,818 @@
...
+        sort: {},
+    };
+
+    if (getCancelled) {
+        requestQuery.terms.push({
+            id: "stepID",
+            operator: "=",
+            match: "deleted",
+        });
+    } else if (!getAction) {
```
Pelentan, line 549: getAction is a string, not a boolean.

LEAF_Request_Portal/templates/reports/LEAF_mass_action.tpl
```diff
@@ -89,11 +95,818 @@
...
+
+    if (!getSubmitted) {
+        requestQuery.terms.push({
+            id: "stepID",
+            operator: "!=",
+            match: "submitted",
+        });
+    }
+
+    if (getReminder) {
```
Pelentan, line 565: getReminder is an int, not a boolean.

LEAF_Request_Portal/templates/reports/LEAF_mass_action.tpl
```diff
@@ -89,11 +95,818 @@
...
+    })
+        .done(function (data) {
+            if (thisSearchID === searchID) {
+                if (Object.keys(data).length) {
+                    let totalCount = 0;
+
+                    $.each(data, function (index, value) {
+                        let displayRecord = true;
+                        // If this is email reminder list, then compare against give time period
+                        if (getReminder) {
```
Pelentan, line 651: getReminder is a number here.

LEAF_Request_Portal/templates/reports/LEAF_mass_action.tpl
```diff
@@ -89,11 +95,818 @@
...
+    // Update global variables for execution - used in updateProgress function
+    // Setting them to default at beginning of mass execution run
+    processedRequests = 0;
+    totalActions = selectedRequests.length;
+    successfulActionRecordIDs = [];
+    failedActionRecordIDs = [];
+
+    $('#progressbar').progressbar('option', 'max', totalActions);
+
+    if (totalActions) {
```
Pelentan, line 771: totalActions is an int here.

Pelentan (changes requested):

mgaoVA (changes requested):
We should expand the scope of this story to include modernization in this refactoring effort.

Several issues exist in the original `mass_action.js` file, which should be corrected:
- Avoid use of jQuery in favor of plain JS
- Never use `async: false`. Use async/await instead.
- Never use direct ajax calls to `url: "./api/form/query"`. Use formQuery.js instead. [Example](https://github.com/department-of-veterans-affairs/LEAF/blob/ea450db7c478b1e5d08cf6d0b950924dc2d668bf/LEAF_Request_Portal/templates/reports/example.tpl#L18-L60)
  - Show progress since the queries can include many thousands of records.
  - Instead of building the query manually, use methods in formQuery.js : See above Example and [Comments](https://github.com/department-of-veterans-affairs/LEAF/blob/ea450db7c478b1e5d08cf6d0b950924dc2d668bf/LEAF_Request_Portal/js/formQuery.js#L29-L59)
- Avoid directly building HTML tables. Use formGrid.js instead. [Example](https://github.com/department-of-veterans-affairs/LEAF/blob/master/LEAF_Request_Portal/templates/reports/example.tpl#L46-L81)
- `queue.setConcurrency(1);` probably does not need to be set

I think @aerinkayne may have other notes related to potential improvement.

LEAF_Request_Portal/templates/reports/LEAF_mass_action.tpl
jampaul3: I didn't add this but updated it to > 0. Also if a variable is a number it will return true as long as it's not 0, 0 returns false.

LEAF_Request_Portal/templates/reports/LEAF_mass_action.tpl
jampaul3: This was just copied in. I did update it to be > 0

LEAF_Request_Portal/templates/reports/LEAF_mass_action.tpl
jampaul3: Same not mine but updated.

Pelentan (approved):

shaneodd (approved):

shaneodd (approved):

Pelentan (approved):

aerinkayne (approved):

mgaoVA (changes requested):
Line 693: Uncaught ReferenceError: actionValue is not defined
 
To reproduce the error in the dev environment:
1. Navigate to Mass Action
2. Select form Type: General Form
3. Select Step: Step 1
4. Select Requirement: Group A
5. Select Action: Approve
6. Click "Take Action"

aerinkayne (approved):

shaneodd (approved):

mgaoVA (approved):

Pelentan (approved):

PR #2714, opened 2025-04-03
Improve feature parity with the old inbox

mgaoVA:
## Summary
This resolves a few feature differences previously missed between the old and new Inboxes.

- Add "Type" column in the "Organize by Roles" view
- Increase the Inbox width

This also removes the 2 second redirection delay, which caused people to believe something was wrong.

## Impact
No dependencies

## Testing
TBD, Draft

### DIFF
--- a/LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
+++ b/LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
@@ -109,6 +109,20 @@
                     }
                 }
             },
+            'type': function(site) {
+                return {
+                    name: 'Type',
+                    indicatorID: 'type',
+                    editable: false,
+                    callback: function(data, blob) {
+                        let types = '';
+                        for(let i in blob[data.recordID].categoryNames) {
+                            types += blob[data.recordID].categoryNames[i] + ' | ';
+                        }
+                        types = types.substr(0, types.length - 3);
+                        document.querySelector(`#${data.cellContainerID}`).innerHTML = types;
+                }}
+            },
             'status': function(site) {
                 return {
                     name: 'Status',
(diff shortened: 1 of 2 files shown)


LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
```diff
@@ -109,6 +109,20 @@
                     }
                 }
             },
+            'type': function(site) {
+                return {
+                    name: 'Type',
+                    indicatorID: 'type',
+                    editable: false,
+                    callback: function(data, blob) {
+                        let types = '';
```
Pelentan, line 118: Something like this would be cleaner.  Wouldn't require trimming the string.

let typesArray = [];
for (let i in blob[data.recordID].categoryNames) {
    typesArray.push(blob[data.recordID].categoryNames[i]);
}
let types = typesArray.join(' | ');

aerinkayne (dismissed):

LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
mgaoVA: True. I'll update the other template where I copied this from too.

jampaul3 (approved):

LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
shaneodd: `document.querySelector(`#${data.cellContainerID}`).innerHTML = blob[data.recordID].categoryNames.join(' | ');` This looks like it was all distilled down to a one liner.

shaneodd (approved):

shaneodd (approved):

jampaul3 (approved):

shaneodd:
image
image
Tests run locally appear to pass, working with Nikesh to see if there are any other items that need to change or be added.

PR #2726, opened 2025-04-10
Example: Feature/load testing

shaneodd:
## Summary
image

This pr is here to show off what this tool can do. 

There is a command line version that we would most likely be using command line for running the tests. More info on that is here https://docs.locust.io/en/stable/quickstart.html#direct-command-line-usage-headless

You can also weight tests to try and steer the load testing to be more realistic. No matter what way we go I would probably want to try and get an overview of what pages are being loaded from Dynatrace and try and mimic what we see there. I could use our existing tests do but I think having an idea of what users are actually doing and basing our load test off of that would be the approach I would take.

### Steps to use
- `docker compose down && docker compose up -d --build` - rebuild your containers
- make sure your database is reset to a known base
- go to http://host.docker.internal:8089/
- I would suggest doing 10 concurrent users
- Run

It can give you statistics like this:
image

### Additional work
- This tool is easy enough to setup, would we want to have it runable on dev
- I would think adding in my "dump random data into the request table" could help here. We could then test with extra data coming back to simulate large reports. I know in the past we needed to add 10k results to get certain issues to actually show up.
- Authentication with auth server, still need to figure that out.
- Exporting csv from tests, I know it can be done just need steps for that, this would be good for reporting
- Getting what load tests we would want to run

## Impact
This would be for load testing our servers.

## Testing
This is going to be a testing tool


shaneodd:
This was originally here to talk about the load testing, I have actual load testing tests elsewhere and we probably do not need a local load test.

PR #2728, opened 2025-04-18
Leaf 4779 - email reminder last action

jampaul3:
## Summary
This PR addresses a bug where the mass action - email reminder, days since last action does not filter the result set. 

## Impact
This will allow filtering of data based on days since last action

## Testing
Navigate to the mass action page.
In the drop down select Email Reminder
The result set should filter down to requests where there is an action history and the date is older than 7 days.
If there are multiple requests with different dates of last actions you should be able to change the days since last action the the result set should change based on those days.


Pelentan (dismissed):

shaneodd (changes requested):
I see console.logs, are they needed?

Pelentan (approved):

aerinkayne (approved):

shaneodd (approved):

PR #2729, opened 2025-04-22
LEAF-4788 - INC38700605 : Inbox category checks

shaneodd:
## Summary
Add in undefined checks to verify data is present.

Note: Full URLs are within the JIRA ticket.
The portal outlined below came in with the Inbox not loading. This portal was due to a form being deleted, restoring the deleted form brought it back.
report.php?a=LEAF_Inbox&adminView&combineIndividuals&organizeByRole

## Impact
Users could have issues loading up to view requests on within the LEAF inbox.
## Testing
1. Have a form with a multi step workflow (EX: General Form)
2. Put a request on one of those steps
3. Delete the form
4. Try and load inbox

## Auto testing
https://github.com/department-of-veterans-affairs/LEAF-Automated-Tests/pull/92


LEAF_Request_Portal/templates/reports/LEAF_Inbox.tpl
```diff
@@ -555,8 +555,7 @@
                             });
                         }
                     });
-                });
-                document.querySelector(`#btn_action${hash}_${stepID}_${data.recordID}`).removeAttribute('disabled');
+                })
```
aerinkayne, line 558: Why is this being removed?
If this is re-added, there should be a check that the selector does not return null before calling its method

jampaul3 (dismissed):

jampaul3 (dismissed):

Pelentan (changes requested):
Need to be able to see the actual python file.  All that's being included is the compiled version.

Pelentan (approved):

aerinkayne (approved):

PR #2760, opened 2025-06-24
LEAF-4884 - Add Tools: Quick Review and Propose Actions

mgaoVA:
## Summary
This adds two new items to the admin toolbox: [Quick Review](https://github.com/department-of-veterans-affairs/LEAF-Developer-Examples/blob/master/utility/Quick_Review.md) and [Propose Actions](https://github.com/department-of-veterans-affairs/LEAF-Developer-Examples/blob/master/utility/Propose_Actions.md), originally provided as custom solutions.

Both tools provide enable bulk processing of reviews, which enhances the flexibility of the platform.

Quick Review:
- Admins can create a custom link to a page similar to the Inbox, where pending records for a specific workflow step are presented. This is intended to be bookmarked or provided on a webpage.

Propose Actions:
- Similar to Quick Review, however the final page URL is intended to be used in coordination with an approving official or committee review process.

## Impact
No dependencies.

## Testing
Navigate to the Admin Panel -> Toolbox -> `Setup Quick Review Link` / `Create Proposed Actions` and follow the prompts.

### DIFF
--- a/LEAF_Request_Portal/admin/templates/view_admin_menu.tpl
+++ b/LEAF_Request_Portal/admin/templates/view_admin_menu.tpl
@@ -203,9 +203,21 @@
     </a>
 
     <a href="../report.php?a=LEAF_Start_Request" role="button" class="leaf-admin-button bg-orange-10 lf-trans-orange">
-        <i class="leaf-admin-btnicon fas fa-solid fa-forward-fast text-orange-50 leaf-icn-narrow2" title="Combined Inbox Editor"></i>
+        <i class="leaf-admin-btnicon fas fa-solid fa-forward-fast text-orange-50 leaf-icn-narrow2" title=""></i>
         <span class="leaf-admin-btntitle">Setup Quickstart Link</span>
         <span class="leaf-admin-btndesc">Create a URL to start New Requests</span>
+    </a>
+
+    <a href="../report.php?a=LEAF_Quick_Review" role="button" class="leaf-admin-button bg-orange-10 lf-trans-orange">
+        <i class="leaf-admin-btnicon fas fa-solid fa-forward-fast text-orange-50 leaf-icn-narrow2" title=""></i>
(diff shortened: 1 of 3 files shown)


shaneodd (approved):

jampaul3 (dismissed):

aerinkayne (changes requested):
508/acc related:
There are input elements on these pages that do not have labels associated with them.  The img (button) that removes added columns should also have a button role and an aria label to indicate its function.  Testing tools can vary on contrast detection, but lighthouse hits 'red' with a contrast warning (something like #c00 should be ok).

Some CSS selectors are applying styling to the top nav buttons

mgaoVA:
> 508/acc related: There are input elements on these pages that do not have labels associated with them. The img (button) that removes added columns should also have a button role and an aria label to indicate its function. Testing tools can vary on contrast detection, but lighthouse hits 'red' with a contrast warning (something like #c00 should be ok).

Updated to resolve findings from Lighthouse.

> Some CSS selectors are applying styling to the top nav buttons

The round buttons are intended.

nk2136 (approved):
The functionality works correctly based on the selected step, as intended. However, there's slight confusion regarding the form selection dropdown:
	•	Currently, the form selection dropdown might imply that requests will be filtered by both form and step, whereas the actual filtering occurs primarily based on the step alone.
	•	Multiple forms can share the same workflow, so the resulting page could show requests from different forms based on the selected step, even if a specific form is chosen initially.
To reduce potential confusion, it might be beneficial to clarify this behavior directly in the UI or add a brief descriptive message. For instance:
"Selecting a form helps narrow available steps, but the resulting Quick Review page displays all requests associated with the chosen step, regardless of form."
This additional clarification would enhance the overall user experience.
Otherwise, this looks good. Approving the PR.

Pelentan (approved):

shaneodd (approved):

jampaul3 (approved):

22MGreen (approved):

PR #2761, opened 2025-06-24
Add missing response codes

mgaoVA:
## Summary
This improves error handling by returning an appropriate HTTP error code when the CSRFToken doesn't match.

This also improves CSRFToken security by using a string comparison function that's [more resistant to timing attacks](https://www.php.net/manual/en/function.hash-equals.php).

## Impact
This may cause $.ajax().error() to capture "new" errors -- which it's supposed to do.

## Testing
Sending an invalid CSRFToken (e.g. https://github.com/department-of-veterans-affairs/LEAF-Developer-Examples/tree/master/forms#create-a-new-record---example) returns HTTP 401.

- [ ] Automated tests pass

### DIFF
--- a/LEAF_Nexus/api/RESTfulResponse.php
+++ b/LEAF_Nexus/api/RESTfulResponse.php
@@ -71,8 +71,11 @@ abstract class RESTfulResponse
                 $DELETE_vars = [];
                 parse_str(file_get_contents('php://input', false, null, 0, 8192), $DELETE_vars); // only parse the first 8192 characters (arbitrary limit)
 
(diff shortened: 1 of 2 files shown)


aerinkayne (dismissed):

shaneodd (dismissed):

22MGreen (dismissed):
Tested and Verified error is displayed in the developer tab.

22MGreen:
This ticket was placed on hold, a conversation is needed to discuss the following: From so I have found that if you add this 
 
isset($DELETE_vars['CSRFToken']) &&

shaneodd (approved):

aerinkayne (approved):

PR #2764, opened 2025-06-30
LEAF 4888 - fix ifthen print display inconsistency

aerinkayne:
## Summary
In the IFTHEN app, users can configure conflicting display states and are not notified that this will cause problems.
A further issue related to this is that different pages behaved inconsistently in this situation.
While processing cannot account for conflicting display states both being true, pages need to behave similarly.

This update adds a warning in the IFTHEN app config if conflicting conditions are set.
An informational message clarifying how subquestion display is handled has also been added.

The data-entry and summary pages will also behave consistently in this situation.  Existing console information has more detail about what the issue is and has been elevated to 'warn'.

## Impact
IFTHEN configuration, summary / print review page behavior

## Testing
If a question is configured to both 'hide except' and 'show except', a clear warning message will be displayed.

If a question with IFTHEN logic has subquestions, information will display to inform the user that all subquestions will also be hidden.

Data-entry page and Summary page will **behave the same way** (this is addressing consistency, not config outcome).


shaneodd:
Closing this out since there were problems merging this one.

PR #2778, opened 2025-07-28
LEAF 4892 update Email Template Editor to allow org groups

aerinkayne:
## Summary
Adds Orgchart Group as an allowed format in the Email Template Editor To and Cc fields
Adds missing smarty email variable assignment to next approver template.

When used within To or Cc fields, the orgchart_group will be used to get a list of group member emails 

## Impact
Email Template Editor, Emailing Events

## Testing
[End2End Testing PR](https://github.com/department-of-veterans-affairs/LEAF-Automated-Tests/pull/138)

Add an emailing event to General Workflow's submit event (or any other suitable action)

Edit that Email Event's template to include a non-sensitive orgchart_group data field in the To or Cc sections.
For example, 
{{$field.9}}

Create a request for a form that uses the General Workflow (eg general form)
Ensure there are users in whichever group is entered into data field 9.
Submit the request.

Use the SMTP Server link to check for email recipients.
Members of the selected group should have been emailed.


shaneodd (dismissed):

Pelentan (changes requested):
Looks like the PHPDoc parts need some updates.

Pelentan (approved):

jampaul3 (approved):

PR #2787, opened 2025-08-14
LEAF-4999 - Control track changes

mgaoVA:
## Summary
:warning: Deployment: This includes a database schema change, which must be applied prior to this patch.

This enables form fields to be used as an efficient data store for ephemeral content.

The LEAF Agent logs metadata server-side, however relevant data is difficult to review for admins. This uses form fields to make metadata such as the "Timestamp for last successful run" and specific error logs more accessible to admins. However since the metadata can be updated as frequently as once every 10 seconds, this can unnecessarily bloat the local database's changelog.

The new portal database column `indicators.trackChanges` provides server-side control to avoid maintaining a full changelog for specific indicators. The direct change is still logged via timestamp and userID.

## Impact
Track changes is enabled by default to avoid impact on all sites. There is no user-facing control for this option, and even when track changes is disabled, the last modified timestamp and related userID is still tracked.

## Testing
In the development environment:
1. Reset and update the database by running the API tester
2. Navigate to the Request Portal (Standard Test DB)
3. Open request # 957
4. Change the Single line text field to `ABCDE`
5. Refresh the page
6. Check the history for the Single line text field
7. You should expect to see a log entry with the change you just made.
8. Navigate to the database: leaf_portal_API_testing.indicators
9. Edit indicatorID # 3, and change `trackChanges` to 0
10. Navigate back to request # 957
11. Change the Single line text field to `12345`
12. Check the history for the Single line text field (note refresh is not needed because the history button is already visible)
13. You should not expect to see a new log entry

### DIFF
--- a/LEAF_Request_Portal/sources/Form.php
+++ b/LEAF_Request_Portal/sources/Form.php
@@ -1117,7 +1117,7 @@ class Form
                       ':indicatorID' => $key,
                       ':series' => $series, );
 
-        $sql = "SELECT `format`, `data` FROM `data`
+        $sql = "SELECT `format`, `trackChanges`, `data` FROM `data`
             LEFT JOIN `indicators` USING (`indicatorID`)
             WHERE `recordID`=:recordID AND `indicators`.`indicatorID`=:indicatorID AND `series`=:series";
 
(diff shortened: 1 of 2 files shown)


aerinkayne (approved):

crystalkirscht:
Followed test steps and they all passed:

**1. **Reset and update the database by running the API tester**** 
image 

**2. **Navigate to the Request Portal (Standard Test DB)**** 
image

**3. Open request # 957** 
image

**4. Change the single-line text field to ABCDE.**
image

**5. Refresh the page**
image

**6. Check the history for the single-line text field**
**7. You should expect to see a log entry with the change you just made.**
image

**8. Navigate to the database: leaf_portal_API_testing.indicators**
image

**9. Edit indicatorID # 3, and change trackChanges to 0**
image
image

**10. Navigate back to request # 957**
**11. Change the single-line text field to 12345**
image

**12. Check the history for the single-line text field (note: refresh is not needed because the history button is already visible)**
**13. You should not expect to see a new log entry.** 
image

crystalkirscht (approved):
Followed test steps and all passed. See the comment above with screenshots

PR #2790, opened 2025-08-15
LEAF-5003 - uriencode double slashes

mgaoVA:
## Summary
This resolves an issue where searching for a URL results in an error.

## Impact
Minimal since this URI encodes double slashes within the URL's query parameters, which generally expect URI encoded data.

## Testing
1. Change a record's title to `https://leaf.va.gov`
2. On the homepage, search for `https://leaf.va.gov`
3. You should expect to see the record from step 1


shaneodd (approved):

aerinkayne (approved):

crystalkirscht (approved):
Created end2end test which confirms a request with a URL in the title can be searched for using both the search field and the advanced options. Both pass.

PR #2793, opened 2025-08-20
Leaf 4789 - SSN Scan

jampaul3:
## Summary
This allows non-admins to view a portal's Admin list.

This modifies the `GET ./api/group/1/members` endpoint's access control, which previously required admin access. This change is needed to support activities related to the SSN Reduction Act.

The access control change was determined to be acceptable, as it's important for a user of the site to be able to contact its administrator, and sites routinely list them.

## Impact
The impact here is that a list of admins for a site are now available to everyone, not just other admins.

## Testing
Nothing to test here.

### DIFF
--- a/LEAF_Request_Portal/api/index.php
+++ b/LEAF_Request_Portal/api/index.php
@@ -21,11 +21,13 @@ $p_db = $db;
 $action = isset($_GET['a']) ? $_GET['a'] : $_SERVER['PATH_INFO'];
 $keyIndex = strpos($action, '/');
 $key = null;
+$parts = null;
 
 if ($keyIndex === false) {
     $key = $action;
 } else {
     $key = substr($action, 0, $keyIndex);
+    $parts = explode('/', substr($action, $keyIndex + 1));
 }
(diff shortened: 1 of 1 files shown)


mgaoVA (changes requested):
Please update the summary to prioritize the actual changes in this PR.

mgaoVA (changes requested):
Thanks for updating the summary, however the SSN scanning explains why this exist, not what this particular PR does.

To be more clear about what this PR does, examine it in isolation. It changes access control for a specific endpoint.

E.g.
```markdown
This allows non-admins to view a portal's Admin list.

This modifies the `GET ./api/group/1/members` endpoint's access control, which previously required admin access. This change is needed to support activities related to the SSN Reduction Act.

The access control change was determined to be acceptable, as it's important for a user of the site to be able to contact its administrator, and sites routinely list them.
```

This speaks specifically to this PR (admin list only for portals), provides context as to why the change was made, and since this deals with access control: why it's OK to change it.

PR #2805, opened 2025-09-22
Leaf 5065 - keeping db clean after dependency removal

jampaul3:
## Summary
This ensures that the database is in the correct state after removing a data field from a workflow.

## Impact
After removing a data field from a workflow step if you re-add it to the same step you will need to reassign the indicatorID to it where as before it was still there.

## Testing
Not sure automated tests need to be completed for this or not. Best way to test this is to:
1. add a Person/Group designated to a workflow step
2. select the data field
3. remove the Person/Group
4. re-add the same thing just removed. 

If this is working properly you will need to select the data field again. (In production now when you re-add it the data field is already there from the last time)

### DIFF
--- a/LEAF_Request_Portal/sources/Workflow.php
+++ b/LEAF_Request_Portal/sources/Workflow.php
@@ -582,8 +582,8 @@ class Workflow
                 )
             )
         );
-        $strSQL = "UPDATE events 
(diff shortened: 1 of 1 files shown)


shaneodd (approved):

jampaul3:
This was merged into 5038 and is no longer needed as a stand alone.

